PR #70, opened 2018-09-07
Fix occasional ObjectDisposedException

MikhailArkhipov:
Analyzer can raise 'new analysis' from task that is still running when Server has been disposed


src/Analysis/Engine/Impl/ProjectEntry.cs
```diff
@@ -291,7 +292,7 @@ private class ActivePythonParse : IPythonParse {
                 string pathPrefix = PathUtils.EnsureEndSeparator(Path.GetDirectoryName(FilePath));
                 var children =
                     from pair in ProjectState.ModulesByFilename
-                    // Is the candidate child package in a subdirectory of our package?
+                        // Is the candidate child package in a subdirectory of our package?
```
AlexanderSher: ?

src/LanguageServer/Impl/Implementation/ProjectFiles.cs
```diff
@@ -29,15 +29,24 @@ internal sealed class ProjectFiles : IDisposable {
...
         public IProjectEntry GetOrAddEntry(Uri documentUri, IProjectEntry entry) {
             ThrowIfDisposed();
             return _projectFiles.GetOrAdd(documentUri, entry);
         }
 
         public IProjectEntry RemoveEntry(Uri documentUri) {
             ThrowIfDisposed();
-            return _projectFiles.TryRemove(documentUri, out var entry) ? entry : null;
+            _projectFiles.TryRemove(documentUri, out var entry);
+            entry?.Dispose();
```
AlexanderSher: `ProjectFiles` doesn't create entries or manage their lifetime otherwise, so it's up to the caller to dispose it.

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -532,7 +532,8 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
                 return Task.FromResult(actualItem);
             }
 
-            pyItem.OnNewAnalysis += (o, e) => OnPythonEntryNewAnalysis(pyItem);
+            pyItem.NewAnalysis += OnProjectEntryNewAnalysis;
+            pyItem.Disposed += (s, e) => ((IPythonProjectEntry)s).NewAnalysis -= OnProjectEntryNewAnalysis;
```
AlexanderSher: I'd say that this is actually a responsibility of `ProjectEntry` itself - it should clear its own event handlers on dispose.

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -545,6 +546,34 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
             return Task.FromResult(item);
         }
 
+        private void OnProjectEntryNewAnalysis(object sender, EventArgs e) {
+            // Events are fired from a background thread and may come
+            // after items were actually already disposed.
+            if(_disposed) {
```
AlexanderSher: Don't we use `DisposableToken` in `Server` implementation?

src/Analysis/Engine/Impl/ProjectEntry.cs
MikhailArkhipov: C# formatting

src/LanguageServer/Impl/Implementation/ProjectFiles.cs
MikhailArkhipov: Except they don't do it. Maybe it should not be disposable and we can just ignore events from it.

src/LanguageServer/Impl/Implementation/Server.cs
MikhailArkhipov: OK

MikhailArkhipov: Actually on RemovedFromProject I guess

MikhailArkhipov: Replace RemovedFromProject  by Dispose?

src/Analysis/Engine/Impl/Infrastructure/Extensions/EventHandlerExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+
+namespace Microsoft.PythonTools.Analysis.Infrastructure {
+    public static class EventHandlerExtensions {
+        public static void DisconnectListeners<T>(this EventHandler<T> handler) where T: EventArgs {
```
AlexanderSher: This method won't work - it will change only parameter, not the original event.

src/Analysis/Engine/Impl/ProjectEntry.cs
```diff
@@ -364,6 +363,9 @@ where fileName.StartsWithOrdinal(pathPrefix, ignoreCase: true)
                 foreach (var moduleReference in MyScope.ModuleReferences.ToList()) {
                     MyScope.RemoveModuleReference(moduleReference);
                 }
+
+                NewParseTree.DisconnectListeners();
```
AlexanderSher: `NewParseTree -= NewParseTree` should be enough.

src/Analysis/Engine/Impl/ProjectEntry.cs
```diff
@@ -81,8 +81,9 @@ IAnalysisCookie cookie
             return u;
         }
 
-        public event EventHandler<EventArgs> OnNewParseTree;
-        public event EventHandler<EventArgs> OnNewAnalysis;
+        public event EventHandler<EventArgs> NewParseTree;
+        public event EventHandler<EventArgs> NewAnalysis;
+        public event EventHandler<EventArgs> Disposed;
```
AlexanderSher: We aren't using `Disposed` event in the last version of this PR

src/LanguageServer/Impl/Disposables/DisposeToken.cs
```diff
@@ -1,82 +0,0 @@
...
-//
-// See the Apache Version 2.0 License for specific language governing
-// permissions and limitations under the License.
-
-using System;
-using System.Threading;
-using static System.FormattableString;
-
-namespace Microsoft.DsTools.Core.Disposables {
-    public sealed class DisposeToken {
```
AlexanderSher: Duplicate?

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -92,33 +89,39 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
+
             _pendingAnalysisEnqueue = new VolatileCounter();
-            _parseQueue = new ParseQueue();
+            ParseQueue = new ParseQueue();
             _pendingParse = new Dictionary<IDocument, VolatileCounter>();
             _editorFiles = new EditorFiles(this);
             _extensions = new ConcurrentDictionary<string, ILanguageServerExtension>();
+
+            _disposableBag
+                .Add(() => ProjectFiles.Dispose())
```
AlexanderSher: `DisposableBag` has an overload of `Add` that accepts `IDisposable`, so `_disposableBag.Add(ProjectFiles)` is enough.

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -92,33 +89,39 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
+                .Add(() => Analyzer?.Dispose())
```
AlexanderSher: `Analyzer` is always null here, it is set only in `DoInitializeAsync`

src/LanguageServer/Impl/Disposables/DisposeToken.cs
MikhailArkhipov: It was in LS, but engine could not use it, I moved it to infra. We soon will need some DSTools.Core :-)

src/LanguageServer/Impl/Implementation/Server.cs
MikhailArkhipov: Right, but when action is invoked it may not be?

src/LanguageServer/Impl/Disposables/DisposeToken.cs
MikhailArkhipov: Speaking of which, I can port disposable tests from RTVS too.

PR #79, opened 2018-09-10
Fix tuple args signature test

MikhailArkhipov:
It was broken with https://github.com/Microsoft/PTVS/pull/4684 that introduced argument types to tuples. However, for basic typeless case we don't need '[]'. Alternatively, we can instead show `[]` that is base enumerable since `*a` accepts both tuples and lists.

### DIFF
--- a/src/Analysis/Engine/Impl/Values/DictionaryInfo.cs
+++ b/src/Analysis/Engine/Impl/Values/DictionaryInfo.cs
@@ -348,7 +348,7 @@ namespace Microsoft.PythonTools.Analysis.Values {
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/Analysis/Engine/Impl/Values/DictionaryInfo.cs
```diff
@@ -348,6 +348,8 @@ public UpdateItemsAnalysisUnit(DictionaryInfo dictInfo)
             }
         }
 
+        public IEnumerable<IVariableDefinition> IndexTypes => new[] { _keyValueTupleVariable };
```
AlexanderSher: Can `_keyValueTupleVariable` be null?

PR #82, opened 2018-09-10
Fix CompletionAfterLoad test

MikhailArkhipov:
Actually looks like product bug. When adding/removing files, for dependent files we should be going through normal chain via `EnqueueItem` which does many useful things and not directly calling `AnalysisQueue.Enqueue`. We do can save on not parsing, but events should be fired and diagnostics should be updated.

### DIFF
--- a/src/LanguageServer/Impl/Implementation/Server.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.cs
@@ -298,7 +298,7 @@ namespace Microsoft.Python.LanguageServer.Implementation {
         public Task<bool> UnloadFileAsync(Uri documentUri) {
             var entry = RemoveEntry(documentUri);
             if (entry != null) {
-                Analyzer.RemoveModule(entry, e => AnalysisQueue.Enqueue(e, AnalysisPriority.Normal));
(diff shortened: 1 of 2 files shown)


AlexanderSher:
Will look closely on this tomorrow. I remember that code being rather fragile.

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -603,26 +603,29 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
-                    TraceMessage($"Parsing document {doc.DocumentUri}");
-                    cookieTask = ParseQueue.Enqueue(doc, Analyzer.LanguageVersion);
-                }
-
                 if (doc is ProjectEntry entry) {
                     entry.ResetCompleteAnalysis();
                 }
 
+                // If we don't need to parse, use null cooking
+                var cookieTask = Task.FromResult<IAnalysisCookie>(null);
```
AlexanderSher: If cookie is null, `EditorFile.UpdateParseDiagnostics` won't be called. Is that expected?

MikhailArkhipov: Yes, since we didn't reparse

PR #86, opened 2018-09-12
Fix AstTypeStubPaths* tests

MikhailArkhipov:
```cs
                analysis.Should().HavePythonModuleVariable("Package")
                    .Which.Should().HaveNestedModule("Module")
                    .Which.Should().HaveMultipleTypesMember("Class");
```
does not always apply, the structure is different in all cases. Original tests only verified actual variable members for completion.

### DIFF
--- a/src/Analysis/Engine/Impl/PythonAnalyzer.cs
+++ b/src/Analysis/Engine/Impl/PythonAnalyzer.cs
@@ -58,7 +58,7 @@ namespace Microsoft.PythonTools.Analysis {
         private readonly List<string> _searchPaths = new List<string>();
         private readonly List<string> _typeStubPaths = new List<string>();
         private readonly Dictionary<string, List<SpecializationInfo>> _specializationInfo = new Dictionary<string, List<SpecializationInfo>>();  // delayed specialization information, for modules not yet loaded...
-        private AnalysisLimits _limits;
+        private AnalysisLimits _limits = AnalysisLimits.GetDefaultLimits();
         private static object _nullKey = new object();
         private readonly SemaphoreSlim _reloadLock = new SemaphoreSlim(1, 1);
         private Dictionary<IProjectEntry[], AggregateProjectEntry> _aggregates = new Dictionary<IProjectEntry[], AggregateProjectEntry>(AggregateComparer.Instance);
@@ -623,11 +623,11 @@ namespace Microsoft.PythonTools.Analysis {
         public AnalysisLimits Limits {
             get { return _limits; }
             set {
+                value = value ?? AnalysisLimits.GetDefaultLimits();
                 var limits = _limits;
                 _limits = value;
 
-                if ((limits == null && _limits != null)
-                    || limits.UseTypeStubPackages ^ _limits.UseTypeStubPackages
+                if (limits.UseTypeStubPackages ^ _limits.UseTypeStubPackages
                     || limits.UseTypeStubPackagesExclusively ^ _limits.UseTypeStubPackagesExclusively) {
                     SearchPathsChanged?.Invoke(this, EventArgs.Empty);
                 }
--- a/src/Analysis/Engine/Impl/PythonAnalyzer.cs
+++ b/src/Analysis/Engine/Impl/PythonAnalyzer.cs
@@ -34,7 +34,7 @@ namespace Microsoft.PythonTools.Analysis {
     /// <summary>
     /// Performs analysis of multiple Python code files and enables interrogation of the resulting analysis.
     /// </summary>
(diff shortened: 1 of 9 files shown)


MikhailArkhipov:
WRT  `FindExpressionsForDefinition` test - it was failing that expectation for def location on `.` position was `no expression`. IMO it is OK if instead it finds `C().f(1)` since `.` is the root operator here.

src/Analysis/Engine/Test/AstAnalysisTests.cs
```diff
@@ -381,91 +377,91 @@ class BankAccount(object):
...
                     .WithValue<IBuiltinInstanceInfo>()
-                    .Which.Should().HaveMemberType(PythonMemberType.Instance)
-                    .And.HavePythonType(type)
-                    .Which.Should().HaveMethod("typed_method_2")
+                    .Which.Should().HaveMembers("typed_method_2")
                     .And.NotHaveMembers("untyped_method", "inferred_method", "typed_method");
             }
         }
 
+        private async Task<IModuleAnalysis> GetStubBasedAnalysis(
```
AlexanderSher: It doesn't look like this method has reduced the amount of code.

MikhailArkhipov: It is more if we want to add additional tests with actual multiple paths and stubs. Existing one are rather primitive.

src/Analysis/Engine/Test/FluentAssertions/AnalysisValueAssertions.cs
```diff
@@ -95,6 +95,22 @@ public AndConstraint<TAssertions> HaveMembers(params string[] memberNames)
             return new AndConstraint<TAssertions>((TAssertions)this);
         }
 
+        public AndConstraint<TAssertions> NotHaveMembers(params string[] memberNames)
+            => NotHaveMembers(memberNames, string.Empty);
+
+        public AndConstraint<TAssertions> NotHaveMembers(IEnumerable<string> memberNames, string because = "", params object[] reasonArgs) {
+            var actualNames = Subject.GetAllMembers(((ModuleScope)OwnerScope.GlobalScope).Module.InterpreterContext).Keys.ToArray();
```
AlexanderSher: `((ModuleScope)OwnerScope.GlobalScope).Module.InterpreterContext` is used in several places. Let's extract it into helper method.

src/Analysis/Engine/Test/AstAnalysisTests.cs
AlexanderSher: Is it actually possible in VSC to change these parameters (Limits, SearchPaths) after server is started?

MikhailArkhipov: Currently no. LS restart is required.

AlexanderSher: In this case, maybe we should simplify these tests by setting all these values from the beginning? We can add another overload of `ServerExtensions.InitializeAsync`.

MikhailArkhipov: But is actually good to set them later. LS doesn't do it for simplicity but it really should not be requiring restart. So at some point we should handle change.

AlexanderSher: Ok

src/Analysis/Engine/Impl/PythonAnalyzer.cs
```diff
@@ -622,7 +622,16 @@ class LazyModuleEnumerator {
 
         public AnalysisLimits Limits {
             get { return _limits; }
-            set { _limits = value; }
+            set {
+                var limits = _limits;
+                _limits = value;
+
+                if ((limits == null && _limits != null)
```
AlexanderSher: Can it be the other way: `limits != null && _limits == null`?

src/Analysis/Engine/Impl/Values/DictionaryInfo.cs
```diff
@@ -208,7 +208,7 @@ public DictionaryInfo(IPythonProjectEntry declaringModule, Node node)
             }
         }
 
-        public override PythonMemberType MemberType => PythonMemberType.Field;
+        public override PythonMemberType MemberType => PythonMemberType.Instance;
```
AlexanderSher: Is it for some other bug?

src/Analysis/Engine/Impl/PythonAnalyzer.cs
MikhailArkhipov: Actually I think it should be using default and never be null

Issue #87, opened 2018-09-12
NullReferenceException in ExpressionEvaluator.EvaluateWorker


MikhailArkhipov:
From https://github.com/Microsoft/vscode-python/issues/2297
```
[Error - 07:34:56] System.NullReferenceException: Object reference not set to an instance of an object.
   at System.Object.GetType()
   at Microsoft.PythonTools.Analysis.Analyzer.ExpressionEvaluator.EvaluateWorker(Node node)
   at Microsoft.PythonTools.Analysis.Analyzer.DDG.Walk(ClassDefinition node)
   at Microsoft.PythonTools.Parsing.Ast.ClassDefinition.Walk(PythonWalker walker)
   at Microsoft.PythonTools.Analysis.Analyzer.DDG.Walk(SuiteStatement node)
   at Microsoft.PythonTools.Parsing.Ast.SuiteStatement.Walk(PythonWalker walker)
   at Microsoft.PythonTools.Analysis.Analyzer.DDG.WalkBody(Node node, AnalysisUnit unit)
   at Microsoft.PythonTools.Analysis.ClassAnalysisUnit.AnalyzeWorker(DDG ddg, CancellationToken cancel)
   at Microsoft.PythonTools.Analysis.Analyzer.DDG.Analyze(Deque`1 queue, CancellationToken cancel, Action`1 reportQueueSize, Int32 reportQueueInterval)
   at Microsoft.PythonTools.Analysis.PythonAnalyzer.AnalyzeQueuedEntries(CancellationToken cancel)
   at Microsoft.PythonTools.Intellisense.AnalysisQueue.GroupAnalysis.Analyze(CancellationToken cancel)
   at Microsoft.PythonTools.Intellisense.AnalysisQueue.Worker(Object threadStarted)
```

PR #92, opened 2018-09-13
Fix existing and potentil null references

MikhailArkhipov:
Basically stems out of https://github.com/Microsoft/python-language-server/issues/87

- Remove BasesInternal and use Bases
- Ditto for Parameters and Decorators
- Return empty arrays rather than nulls
- Use MaybeEnumerate where appropriate
- Add and use ExceptDefault()


src/Analysis/Engine/Impl/Intellisense/ExtractedMethodCreator.cs
```diff
@@ -61,25 +63,18 @@ class OutOfProcExtractedMethodCreator {
             if (_targetScope is ClassDefinition) {
                 var fromScope = _scopes[_scopes.Count - 1] as FunctionDefinition;
```
AlexanderSher: Maybe use pattern matching here:
```cs
if (_targetScope is ClassDefinition && _scopes[_scopes.Count - 1] is FunctionDefinition fromScope)
```

PR #96, opened 2018-09-14
Better progress reporting

### LINKED ISSUE #94: Progress reporting does not always clear when analysis completes
https://github.com/Microsoft/vscode-python/issues/2297

Trace shows analysis completed but status bar is not clear

MikhailArkhipov:
Fixes #94 

- Make separate class for progress reporting
- Do not use 'estimate work' and rather track specific events
- Make reporting thread safe so we don't end up with multiple competing reporting tasks

### DIFF
--- a/src/LanguageServer/Impl/Definitions/ILogger.cs
+++ b/src/LanguageServer/Impl/Definitions/ILogger.cs
@@ -19,5 +19,6 @@ using System;
 namespace Microsoft.Python.LanguageServer {
     public interface ILogger {
         void TraceMessage(IFormattable message);
+        void TraceMessage(string message);
     }
 }
--- a/src/LanguageServer/Impl/Definitions/ILogger.cs
+++ b/src/LanguageServer/Impl/Definitions/ILogger.cs
@@ -19,6 +19,5 @@ using System;
 namespace Microsoft.Python.LanguageServer {
     public interface ILogger {
         void TraceMessage(IFormattable message);
-        void TraceMessage(string message);
     }
 }
--- a/src/LanguageServer/Impl/Implementation/AnalysisProgressReporter.cs
+++ b/src/LanguageServer/Impl/Implementation/AnalysisProgressReporter.cs
@@ -1,0 +1,85 @@
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using Microsoft.PythonTools.Analysis.Infrastructure;
+
+namespace Microsoft.Python.LanguageServer.Implementation {
+    sealed class AnalysisProgressReporter : IDisposable {
+        private readonly DisposableBag _disposables = new DisposableBag(nameof(AnalysisProgressReporter));
+        private readonly Dictionary<Uri, int> _activeAnalysis = new Dictionary<Uri, int>();
+        private readonly IProgressService _progressService;
+        private readonly ILogger _logger;
+        private readonly Server _server;
+        private readonly object _lock = new object();
+        private IProgress _progress;
+
+        public AnalysisProgressReporter(Server server, IProgressService progressService, ILogger logger) {
+            _progressService = progressService;
+            _logger = logger;
+
+            _server = server;
+            _server.OnAnalysisQueued += OnAnalysisQueued;
+            _server.OnAnalysisComplete += OnAnalysisComplete;
+            _disposables
+                .Add(() => _server.OnAnalysisQueued -= OnAnalysisQueued)
+                .Add(() => _server.OnAnalysisComplete -= OnAnalysisComplete)
+                .Add(() => _progress?.Dispose());
+        }
+
+        public void Dispose() {
+            _disposables.TryDispose();
+        }
+
+        private void OnAnalysisQueued(object sender, AnalysisQueuedEventArgs e) {
+            lock (_lock) {
+                if (_activeAnalysis.ContainsKey(e.uri)) {
+                    _activeAnalysis[e.uri]++;
+                } else {
+                    _activeAnalysis[e.uri] = 1;
+                }
+                UpdateProgressMessage();
+            }
+        }
+        private void OnAnalysisComplete(object sender, AnalysisCompleteEventArgs e) {
+            lock (_lock) {
+                if (_activeAnalysis.TryGetValue(e.uri, out var count)) {
+                    if (count > 1) {
+                        _activeAnalysis[e.uri]--;
+                    } else {
+                        _activeAnalysis.Remove(e.uri);
+                    }
+                } else {
+                    _logger.TraceMessage($"Analysis completed for {e.uri} that is not in the dictionary.");
+                }
+                UpdateProgressMessage();
+            }
+        }
+
+        private void UpdateProgressMessage() {
+            if(_activeAnalysis.Count > 0) {
+                _progress = _progress ?? _progressService.BeginProgress();
+                _progress.Report(_activeAnalysis.Count == 1
+                    ? Resources.AnalysisProgress_SingleItemRemaining
+                    : Resources.AnalysisProgress_MultipleItemsRemaining.FormatInvariant(_activeAnalysis.Count)).DoNotWait();
+            } else {
+                _progress?.Dispose();
+                _progress = null;
+            }
+        }
+    }
+}
--- a/src/LanguageServer/Impl/Implementation/AnalysisProgressReporter.cs
+++ b/src/LanguageServer/Impl/Implementation/AnalysisProgressReporter.cs
@@ -15,18 +15,18 @@
 // permissions and limitations under the License.
 
 using System;
-using System.Collections.Generic;
+using System.Threading.Tasks;
 using Microsoft.PythonTools.Analysis.Infrastructure;
 
 namespace Microsoft.Python.LanguageServer.Implementation {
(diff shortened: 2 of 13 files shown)


src/LanguageServer/Impl/Definitions/ILogger.cs
```diff
@@ -19,5 +19,6 @@
 namespace Microsoft.Python.LanguageServer {
     public interface ILogger {
         void TraceMessage(IFormattable message);
+        void TraceMessage(string message);
```
AlexanderSher: This has an issue of creating strings (lots of strings) even when logging is switched off or doesn't meet the required severity. It creates noticeable performance impact. If we need to get formatted strings from resources and can't use `IFormattable`, we can add `TraceMessageFormat(string format, params string[] parameters)`.

MikhailArkhipov: OK

src/LanguageServer/Impl/Resources.resx
```diff
@@ -117,6 +117,12 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="AnalysisProgress_MultipleItemsRemaining" xml:space="preserve">
```
AlexanderSher: If those strings go directly to the client, how do we pick up the culture?

MikhailArkhipov: It is to be determined.... Unfortunately, LSP does not specify locale, so we'll have to add some custom notification. Which reminds me to file https://github.com/Microsoft/python-language-server/issues/98

AlexanderSher: Can't we have those strings on VSC side and provide only numbers?

MikhailArkhipov: VSC is not the only client. LSP is public and supported by Eclipse, Vim, Emacs and more. So there is actually a plan to try LS with one of those.

AlexanderSher: Ok, but isn't this part of the LS protocol is python specific and we will have to add something to the client-side anyway?

MikhailArkhipov: Something like `python/setLocale`. If client doesn't send it, we'll just use English. Also, clients typically expect string in exceptions which they display in UI. We throw with descriptive message when we cannot rename, for example.

src/LanguageServer/Impl/Implementation/AnalysisProgressReporter.cs
```diff
@@ -0,0 +1,85 @@
...
+                    _logger.TraceMessage($"Analysis completed for {e.uri} that is not in the dictionary.");
+                }
+                UpdateProgressMessage();
+            }
+        }
+
+        private void UpdateProgressMessage() {
+            if(_activeAnalysis.Count > 0) {
+                _progress = _progress ?? _progressService.BeginProgress();
+                _progress.Report(_activeAnalysis.Count == 1
```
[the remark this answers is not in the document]
MikhailArkhipov: Yeah. This is how progresss is reported in VSC - via spinner and message in the status bar. I am not 100% about count but since analysis is recursive and event may fire from threads, I better be safe

src/LanguageServer/Impl/Implementation/AnalysisProgressReporter.cs
```diff
@@ -0,0 +1,85 @@
...
+                    _activeAnalysis[e.uri] = 1;
+                }
+                UpdateProgressMessage();
+            }
+        }
+        private void OnAnalysisComplete(object sender, AnalysisCompleteEventArgs e) {
+            lock (_lock) {
+                if (_activeAnalysis.TryGetValue(e.uri, out var count)) {
+                    if (count > 1) {
+                        _activeAnalysis[e.uri]--;
```
AlexanderSher: `PythonAnalyzer.AnalyzeQueuedEntries` may exit before raising `AnalysisComplete`. I'm not sure these events are actually guaranteed to be raised in pairs.

MikhailArkhipov: Interesting. This may explain stale messages in the status bar. I will check it out.

MikhailArkhipov: OK then, back to monitoring queue. The important part is to make sure we do NOT display progress when nothing is being analyzed.

sneridagh:
Can we have a release? The problem that this fix is very annoying...

MikhailArkhipov:
@sneridagh - try today's insiders build, see https://github.com/Microsoft/vscode-python/blob/master/CONTRIBUTING.md#development-build

fgrcon:
@MikhailArkhipov  - i installed the insider beta right now, but it does not help - still get stuck with "Analyzing workspace, 3467 items left" number increasing .....

sneridagh:
using 20/9 build:
Same behavior, the issue persists.
One CPU locked, fans spinning, Memory usage 2GB.

sneridagh:
@MikhailArkhipov any insight?

MikhailArkhipov:
@sneridagh - Unfortunately, new builds are not available yet, we are changing build process in order to better support dynamic updates of the LS in the VS Code extension. You can try and build it yourself or wait couple of days until we settle on URLs and versioning.

fgrcon:
@MikhailArkhipov  where can I dowload an older (last stable) extension?

neilsustc:
Hope it can be solved soon 🥺

@fgrcon Maybe try uninstalling the current one, restarting vscode, and then installing from the marketplace.

fgrcon:
> Hope it can be solved soon 🥺
> 
> @fgrcon Maybe try uninstalling the current one, restarting vscode, and then installing from the marketplace.

@neilsustc : that'is what I tried, but i have no clue  where to find previous releases ...

sneridagh:
@MikhailArkhipov Thanks for your patience and support... but somehow I'm unable to build it on MACOS. I even tried to follow the steps by hand but multiple points of failure... The build process just don't builds on it. I cannot spend more time on this, I will wait until October... it's a pity that the release process is that tight and cannot be bent when it's required. Something really uncommon in OSS.

What it really concerns me is that we haven't been able to test the patch properly... what if it doesn't work? We will have to wait until November and hope that all is fine?

Also, I don't understand yet the source of the problem. There's a setting that states clearly:
```
  // Only show errors and warnings for open files rather than for the entire workspace.
  "python.analysis.openFilesOnly": true,
```

I don't want my entire workspace to be analyzed, and it seems that it's what is happening. Is this setting working as expected?

By the way, I workaround it by (literally)

`$ rm -rf ~/.vscode/extensions/ms-python.python-2018.8.0/languageServer`

and I have at the end the same outcome but without my CPU spinning like crazy during all day.

DonJayamanne:
@sneridagh 	
If you would like to test the Python extension with the latest version of the language server then please:
* Install the development version of the Python Extension from here https://pvsc.blob.core.windows.net/extension-builds/ms-python-insiders.vsix
* @MikhailArkhipov I don't know what version of the language server they need.
* The version of Language Server that will get downloaded by the dev version of the extension is `0.1.24`

sneridagh:
@DonJayamanne I already tried this build (2018.8.1-beta (20 September 2018) and as confirmed by @MikhailArkhipov, it doesn't contain the fix or since it doesn't have changelog entry, it doesn't work.

neilsustc:
@DonJayamanne Sorry for digression. Could you please also publish (or package) a new version of your vscodeJupyter extension? Some PRs were merge many monthes ago but we still cannot benefit from them 😥 now.
(I tried to package it by myself but failed for knowing little about webpack and some package version conflicts(?))

MikhailArkhipov:
20th was still an old build. We published update on 28th.

fgrcon:
@MikhailArkhipov interesting: you are able to publish in the future ? ;-) 
https://pvsc.blob.core.windows.net/extension-builds/ms-python-insiders.vsix is still the same from 9/20!

MikhailArkhipov:
@fgrcon - Right but it downloads proper LS. VSIX itself was not an issue, content of the storage blob was. I guess it was 27th. :-)

sneridagh:
@MikhailArkhipov FYI I was giving it another try, then the process blew and rebooted my mac in my face.

AlexanderSher:
Is it a publicly available source code?

MikhailArkhipov:
With huge workspace it is a different problem. The bug was that analysis *reporting* was stuck, i.e. reporting items outstanding while actually there were none. With huge projects analysis may take a long time. That does not mean though that product is not usable. You don't have to wait for everything to complete.

sneridagh:
@MikhailArkhipov it gets stuck as well, in fact it's a combination of both factors. The analysis, being huge, never ends, the process eats all the memory available. The message in the status bar "X items remaining" doesn't go away.

TomFaulkner:
While it isn't unusable, the whole computer slows to a crawl while the Python language server is analyzing. A 2016 MacBook Pro has the Microsoft Python Language Server taking nearly two of the four cores to itself, and half of the 16gb of RAM the system has.

sneridagh:
@MikhailArkhipov @DonJayamanne Please provide a setting for disable **completely** both Jedi and languageServer analyzers. I prefer not to have Intellisense at all that having my CPU fans blowing continuously. 

@TomFaulkner it works for a while... until then it starts again (even after some hours).

@MikhailArkhipov now it does it even partitioning my workspace into smaller parts (not loading the big ones, eggs sources). It stays calmed, until it begins again. Just killed it manually (the languageServer process) and it stays calm again for a while. Don't know why it causes to start to get crazy again.

TomFaulkner:
@sneridagh, you can remove the Python extension, or disable it. That is what I did prior to searching out the issue and using the insider build above. (Which is so far working.)

TomFaulkner:
Sadly, yes, that is the impact.

fgrcon:
sadly giving up!
 the latest release still eats up the whole memory within minutes and everything gets stuck.

how to get back to working release (i think early summer this year )  - I need debugging, linting  & outline  but can live without  greedy intellisense, if there is no other way.

TomFaulkner:
> sadly giving up!
> the latest release still eats up the whole memory within minutes and everything gets stuck.
> 
> how to get back to working release (i think early summer this year ) - I need debugging, linting & outline but can live without greedy intellisense, if there is no other way.

@fgrcon, I was using the beta release to get around this issue. Then after an update it got replaced and I was back to this problem. However, I noticed there was an update available for the extension, and that update resolved things. I hope this helps you, I was ready to switch as well.

sneridagh:
@TomFaulkner @fgrcon About to give up as well... Returning either to Atom or Sublime. @fgrcon where did you get the update? I'm on 2018.10.1 (09 Nov 2018) (it seems).

/cc @MikhailArkhipov Again, please provide a way to disable intellisense (either via Jedi or PLS) in the plugin, for those that we don't care about it, but we do about the other things the plugin provides. Thanks.

PR #100, opened 2018-09-17
More test fixes and related bug fixes

AlexanderSher:
- Make PriorityProducerConsumer clear its queues on dispose
- Raise AnalysisQueue.AnalysisComplete at the end of consumer loop
- Derive IModule from IAnalysisValue
- Add more assertions
- Fix more tests

### DIFF
--- a/src/Analysis/Engine/Impl/Infrastructure/PriorityProducerConsumer.cs
+++ b/src/Analysis/Engine/Impl/Infrastructure/PriorityProducerConsumer.cs
@@ -58,7 +58,17 @@ namespace Microsoft.PythonTools.Analysis.Infrastructure {
             _comparer = comparer ?? EqualityComparer<T>.Default;
         }
 
-        public void Dispose() => _disposeToken.TryMarkDisposed();
+        public void Dispose() {
+            if (_disposeToken.TryMarkDisposed()) {
+                lock (_syncObj) {
+                    Debug.Assert(_pendingTasks.Count == 0 || _firstAvailablePriority == _maxPriority);
+                    _pendingTasks.Clear();
+                    foreach (var queue in _queues) {
+                        queue.Clear();
+                    }
+                }
+            };
+        }
 
         public void Produce(T value, int priority = 0) {
             if (priority < 0 || priority >= _maxPriority) throw new ArgumentOutOfRangeException(nameof(priority));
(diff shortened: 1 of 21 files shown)


src/Analysis/Engine/Impl/Infrastructure/PriorityProducerConsumer.cs
```diff
@@ -58,7 +58,17 @@ internal sealed class PriorityProducerConsumer<T> : IDisposable {
             _comparer = comparer ?? EqualityComparer<T>.Default;
         }
 
-        public void Dispose() => _disposeToken.TryMarkDisposed();
+        public void Dispose() {
+            if (_disposeToken.TryMarkDisposed()) {
+                lock (_syncObj) {
+                    Debug.Assert(_pendingTasks.Count == 0 || _firstAvailablePriority == _maxPriority);
+                    _pendingTasks.Clear();
```
MikhailArkhipov: Should we cancel `pendingTcs`?

src/Analysis/Engine/Impl/Infrastructure/PriorityProducerConsumer.cs
```diff
@@ -58,7 +58,17 @@ internal sealed class PriorityProducerConsumer<T> : IDisposable {
...
+                    foreach (var queue in _queues) {
+                        queue.Clear();
+                    }
+                }
+            };
+        }
 
         public void Produce(T value, int priority = 0) {
             if (priority < 0 || priority >= _maxPriority) throw new ArgumentOutOfRangeException(nameof(priority));
```
MikhailArkhipov: If you want, you can use `Check` that is now in infra

src/Analysis/Engine/Impl/Infrastructure/PriorityProducerConsumer.cs
AlexanderSher: They are canceled by `_disposeToken` itself:
https://github.com/Microsoft/python-language-server/blob/1d0c6244463ba740896ca78cc444f334b97454c0/src/Analysis/Engine/Impl/Infrastructure/PriorityProducerConsumer.cs#L135

src/Analysis/Engine/Impl/Infrastructure/PriorityProducerConsumer.cs
AlexanderSher: Good to know!

PR #103, opened 2018-09-18
Add ability to get document content from API

MikhailArkhipov:
And few more useful things to extensibility interfaces

### DIFF
--- a/src/Analysis/Engine/Impl/ProjectEntry.cs
+++ b/src/Analysis/Engine/Impl/ProjectEntry.cs
@@ -320,6 +320,7 @@ namespace Microsoft.PythonTools.Analysis {
 
         public IGroupableAnalysisProject AnalysisGroup => ProjectState;
 
+        #region IPythonProjectEntry
         public IModuleAnalysis Analysis { get; private set; }
 
         public string FilePath { get; }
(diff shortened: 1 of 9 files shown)


src/Analysis/Engine/Impl/ProjectEntry.cs
```diff
@@ -340,6 +341,9 @@ where fileName.StartsWithOrdinal(pathPrefix, ignoreCase: true)
 
         public Dictionary<object, object> Properties { get; } = new Dictionary<object, object>();
 
+        public IDocument Document => this as IDocument;
```
AlexanderSher: `public IDocument Document => this` doesn't work?

src/Analysis/Engine/Test/AnalysisTest.cs
```diff
@@ -4403,7 +4403,7 @@ class oar(list):
                 var completions = await server.SendCompletion(uri, 2, 8);
 
                 completions.Should().HaveItem("append")
-                    .Which.Should().HaveInsertText("append(self, value):\r\n\treturn super(oar, self).append(value)");
+                    .Which.Should().HaveInsertText("append(self, value):\r\n    return super(oar, self).append(value)");
```
AlexanderSher: How do we determine if it is a `\t` or a set of spaces? Is there a setting for it?

MikhailArkhipov: Per `GetCompletionsForOverride` it always uses spaces. I believe per PEP spaces are strongly preferred.

PR #115, opened 2018-09-20
Port PR from PTVS: https://github.com/Microsoft/PTVS/pull/4716

AlexanderSher:
From @lostmsu:

This change fixes https://github.com/Microsoft/PTVS/pull/4715

Rationale:
If class A(B), and both have method foo, then any variable v of type B can potentially hold value of type A. Which means, that v.foo() can return a value of type that of return value of A.foo or B.foo.

To achieve that, when function return type is determined by DDG, it is also added as a potential return type for all of the function's bases (e.g. functions, that this function overrides).

In addition to that, this change also improves DDG.LookupBaseMethods, which previously only took into account built-in classes.


lostmsu:
Should this repository be used instead of PTVS?

PR #118, opened 2018-09-20
Fix References_Parameters_NestedFunction

### LINKED ISSUE #39: References_Parameters_NestedFunction test fails

MikhailArkhipov:
Fixes #39

### DIFF
--- a/src/Analysis/Engine/Impl/ModuleAnalysis.cs
+++ b/src/Analysis/Engine/Impl/ModuleAnalysis.cs
@@ -229,14 +229,14 @@ namespace Microsoft.PythonTools.Analysis {
             }
 
             if (expr is NameExpression name) {
-                var defScope = scope.EnumerateTowardsGlobal.FirstOrDefault(s =>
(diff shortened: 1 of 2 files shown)


src/Analysis/Engine/Impl/ModuleAnalysis.cs
```diff
@@ -229,14 +229,14 @@ private class DefinitionWalker : PythonWalkerWithLocation {
             }
 
             if (expr is NameExpression name) {
-                var defScope = scope.EnumerateTowardsGlobal.FirstOrDefault(s =>
-                    s.ContainsVariable(name.Name) && (s == scope || s.VisibleToChildren || IsFirstLineOfFunction(scope, s, location)));
-
-                if (defScope == null) {
-                    variables = _unit.State.BuiltinModule.GetDefinitions(name.Name)
-                        .SelectMany(ToVariables);
+                if (!scope.EnumerateTowardsGlobal.Any()) {
```
AlexanderSher: We don't need to call `EnumerateTowardsGlobal.Any()`, it should be enough to check after the `foreach` if `variables` is empty.

PR #122, opened 2018-09-22
use correct case for build directory

dimbleby:
trying to compile on a non-Windows platform...


dimbleby:
Sorry to be awkward, but figuring out whether it's OK for me to sign this CLA is going to be more trouble than it's worth for such a trivial change.  

Take it or leave it.

MikhailArkhipov:
Yes, Linux is case sensitive. Thanks for your fix!

PR #126, opened 2018-09-24
On type formatting

jakebailey:
This PR implements line formatting via on-type formatting, triggering on newline and semicolon. It only tokenizes the code to be formatted, and does not require parsing.

This covers all of the existing formatOnType issues in vscode-python (Microsoft/vscode-python#1799, Microsoft/vscode-python#1784, Microsoft/vscode-python#1792), as well as some other cases that weren't being handled already (some multiline stuff, commas in brackets, all of the different types of unpacking, more).

Slicing follows the PEP8 pet peeves example list by using a simple heuristic to decide when to insert a space. If the thing to the left/right of a colon is "simple" (a single identifier/constant, with or without a unary operator), then no space is inserted. The heuristic also checks for commas to handle multi-index slicing.

Tests have been copied from vscode-python, with some minor modifications for PEP8 correctness. Extra tests have been added for the new fixes. 

Some issues/possible improvements:

- TextBuilder is in Analysis.Engine.Infrastructure, but the rest of the formatting code is in LanguageServer.Implementation. This is probably wrong.
- The heuristic used for spacing out slicing is pretty simple, and not the same one used by tools like black. Using an existing algorithm may be preferred to prevent situations where you hit enter, get one thing, format the document, get another, hit enter again and go back, and so on. Perhaps it would be even better to outsource line formatting to external tools if they're enabled for (upcoming) full-document formatting.
- The existing tokenizer does not capture the line continuation token if there's no newline after the `\`. Since the token is never captured, the formatter won't know it's there and will not insert it. This issue isn't hit in the editor when hitting enter after the `\`, since the formatter is called after the newline gets inserted. The only way to trigger this issue from the editor is to hit `;` when there is no next line, or to somehow invoke a DocumentOnTypeFormatter RPC call directly without inserting the newline.
    - The easiest fix would be to just append a newline to the end of the incoming TextReader. Since the formatter only works with tokens, and by request, there's no way this extra newline would ever make it back into the output.
    - Alternatively, the tokenizer could be modified to treat the `\` token separately from the following newline, but I haven't investigated the implications of that change.
- Unknown tokens will currently just have spaces put around them. It may be better to throw an execption or spit out a warning.
- This formatter could be converted into a makeshift full-document formatter by formatting each line and returning the edits (or being smarter and combining them into a single edit). It might be an acceptable default when no other document formatter exists.


src/Analysis/Engine/Impl/Infrastructure/TextBuilder.cs
```diff
@@ -0,0 +1,59 @@
...
+            if (IsLastWhitespace()) {
+                // Don't mutate when producing the string.
+                seq = _segments.Take(_segments.Count - 1);
+            }
+            return string.Concat(seq);
+        }
+
+        public void Append(string s) => _segments.Add(s);
+
+        public void Append(object o) => _segments.Add(o.ToString());
```
MikhailArkhipov: I think string is kind of object, so string-specific one may not be necessary

src/Analysis/Engine/Test/LanguageServerTests.cs
```diff
@@ -992,6 +992,19 @@ class C:
             );
         }
 
+        [TestMethod, Priority(0)]
+        public async Task OnTypeFormatting() {
+            var s = await CreateServer();
```
AlexanderSher: Don't forget to dispose server object at the end of the test.

src/Analysis/Engine/Impl/Infrastructure/TextBuilder.cs
```diff
@@ -0,0 +1,59 @@
...
+            if (_segments.Count == 0 && !allowLeading) {
+                return;
+            }
+
+            if (IsLastWhitespace()) {
+                count--;
+            }
+
+            for (var i = 0; i < count; i++) {
+                _segments.Add(" ");
```
MikhailArkhipov: `.Add(new string(' ', count))` ?

src/Analysis/Engine/Test/LanguageServerTests.cs
jakebailey: I can do this, but no other tests in this file dispose of the server either (which is why I didn't think to do so).

src/Analysis/Engine/Test/LanguageServerTests.cs
```diff
@@ -992,6 +992,19 @@ class C:
             );
         }
```
MikhailArkhipov: This file is getting huge. Can we create LanguageServer folder and then separate FormattingTests file?

src/Analysis/Engine/Test/LanguageServerTests.cs
MikhailArkhipov:
```cs
using(var s = CreateServer()) {
}
```
since it is disposable

jakebailey: Mentioned in the other thread on this line, I can do this, but no other test in this file actually disposes of the server either. I can submit a second PR to mass-dispose if you'd like.

MikhailArkhipov: Those are old tests we do need to fix at some point :-)

src/Analysis/Engine/Test/LanguageServerTests.cs
```diff
@@ -992,6 +992,19 @@ class C:
...
 
+        [TestMethod, Priority(0)]
+        public async Task OnTypeFormatting() {
+            var s = await CreateServer();
+
+            var mod = await AddModule(s, "def foo  ( ) :\n    x = a + b\n    x+= 1");
```
AlexanderSher: One of the good rules about the test is that the person who sees the test for the first time (but familiar with the code base) should be able to understand what's going on. 

Here, without digging into implementation, I can only understand that you have created and initialized a `Server` instance and added a ProjectEntry to it with the content above. The rest is unclear.

There is no problem for a test to have some more lines of code, but it is important for a test to be as clear as possible.

src/Analysis/Engine/Test/LanguageServerTests.cs
```diff
@@ -1185,6 +1198,22 @@ private class GetAllExtension : ILanguageServerExtension {
             refs.Select(r => $"{r._kind ?? ReferenceKind.Reference};{r.range}").Should().Contain(contains).And.NotContain(excludes);
         }
 
+        public static async Task AssertLineFormat(Server s, TextDocumentIdentifier document, int line, string ch, string newText, SourceLocation start, SourceLocation end) {
+            var edits = await s.DocumentOnTypeFormatting(new DocumentOnTypeFormattingParams {
```
AlexanderSher: We have `ServerExtension` type that accumulates helper methods that help with simulating server activity. You can add `SendDocumentOnTypeFormatting(Uri, Position, char)` helper for it and call it directly from the test.

src/Analysis/Engine/Test/LanguageServerTests.cs
jakebailey: Can you be more specific as to how you'd prefer it look? I'm not sure how this is unclear (or at least more unclear than tests like the Hover tests above it).

MikhailArkhipov: I personally would write most tests for the simplest object - line formatter itself and add just few that involve the LS itself as to verify the integration. This helps if someone wants to borrow the formatter but doesn't need the LS - tests then are easy to take with the code.

src/LanguageServer/Impl/Implementation/LineFormatter.cs
```diff
@@ -0,0 +1,643 @@
...
+        /// stored in _lineTokens. If the provided line number is past the
+        /// end of the input text, then the tokenizer will stop.
+        /// Additionally, this function will attempt to read ahead onto the
+        /// next line to the first non-ignored token so that the formatter
+        /// can look ahead.
+        /// </summary>
+        /// <param name="line">One-indexed line number.</param>
+        /// <returns>A non-null list of tokens on that line.</returns>
+        private List<TokenExt> TokenizeLine(int line) {
+            if (line < 1) {
```
MikhailArkhipov: You can use `Check` class to simplify arguments check

src/Analysis/Engine/Test/LanguageServerTests.cs
jakebailey: I think that's already how I have it set up. LineFormatterTests contains all of the test cases for the formatter itself. This test is just checking that the server does something. If you mean that I should remove any actual stylistic choices from here and do a simple check of "does the server return an edit instead of throwing NotImplemented", then I can do that too.

src/LanguageServer/Impl/Implementation/LineFormatter.cs
```diff
@@ -0,0 +1,643 @@
...
+        /// <returns>A list of TextEdits needed to format the line.</returns>
+        public TextEdit[] FormatLine(int line) {
+            if (line < 1) {
+                return NoEdits;
+            }
+
+            var tokens = TokenizeLine(line);
+
+            // Keep ExplictLineJoin because it has text associated with it.
+            tokens = tokens.Where(t => !t.IsIgnored || t.Kind == TokenKind.ExplicitLineJoin).ToList();
```
MikhailArkhipov: Looks like we build list, then insta-filter it. Perhaps filter out at creation time in order to reduce duplication? Like pass a filtering lambda to `TokenizeLine`?

src/LanguageServer/Impl/Implementation/LineFormatter.cs
```diff
@@ -0,0 +1,643 @@
...
+                    }
+
+                    return (a.IsNumber || a.Kind == TokenKind.Name) && b.Is(TokenKind.RightBracket, TokenKind.Colon, TokenKind.Comma);
+                }
+            }
+
+            public TokenExt PrevNonIgnored
+            {
+                get
+                {
```
MikhailArkhipov: Formatting

jakebailey: The formatting settings you gave me for VS did this. Do you have an updated set of formatting rules I can use?

src/Analysis/Engine/Test/LineFormatterTests.cs
```diff
@@ -0,0 +1,460 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using TestUtilities;
+
+namespace AnalysisTests {
+    [TestClass]
+    public class LineFormatterTests {
+        [TestMethod, Priority(0)]
+        public void LineOutOfBounds() {
+            AssertNoEdits("a+b", line: 0);
+            AssertNoEdits("a+b", line: -1);
```
AlexanderSher: With `DataTestMethodAttribute`, you can have parametrized tests which allows us to get separate result for each case.

src/LanguageServer/Impl/Implementation/LineFormatter.cs
MikhailArkhipov: We should probably update and add settings file to the PR

src/Analysis/Engine/Test/LineFormatterTests.cs
MikhailArkhipov: Jake may need an example :-)

src/Analysis/Engine/Test/LineFormatterTests.cs
```diff
@@ -0,0 +1,460 @@
...
+using Microsoft.PythonTools.Analysis;
+using Microsoft.PythonTools.Analysis.FluentAssertions;
+using Microsoft.PythonTools.Parsing;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using TestUtilities;
+
+namespace AnalysisTests {
+    [TestClass]
+    public class LineFormatterTests {
+        [TestMethod, Priority(0)]
```
AlexanderSher: Please add 
```cs
        public TestContext TestContext { get; set; }

        [TestInitialize]
        public void TestInitialize() {
            TestEnvironmentImpl.TestInitialize($"{TestContext.FullyQualifiedTestClassName}.{TestContext.TestName}");
        }

        [TestCleanup]
        public void TestCleanup() {
            TestEnvironmentImpl.TestCleanup();
        }
```

to the test class so we can have monitoring of `DoNotWait` tasks and proper folder for test artifacts. I know it's ugly, but MSTest doesn't give us any better way of doing it.

src/Analysis/Engine/Test/LineFormatterTests.cs
```diff
@@ -0,0 +1,460 @@
...
+
+            string fileContents;
+            using (var reader = new StreamReader(src, true)) {
+                fileContents = reader.ReadToEnd();
+            }
+
+            var lines = fileContents.Split(new[] { "\r\n", "\r", "\n" }, StringSplitOptions.None);
+
+            using (var reader = new StringReader(fileContents)) {
+                var lineFormatter = new LineFormatter(reader, PythonLanguageVersion.V37);
```
AlexanderSher: Can we use `PythonVersions.LatestAvailable3X`, or that code works specifically with `V37`?

src/LanguageServer/Impl/Implementation/LineFormatter.cs
jakebailey: Perhaps, but I'd be really wary of this since those newline tokens can contain the preceeding whitespace strings which are useful when working out how to edit things. Right now, I've been able to eliminate that need by using the token spans, but it may make future improvements more difficult.

src/Analysis/Engine/Test/LineFormatterTests.cs
AlexanderSher: `AnalysisTest.PackageRelativeImportPep328`

src/Analysis/Engine/Test/LineFormatterTests.cs
jakebailey: There's nothing special about V37, but I need a `PythonLanguageVersion` (the enum). LatestAvailable3X is an `InterpretorConfiguration` which does not contain an enum value.

src/Analysis/Engine/Test/LanguageServerTests.cs
jakebailey: Closing this in favor of the other one.

src/Analysis/Engine/Test/LineFormatterTests.cs
AlexanderSher: Ok, do you actually need python to be installed for these tests?

src/Analysis/Engine/Test/LanguageServerTests.cs
jakebailey: I fixed it in this PR, and I can probably go do it in another PR for the rest of them (and move the tests out of this giant file).

src/Analysis/Engine/Test/LineFormatterTests.cs
MikhailArkhipov: AFAIK this setting is to tokenizer to know which language syntax to use

AlexanderSher: Ok, ignore the comment.

src/LanguageServer/Impl/Implementation/LineFormatter.cs
jakebailey: @AlexanderSher Any preference for this being `Func<TokenExt, bool>` vs `Predicate<TokenExt>`?

AlexanderSher: Haven't made it to that line yet :)

src/LanguageServer/Impl/Implementation/LineFormatter.cs
```diff
@@ -0,0 +1,631 @@
...
+            return new[] { edit };
+        }
+
+        private class TokenExt {
+            public Token Token { get; set; }
+            public SourceSpan Span { get; set; }
+            public int Line => Span.End.Line;
+            public TokenExt Inside { get; set; }
+            public TokenExt Prev { get; set; }
+            public TokenExt Next { get; set; }
```
AlexanderSher: So `TokenExt` forms a linked list of tokens. How many items it will have on our testing `pythonGrammar.py`?

src/LanguageServer/Impl/Implementation/LineFormatter.cs
jakebailey: A bit of searching says that Func/Action are the preferred (used in LINQ and others), and Predicate was the original that came with List. I'll probably use Func if there's no objection.

src/LanguageServer/Impl/Implementation/LineFormatter.cs
```diff
@@ -0,0 +1,631 @@
...
+
+            return tokens;
+        }
+
+        /// <summary>
+        /// Formats a single line and returns TextEdits to replace the old text.
+        /// </summary>
+        /// <param name="line">One-indexed line number.</param>
+        /// <returns>A list of TextEdits needed to format the line.</returns>
+        public TextEdit[] FormatLine(int line) {
```
AlexanderSher: This method is the only public one. Is it assumed that it can be called only once?

src/LanguageServer/Impl/Implementation/LineFormatter.cs
jakebailey: I did a quick modification of the wrapper to count, and the grammar file is 11278 tokens. In the common case, any sort of line formatting would only tokenize up to (and a bit past) the line to be formatted, so usually you wouldn't have this many at once. I'm not sure how much it matters given that the tokenizer essentially reads the whole file into memory. Mikhail said it would even be okay for me to read the whole file into memory (to figure out line lengths) instead of looking at spans to figure out what to replace.

There's definitely a way that I could write this to only look back so far (so that unneeded tokens can get thrown away), but I chose not to do it so that the LineFormatter could potentially be reused and to simplify the initial implementation.

src/Analysis/Engine/Test/LineFormatterTests.cs
jakebailey: That looks really neat. I'll give it a go.

src/Analysis/Engine/Test/LineFormatterTests.cs
```diff
@@ -0,0 +1,472 @@
...
+                    var lineTextOrig = lines[i];
+
+                    foreach (var edit in edits) {
+                        edit.range.start.line.Should().Be(i);
+                        edit.range.end.line.Should().Be(i);
+
+                        lineText = ApplyLineEdit(lineText, edit);
+                    }
+
+                    lineText.Should().Be(lineTextOrig, $"because line {lineNum} should be unchanged");
```
AlexanderSher: `ApplyLineEdit` doesn't refer any product code. What's the purpose of this assert?

jakebailey: I'm not sure what you mean. `ApplyLineEdit` applies a `TextEdit` to a single line of text and returns the edited string. I use it to apply the formatter's edit to make sure that it didn't change anything about the text.

src/LanguageServer/Impl/Implementation/LineFormatter.cs
MikhailArkhipov: Unfortunately, looking back has to start from the beginning. This is somewhat akin to [Kanji backspace](http://discuss.joelonsoftware.com/default.asp?interview.11.334807.4) problem

src/Analysis/Engine/Test/LineFormatterTests.cs
jakebailey: Done.

src/Analysis/Engine/Impl/Infrastructure/TextBuilder.cs
jakebailey: That case doesn't actually break, so I've made the change.

src/Analysis/Engine/Impl/Infrastructure/TextBuilder.cs
jakebailey: I removed the string specialization.

MikhailArkhipov:
LGTM pending @AlexanderSher approval

src/Analysis/Engine/Test/LineFormatterTests.cs
AlexanderSher: Ok, is it different from 
```cs
foreach (var edit in edits) {
    var start = edit.range.start;
    var end = edit.range.end;
    start.line.Should().Be(i);
    end.line.Should().Be(i);
    edit.newText.Should().Be(lineText.Substring(start.character, end.character - start.character - 1));
}
```

jakebailey: Yes, it's different than that. You couldn't check that each edit works, since they'd technically be applied by the editor in order, and the final result is what should actually be unchanged, but in my case it would probably be more correct to assert that either there are no edits (so therefore no change), or that there's a single edit which doesn't change the line. I could also do something like:

```csharp
var edits = lineFormatter.FormatLine(lineNum);
edits.Should().NotBeNull();

if (edits.Length == 0) {
    continue;
}

edits.Length.Should().Be(1);
var edit = edits[0];

var start = edit.range.start;
var end = edit.range.end;

start.line.Should().Be(i);
end.line.Should().Be(i);

var startIndex = start.character;
var removeCount = end.character - start.character - 1;

var lineText = lines[i];
var newLineText = lineText.Remove(startIndex, removeCount).Insert(startIndex, edit.newText);
newLineText.Should().Be(lineText, $"because line {lineNum} should be unchanged");
```

Which would do the same thing but avoids the helper.

AlexanderSher: Is it to show the whole string in the error message?

jakebailey: You're totally right. I hadn't thought of that. I'll commit the changes.

jakebailey: Showing the whole string isn't super important, since the only thing that would be cut off is whitespace (which currently does not get replaced, but modifying the formatter to replace it would not break this method of testing).

src/Analysis/Engine/Test/LanguageServerTests.cs
jakebailey: Would you like me to do that in this PR, or a subsequent one? It might be more straightforward to make a PR to move all of the tests at once, since a lot of the "CreateServer" logic is shared between the tests.

src/Analysis/Engine/Test/LanguageServerTests.cs
jakebailey: I implemented this, if you'd like to take a look.

src/Analysis/Engine/Test/LanguageServerTests.cs
jakebailey: I modified this test using the suggested added extension method to be more explicit, if you'd like to look again.

src/Analysis/Engine/Impl/Infrastructure/TextBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.PythonTools.Analysis.Infrastructure {
+    internal class TextBuilder {
```
AlexanderSher: @jakebailey, what is the benefit of having this class instead of common `StringBuilder`? `SoftAppendSpace` is specific for `LineFormatter`, and while there is no built-in method to find the last non-whitespace character, it is very simple to implement.

src/Analysis/Engine/Impl/Infrastructure/TextBuilder.cs
```diff
@@ -0,0 +1,55 @@
...
+        /// </summary>
+        /// <param name="count">How many spaces to append.</param>
+        /// <param name="allowLeading">True if spacing can be appended as the first item.</param>
+        public void SoftAppendSpace(int count = 1, bool allowLeading = false) {
+            if (_segments.Count == 0 && !allowLeading) {
+                return;
+            }
+
+            if (IsLastWhitespace()) {
+                count--;
```
AlexanderSher: Do we care about the length of the last whitespace segment?

src/Analysis/Engine/Test/LanguageServerTests.cs
```diff
@@ -992,6 +992,42 @@ class C:
             );
         }
 
+        [TestMethod, Priority(0)]
+        public async Task OnTypeFormatting() {
```
AlexanderSher: 👍

src/Analysis/Engine/Impl/Infrastructure/TextBuilder.cs
jakebailey: This class is pulled from the vscode-python extension. I need this sort of class in order to be able to insert spaces without duplicating them (unless that is needed). I originally avoided StringBuilder because I thought I would need to invoke ToString repeatedly to figure out how the builder ends, but looking at StringBuilderExtensions, I guess you can just index into the builder and let it figure it out.

If you'd prefer, I could probably write an extension that does something along the lines of ensuring there's some number of spaces at the end of a builder (and just trim it at the end) and remove the TextBuilder.

src/Analysis/Engine/Impl/Infrastructure/TextBuilder.cs
jakebailey: That would be more correct, yes. In the context of how the builder is currently used, I avoid this sort of issue because I only ever insert single spaces (except before a comment).

jakebailey: If I remove TextBuilder in favor of an extension method onto StringBuilder, I can likely make it a bit smarter than just assuming that the end is a single space and decrementing count.

src/Analysis/Engine/Impl/Infrastructure/TextBuilder.cs
jakebailey: I've written a single extension method for StringBuilder that replaces TextBuilder entirely:
```csharp
        public static StringBuilder EnsureEndsWithSpace(this StringBuilder sb, int count = 1, bool allowLeading = false) {
            if (sb.Length == 0 && !allowLeading) {
                return sb;
            }

            for (var i = sb.Length - 1; i >= 0; i--) {
                if (!char.IsWhiteSpace(sb[i])) {
                    break;
                }
                count--;
            }

            if (count > 0) {
                sb.Append(new string(' ', count));
            }

            return sb;
        }
```
If there's no objection I can remove TextBuilder and just use this.

MikhailArkhipov: OK. In RTVS and other places I used it in it was handling quite a bit more than just trailing ws - such as type of indentation, conversion of spaces to tabs and vice versa, etc. SB would be very limiting there.

PR #131, opened 2018-09-25
A test to ensure @abc.abstractproperty return value is analyzed correctly

lostmsu:
This adds a test for the analysis of return values of abstract properties (properties, marked with `@abc.abstractproperty`).

This also adds a partial fix for it not being analyzed properly. (related to #115)

Unfortunately, due to some regression (?) between PTVS/master and python-language-server/master, this does not seem to be enough: test variable `b` analysis results in two possible return values, `int` (correct) and `A.virt` (which is incorrect, and eventually resolves to be `Unknown`). The later seems to be happening because on the first pass `ddg._eval.Evaluate(@abc.abstractproperty)` results in empty set.

### DIFF
--- a/src/Analysis/Engine/Impl/Values/SpecializedNamespace.cs
+++ b/src/Analysis/Engine/Impl/Values/SpecializedNamespace.cs
@@ -217,6 +217,15 @@ namespace Microsoft.PythonTools.Analysis.Values {
             }
         }
 
+        public override string Name {
+            get {
+                if (_original == null)
+                    return base.Name;
+
+                return _original.Name;
+            }
+        }
+
         public override IEnumerable<OverloadResult> Overloads {
             get {
                 if (_original == null) {
--- a/src/Analysis/Engine/Impl/Values/SpecializedNamespace.cs
+++ b/src/Analysis/Engine/Impl/Values/SpecializedNamespace.cs
@@ -217,14 +217,7 @@ namespace Microsoft.PythonTools.Analysis.Values {
             }
         }
 
-        public override string Name {
-            get {
-                if (_original == null)
-                    return base.Name;
-
-                return _original.Name;
-            }
-        }
+        public override string Name => _original == null ? base.Name : this._original.Name;
 
         public override IEnumerable<OverloadResult> Overloads {
             get {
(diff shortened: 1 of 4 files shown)


lostmsu:
Speaking of the later: what is the reasoning behind module's members being analyzed before overall module is? I'd expect `import` statements in a module to be analyzed before any class definitions in it.

src/Analysis/Engine/Test/Microsoft.Python.Analysis.Engine.Tests.csproj
```diff
@@ -1,7 +1,7 @@
 ﻿<Project>
   <PropertyGroup>
     <TargetFramework>netcoreapp2.1</TargetFramework>
-    <RootNamespace>Microsoft.PythonTools.Analysis</RootNamespace>
+    <RootNamespace>AnalysisTests</RootNamespace>
```
AlexanderSher: Let's not change namespaces for now.

src/Analysis/Engine/Impl/Values/SpecializedNamespace.cs
```diff
@@ -217,6 +217,15 @@ abstract class SpecializedNamespace : AnalysisValue {
             }
         }
 
+        public override string Name {
+            get {
```
AlexanderSher: This can be written with expression:
```cs
public override string Name => _original != null ? _original.Name : base.Name;
```

MikhailArkhipov: Probable even
```cs
public override string Name => _original?.Name ?? base.Name;
```

AlexanderSher: No, that's a bit different semantics. In original code, if `_original != null` and `_original.Name == null`, it will return `null`. In case of `_original?.Name ?? base.Name`, it will return `base.Name`. Can't say what is the right behavior though.

src/Analysis/Engine/Test/Microsoft.Python.Analysis.Engine.Tests.csproj
lostmsu: This only affects what namespace VS uses when you create a new file. It was set to AnalysisTests in PTVS but, I guess, was overlooked when porting, which in turn accidentally caused my new file InheritanceTests to get wrong namespace.

PR #144, opened 2018-09-26
Fix string concatenation spacing in line formatter

jakebailey:
I thought of this case while reading the fstring spec. The line formatter defaults to not spacing around string literals, but it's likely better to insert a space between two string literals next to each other.

For reference:
- autopep8 leaves spacing as-is, so `'a''b'` and `'a' 'b'` would not be changed (even with more than one space between). I could emulate this behavior in the line formatter, since `TokenExt` includes `PreceedingWhitespace`.
- yapf ensures there is a space between two strings.
- black behaves the same as yapf (barring its replacement of `'` with `"`).

### DIFF
--- a/src/LanguageServer/Impl/Implementation/LineFormatter.cs
+++ b/src/LanguageServer/Impl/Implementation/LineFormatter.cs
@@ -287,8 +287,15 @@ namespace Microsoft.Python.LanguageServer.Implementation {
                         builder.EnsureEndsWithSpace();
                         break;
 
-                    case TokenKind.Name:
                     case TokenKind.Constant:
+                        if (token.IsString && next != null && next.IsString) {
+                            builder.Append(token);
+                            builder.EnsureEndsWithSpace();
+                            break;
+                        }
+                        goto case TokenKind.Name;
+
+                    case TokenKind.Name:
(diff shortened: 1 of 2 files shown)


src/LanguageServer/Impl/Implementation/LineFormatter.cs
```diff
@@ -287,8 +287,15 @@ public class LineFormatter {
                         builder.EnsureEndsWithSpace();
                         break;
 
-                    case TokenKind.Name:
                     case TokenKind.Constant:
+                        if (token.IsString && next != null && next.IsString) {
```
AlexanderSher: Does `next == null` means that it is the last token?
Also, do you need to check if they are on the same line?

jakebailey: Yes, that `next` would be null if the current token were the last token on the line. That `next` is pulled from the list of tokens of the line, not the linked-list that's happening in `TokenExt` which could peer forward. I wouldn't want to look at the next line, since even if the next token (ignoring whitespace) happens to be a string literal, the fact that it's on another line implies that there's already spacing and adding a space here would only be adding trailing whitespace (which would then be trimmed away).

PR #146, opened 2018-09-27
Fix type annotation on overloads

MikhailArkhipov:
- Make type overloads use same way of generating parameter types as tooltip/hover. This way get get proper short types and not something like `a:T, int` on a generic or `a:list[float, float, float], list[float]`.
- Return value should not have `[ ]` around it if it is a single type.
- If function returns nothing, then display `None` and not `[ ]`
- Prettify documentation with proper spacing
- Update all TypingModuleDocumentation* tests baselines

### DIFF
--- a/src/Analysis/Engine/Impl/OverloadResult.cs
+++ b/src/Analysis/Engine/Impl/OverloadResult.cs
@@ -144,7 +144,7 @@ namespace Microsoft.PythonTools.Analysis {
         private string _name;
         private string _doc;
         private string[] _pnames;
-        private IAnalysisSet[] _ptypes;
+        private string[] _ptypes;
         private string[] _pdefaults;
         private readonly HashSet<string> _rtypes;
 
@@ -152,7 +152,7 @@ namespace Microsoft.PythonTools.Analysis {
             _name = name;
             _doc = documentation;
             _pnames = new string[parameters];
-            _ptypes = new IAnalysisSet[parameters];
+            _ptypes = new string[parameters];
             _pdefaults = new string[parameters];
             ParameterCount = parameters;
             _rtypes = new HashSet<string>();
@@ -181,7 +181,7 @@ namespace Microsoft.PythonTools.Analysis {
 
         private IAnalysisSet ChooseBest(IAnalysisSet x, IAnalysisSet y) {
             if (x == null || x.IsObjectOrUnknown()) {
-                return (y == null) ? AnalysisSet.Empty : y;
+                return (y == null || y.IsObjectOrUnknown()) ? AnalysisSet.Empty : y;
             }
             if (y == null || y.IsObjectOrUnknown()) {
                 return AnalysisSet.Empty;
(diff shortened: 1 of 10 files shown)


src/Analysis/Engine/Impl/OverloadResult.cs
```diff
@@ -117,44 +112,54 @@ public class OverloadResult : IOverloadResult {
...
             return new OverloadResult(parameters, name, doc, returnType);
         }
 
         public override string ToString() {
-            return "{0}({1})->[{2}]{3}".FormatInvariant(
-                Name,
-                string.Join(",", Parameters.Select(p => "{0}{1}:{2}={3}".FormatInvariant(p.Name, p.IsOptional ? "?" : "", p.Type ?? "", p.DefaultValue ?? ""))),
-                string.Join(",", ReturnType.OrderBy(k => k)),
-                string.IsNullOrEmpty(Documentation) ? "" : ("'''{0}'''".FormatInvariant(Documentation))
+            var parameters = string.Join(", ",
```
AlexanderSher: Where do we use this `ToString` implementation? In serialization?

MikhailArkhipov: It actually generates the signature

src/Analysis/Engine/Test/TypeAnnotationTests.cs
```diff
@@ -392,13 +391,13 @@ private class StringConverter : TypeAnnotationConverter<string> {
 new_vector = scale(2.0, [1.0, -4.2, 5.4])
 ",
                 new[] {
-                    "scale:scale(scalar:float=,vector:list[float]=)->[list,list[float]]"
+                    "scale:scale(scalar:float, vector:list[float]) -> list[float]"
```
AlexanderSher: So `[list,list[float]]` is actually merged into `list[float]`?

src/Analysis/Engine/Impl/Values/BuiltinInstanceInfo.cs
```diff
@@ -33,6 +33,7 @@ public BuiltinInstanceInfo(BuiltinClassInfo klass)
         IBuiltinClassInfo IBuiltinInstanceInfo.ClassInfo => ClassInfo;
         public BuiltinClassInfo ClassInfo => _klass;
 
+        public override string Name => _klass.Name;
```
AlexanderSher: Do we need similar change for `InstanceInfo`?

src/Analysis/Engine/Test/TypeAnnotationTests.cs
MikhailArkhipov: Yes. Tooltips were correct while signatures were really weird

src/Analysis/Engine/Impl/OverloadResult.cs
```diff
@@ -172,17 +177,7 @@ class AccumulatedOverloadResult {
             return x.Length >= y.Length ? x : y;
         }
 
-        private IAnalysisSet ChooseBest(IAnalysisSet x, IAnalysisSet y) {
```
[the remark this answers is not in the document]
MikhailArkhipov: Yes

src/Analysis/Engine/Impl/Values/BuiltinInstanceInfo.cs
MikhailArkhipov: Won't hurt, I'll add them.

PR #162, opened 2018-09-27
Properly search for the colon when checking for type hints in function arguments

### LINKED ISSUE #158: Fix multi-token type hints in line formatter
The line formatter handles the spacing of `=` inside of a function declaration with a type hint by looking two tokens behind the `=` to see if there's a colon. This works for simple declarations like `def foo(a: str = "bar")`, but does not work for more complicated types such as `def foo(a: Union[int, str] = "bar")`.

A more correct method would be to look for a colon anywhere behind the `=` until reaching a `(` or `,`.

jakebailey:
Fixes #158.

### DIFF
--- a/src/LanguageServer/Impl/Implementation/LineFormatter.cs
+++ b/src/LanguageServer/Impl/Implementation/LineFormatter.cs
@@ -146,7 +146,31 @@ namespace Microsoft.Python.LanguageServer.Implementation {
                         break;
 
                     case TokenKind.Assign:
-                        if (token.IsInsideFunctionArgs && prev?.PrevNonIgnored?.Kind != TokenKind.Colon) {
+                        if (token.IsInsideFunctionArgs) {
+                            // Search backwards through the tokens looking for a colon for this argument,
+                            // indicating that there's a type hint and spacing should surround the equals.
+                            for (var p = token.PrevNonIgnored; p != null; p = p.PrevNonIgnored) {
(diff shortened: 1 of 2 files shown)


src/LanguageServer/Impl/Implementation/LineFormatter.cs
```diff
@@ -146,7 +146,31 @@ public class LineFormatter {
...
+                                }
+
+                                if (p.Kind == TokenKind.Comma) {
+                                    // Found a comma, indicating the end of another argument, so stop.
+                                    break;
+                                }
+
+                                if (p.Kind == TokenKind.Colon) {
+                                    // Found a colon before hitting another argument or the opening parenthesis, so add spacing.
+                                    goto case TokenKind.AddEqual;
```
AlexanderSher: Please don't use `goto`.
I'd actually extract the whole case body into separate method and use `case … when …`

AlexanderSher: > against using goto 

I'm actually against fall-throughs from one non-empty case to another. Leave it as it is for now, I'll take a look myself.

MikhailArkhipov: It can bet something like HandleABC() function called from two cases.

AlexanderSher:
Closed in favor of #168.

PR #110, opened 2018-09-19
Fix wait for analysis complete, fix tests and avoid reload on all config changes

MikhailArkhipov:
Fixes #111 

- Improves #109 - avoid reloading modules on any configuration change. Still need to debug why reloading modules leaks.

- Improves test stability as well as Rename/FindReferences/DocumentSymbols/GotoDefinition by waiting for the complete analysis.

PR #169, opened 2018-09-28
RELEASE: Remove requeue of dependent modules

AlexanderSher:
This is a hot fix to avoid infinite queuing of modules for analysis.


MikhailArkhipov:
OK, this will undo some of the GotoDef/FindReferences stuff in https://github.com/Microsoft/python-language-server/pull/110, but we'll try and find another approach later.

PR #171, opened 2018-09-30
RELEASE: Restore path watching fix lost in merges

MikhailArkhipov:
- Apparently new code was never called. Old code remained despite new function exists.
- Fix issue in throttling

### DIFF
--- a/src/LanguageServer/Impl/PathsWatcher.cs
+++ b/src/LanguageServer/Impl/PathsWatcher.cs
@@ -62,6 +62,7 @@ namespace Microsoft.Python.LanguageServer {
                     fsw.Deleted += OnChanged;
 
                     _disposableBag
+                        .Add(() => _throttleTimer?.Dispose())
                         .Add(() => fsw.Created -= OnChanged)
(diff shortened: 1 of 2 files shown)


src/LanguageServer/Impl/PathsWatcher.cs
```diff
@@ -83,7 +84,7 @@ internal sealed class PathsWatcher : IDisposable {
 
         private void TimerProc(object o) {
             lock (_lock) {
-                if (_changedSinceLastTick) {
+                if (!_changedSinceLastTick) {
```
AlexanderSher: I believe the old comparison was correct.
Otherwise it's not clear why later `_changedSinceLastTick` is set to `false`.

MikhailArkhipov: It is throttling. Flag tells when there were no changes since last timer tick - i.e. multiple file change activity (such as installation of a large package) is over and it is time to reload modules. In fact, I am thinking 500 ms is too low.

PR #176, opened 2018-10-01
Fix bunch of issues with reference related tests

MikhailArkhipov:
- Properly resolve function arguments vs global variables
- Treat reassignments as reference, not as definition
- Improve CancelAnalysis test (took forever with my 6500 files library)
- Fix some dispose issues (queue was not always empty after dispose/cancel)
- Dispose of stuff in order
- Update signature test to new behavior
- Update reference tests to new behavior


src/Analysis/Engine/Impl/ModuleAnalysis.cs
```diff
@@ -114,7 +114,9 @@ public sealed class ModuleAnalysis : IModuleAnalysis {
         }
 
         internal IEnumerable<AnalysisVariable> ToVariables(IReferenceable referenceable) {
-            if (referenceable is VariableDef def) {
+            var def = referenceable as VariableDef;
```
AlexanderSher: Why pattern matching didn't work here?

src/Analysis/Engine/Impl/ModuleAnalysis.cs
```diff
@@ -251,6 +270,18 @@ private class DefinitionWalker : PythonWalkerWithLocation {
             return new VariablesResult(variables, unit.Tree);
         }
 
+        private bool IsFunctionArgument(IScope scope, IAnalysisVariable v) {
+            if (v.Variable?.Types?.MaybeEnumerate().FirstOrDefault() is ParameterInfo) {
```
AlexanderSher: Can `v.Variable` actually be `null`?

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -95,18 +95,18 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
                 .Add(() => {
                     foreach (var ext in _extensions.Values) {
                         (ext as IDisposable)?.Dispose();
                     }
                     foreach (var ext in _oldExtensions.Values) {
                         (ext as IDisposable)?.Dispose();
                     }
-                });
+                })
+                .Add(ProjectFiles)
```
AlexanderSher: What does this reordering affects?

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -625,6 +625,7 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
                         LogMessage(MessageType.Error, t.Exception.Message);
                         return;
                     }
+                    _shutdownCts.Token.ThrowIfCancellationRequested();
```
AlexanderSher: It's enough to call it either here or in `OnDocumentChangeProcessingCompleteAsync`.

src/LanguageServer/Impl/Implementation/Server.cs
MikhailArkhipov: Mostly correct order - send cancellation, clear queue, then dispose analyzer, then the rest.

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -645,6 +647,7 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
                 }
 
                 if (doc is IAnalyzable analyzable && enqueueForAnalysis) {
+                    _shutdownCts.Token.ThrowIfCancellationRequested();
```
AlexanderSher: We can check this unconditionally.

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -635,6 +636,7 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
 
         private async Task OnDocumentChangeProcessingCompleteAsync(IDocument doc, VersionCookie vc, bool enqueueForAnalysis, AnalysisPriority priority, IDisposable disposeWhenEnqueued) {
             try {
+                _shutdownCts.Token.ThrowIfCancellationRequested();
```
AlexanderSher: Then we don't need to call `_disposableBag.ThrowIfDisposed();` here (and catch it at the end).

MikhailArkhipov: Actually OnDocumentChangeProcessingCompleteAsync is not even async anymore

src/Analysis/Engine/Impl/ModuleAnalysis.cs
```diff
@@ -234,15 +236,32 @@ private class DefinitionWalker : PythonWalkerWithLocation {
                 if (!scope.EnumerateTowardsGlobal.Any()) {
                     variables = _unit.State.BuiltinModule.GetDefinitions(name.Name).SelectMany(ToVariables);
                 } else {
-                    foreach (var defScope in scope.EnumerateTowardsGlobal
-                        .Where(s => s.ContainsVariable(name.Name) && (s == scope || s.VisibleToChildren || IsFirstLineOfFunction(scope, s, location)))) {
-                        var scopeVariables = GetVariablesInScope(name, defScope).Distinct();
+                    foreach (var s in scope.EnumerateTowardsGlobal) {
```
AlexanderSher: Can we extract the body of `if (expr is NameExpression name) {` into separate method (something like `GetVariablesFromNameExpression`)?

src/Analysis/Engine/Impl/ModuleAnalysis.cs
MikhailArkhipov: I am not 100% sure. Original code did `if (referenceable is VariableDef def) { `. I think is is *always* a variable but I'd rather be safe.

src/Analysis/Engine/Impl/ModuleAnalysis.cs
MikhailArkhipov: Below code didnt use def before and now it does

AlexanderSher: Ok

src/Analysis/Engine/Impl/ModuleAnalysis.cs
MikhailArkhipov: Sure

AlexanderSher:
Do you want the last change to be part of the same review?

MikhailArkhipov:
Yes. It actually fixes `FindReferences` and makes it work as PyCharm does. Also adds test for class local variables.

AlexanderSher:
Ok, I'll take a look tomorrow morning.

src/Analysis/Engine/Test/AnalysisTest.cs
```diff
@@ -3132,23 +3132,45 @@ not abc
         print(a)
         assert isinstance(a, int)
         a = 200
+        print(a)
 ";
                 await server.SendDidOpenTextDocument(uri, text);
 
-                var references = await server.SendFindReferences(uri, 3, 15);
-                references.Should().OnlyHaveReferences(
+                var expected = new (Uri documentUri, (int, int, int, int), ReferenceKind?) [] {
```
AlexanderSher: You don't need `documentUri` here: `var expected = new (Uri, (int, int, int, int), ReferenceKind?) [] {`

AlexanderSher: Or maybe we should just add an overload for `OnlyHaveReferences` with `(Uri documentUri, (int, int, int, int), ReferenceKind)` tuple.

src/Analysis/Engine/Test/AnalysisTest.cs
```diff
@@ -4068,7 +4091,7 @@ class x(object):
                 await server.OpenDefaultDocumentAndGetAnalysisAsync(code);
                 var signatures = await server.SendSignatureHelp(TestData.GetDefaultModuleUri(), signatureLine, 1);
 
-                signatures.Should().OnlyHaveSignature($"{expectedName}(a: complex, b: int, c: str, *d: {expectedDType})");
+                signatures.Should().OnlyHaveSignature($"{expectedName}(a, b, c, *d)");
```
AlexanderSher: So we aren't going to show types in signatures, right?

src/Analysis/Engine/Test/AnalysisTest.cs
```diff
@@ -5341,7 +5365,7 @@ class test():
 
             var files = Directory.GetFiles(Path.Combine(configuration.PrefixPath, "Lib"), "*.py", SearchOption.AllDirectories);
             Trace.TraceInformation($"Files count: {files}");
-            var contentTasks = files.Select(f => File.ReadAllTextAsync(f)).ToArray();
+            var contentTasks = files.Take(Math.Min(50, files.Length)).Select(f => File.ReadAllTextAsync(f)).ToArray();
```
AlexanderSher: `files.Take(50)` is enough.
From sources:
```cs
    private static IEnumerable<TSource> TakeIterator<TSource>(IEnumerable<TSource> source, int count)
    {
      if (count > 0)
      {
        foreach (TSource source1 in source)
        {
          yield return source1;
          if (--count == 0)
            break;
        }
      }
    }
```

src/Analysis/Engine/Test/AnalysisTest.cs
```diff
@@ -5350,15 +5374,15 @@ class test():
...
+                for (var i = 0; i < contentTasks.Length && server.AnalysisQueue.Count < 50; i++) {
                     await server.SendDidOpenTextDocument(new Uri(files[i]), contentTasks[i].Result);
                 }
 
                 server.AnalysisQueue.Count.Should().NotBe(0);
             } finally {
                 if (server != null) {
                     analysisCompleteTask = EventTaskSources.AnalysisQueue.AnalysisComplete.Create(server.AnalysisQueue, new CancellationTokenSource(10000).Token);
-                    serverDisposeTask = Task.WhenAny(Task.Run(() => server.Dispose()), Task.Delay(1000));
+                    serverDisposeTask = Task.WhenAny(Task.Run(() => server.Dispose()), Task.Delay(5000));
```
AlexanderSher: Do we really need 5 seconds to dispose?

src/Analysis/Engine/Test/AnalysisTest.cs
```diff
@@ -5864,1796 +5889,1796 @@ import mod1
...
-def dec1(C):
-    def sub_method(self): pass
-    C.sub_method = sub_method
-    return C
-
-@dec1
-class MyBaseClass1(object):
-    def base_method(self): pass
+        /*
+                [TestMethod, Priority(0)]
```
AlexanderSher: The whole block of commented tests is offset, please undo it.

src/Analysis/Engine/Test/AnalysisTest.cs
MikhailArkhipov: We show it if it is declared in the function definition (i.e. `def foo()`). We do not show 'derived' signatures that depend on actual passed parameters. Same as PyCharm.

src/Analysis/Engine/Test/AnalysisTest.cs
MikhailArkhipov: Not really.

src/Analysis/Engine/Test/AnalysisTest.cs
MikhailArkhipov: OK. Prob formatter.

PR #183, opened 2018-10-03
Remove deque & limit analysis passes

MikhailArkhipov:
If all variables in the module already have types, there is no need to re-analyze it.


src/Analysis/Engine/Impl/Analyzer/DDG.cs
```diff
@@ -46,7 +46,7 @@ internal class DDG : PythonWalker {
                 }
 
                 while (queue.Count > 0 && !cancel.IsCancellationRequested) {
-                    _unit = queue.PopLeft();
+                    _unit = queue.Dequeue();
```
AlexanderSher: We need to call `AssertScope` here. Or maybe before the loop.

PR #189, opened 2018-10-03
Fix References_GrammarTest_Statements

### LINKED ISSUE #40: References_GrammarTest_Statements test fails
Contains unexpected reference `abc.py, (0, 0) - (0, 0), Definition`

MikhailArkhipov:
Fixes #40 

Code is pretty confusing, with reassignments and import names same as variable names. Various apps behave differently. Basically the fix is

- Keep any reference to external files as definition (so use can navigate there)
- Keep `import abc` as reference (so use can find all usages of 'abc').

### DIFF
--- a/src/Analysis/Engine/Impl/ModuleAnalysis.cs
+++ b/src/Analysis/Engine/Impl/ModuleAnalysis.cs
@@ -294,15 +294,23 @@ namespace Microsoft.PythonTools.Analysis {
             mainDefinition = mainDefinition ?? definitions.FirstOrDefault();
             if (mainDefinition != null) {
                 // Drop definitions in outer scopes and convert those in inner scopes to references.
+                // Exception is when variable is an import as in 
+                //   abc = 1
+                //   import abc
+                 var imports = definitions
+                    .Where(d => d.Variable.Variable.Types.Any(t => t.TypeId == BuiltinTypeId.Module) && d.Location.DocumentUri != DocumentUri)
+                    .ToArray();
+
                 // Scope levels are numbered in reverse (X == main definition level, x+1 == one up).
                 var defsToRefs = definitions
+                    .Except(imports)
                     .Where(d => d != mainDefinition && d.ScopeLevel <= mainDefinition.ScopeLevel)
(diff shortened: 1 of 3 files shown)


src/Analysis/Engine/Test/AnalysisTest.cs
```diff
@@ -2992,12 +2991,19 @@ print abc
                 await server.SendDidOpenTextDocument(uri, text);
 
                 var references = await server.SendFindReferences(uri, 3, 12);
+
+                // External module 'abc', URI varies depending on install
+                var externalUri = references[1].uri;
+                externalUri.LocalPath.Should().Contain("abc.py");
```
AlexanderSher: `Contain` or `EndsWith`?

src/Analysis/Engine/Impl/ModuleAnalysis.cs
```diff
@@ -294,15 +294,23 @@ private class VariableScopePair : Tuple<IAnalysisVariable, int> {
...
                 var defsToRefs = definitions
+                    .Except(imports)
                     .Where(d => d != mainDefinition && d.ScopeLevel <= mainDefinition.ScopeLevel)
                     .Select(v => new VariableScopePair(new AnalysisVariable(v.Variable.Variable, VariableType.Reference, v.Location), v.ScopeLevel));
 
                 var others = variables
                             .Where(v => (v.VariableType == VariableType.Reference || v.VariableType == VariableType.Value) &&
                                          v.ScopeLevel <= mainDefinition.ScopeLevel);
-                variables = new[] { mainDefinition }.Concat(others.Concat(defsToRefs));
+                variables = new[] { mainDefinition }.Concat(imports).Concat(others.Concat(defsToRefs));
```
AlexanderSher: Is it just rearrangement, or some of the variables will be excluded?

src/Analysis/Engine/Test/AnalysisTest.cs
MikhailArkhipov: EndsWith

src/Analysis/Engine/Impl/ModuleAnalysis.cs
MikhailArkhipov: No exclusion, just keeping imports as definitions. Before they were converted to references since they were below in the same scope such as
```python
abc = 1
import abc
```

PR #191, opened 2018-10-04
mark abstract functions as such

lostmsu:
Adds `IsAbstract` property to `IFunctionInfo` and `FunctionInfo`, that tells if the function is abstract or not.

Note: should we consider using tri-state `bool?` here? During analysis, until all decorators are analyzed, `IsAbstract` may return `false`, when the final analysis result will be `true` (it is the case right now, as in many cases imports of `abc` are processed after functions).

### DIFF
--- a/src/Analysis/Engine/Impl/Values/Definitions/IFunctionInfo.cs
+++ b/src/Analysis/Engine/Impl/Values/Definitions/IFunctionInfo.cs
@@ -25,4 +25,8 @@ namespace Microsoft.PythonTools.Analysis.Values {
         bool IsClosure { get; }
         IPythonProjectEntry ProjectEntry { get; }
     }
+
+    public interface IFunctionInfo2 : IFunctionInfo {
+        bool IsAbstract { get; }
+    }
 }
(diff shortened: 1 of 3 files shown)


src/Analysis/Engine/Impl/Values/Definitions/IFunctionInfo.cs
```diff
@@ -23,6 +23,7 @@ public interface IFunctionInfo: IAnalysisValue {
         bool IsClassMethod { get; }
         bool IsProperty { get; }
         bool IsClosure { get; }
+        bool IsAbstract { get; }
```
MikhailArkhipov: Since these are public interfaces you may have to create `IFunctionInfo2` deriving from `IFunctionInfo` and add property there. This way code that uses public LS extensibility API will continue working. Yes, there is at least one VS Code extension that uses the API.

lostmsu:
@MikhailArkhipov , makes sense, updated with `IFunctionInfo2`.

lostmsu:
@MikhailArkhipov , any thoughts on tri-state?

PR #198, opened 2018-10-04
Make tuples with different arguments different types

### LINKED ISSUE #173: Got message "Hash code changed when merging AnalysisValues" with debug build
I wonder if it is an indication of some internal error potentially corrupting analysis?

It happens when I try to analyze installed tensorflow package. (e.g. pip install tensorflow-gpu, used version 1.10, then analyze site-packages/tensorflow).

MikhailArkhipov:
Fixes #173

Tuple protocols are compared by name which makes tuples with different types the same, which causes trouble in analysis hash set when it merges types of function parameters in overloads.

### DIFF
--- a/src/Analysis/Engine/Impl/Values/Protocols.cs
+++ b/src/Analysis/Engine/Impl/Values/Protocols.cs
@@ -415,7 +415,7 @@ namespace Microsoft.PythonTools.Analysis.Values {
             return AnalysisSet.UnionAll(constants.Select(GetItem));
         }
 
-        public override string Name => "tuple";
+        public override string Name => "tuple[{0}]".FormatInvariant(string.Join(", ", _values.Select(v => v.GetShortDescriptions())));
 
         public override IEnumerable<KeyValuePair<string, string>> GetRichDescription() {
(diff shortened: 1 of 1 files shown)


src/Analysis/Engine/Impl/Values/Protocols.cs
```diff
@@ -415,7 +415,7 @@ class TupleProtocol : IterableProtocol {
             return AnalysisSet.UnionAll(constants.Select(GetItem));
         }
 
-        public override string Name => "tuple";
+        public override string Name => "tuple[{0}]".FormatInvariant(string.Join(", ", _values.Select(v => v.GetShortDescriptions())));
```
AlexanderSher: We can save it `ctor` to avoid multiple string creation.

lostmsu:
Is short description unique enough? Should not `FullyQualifiedName` be used?

MikhailArkhipov:
@lostmsu - sure, why not

lostmsu:
@MikhailArkhipov FYI, just pulled this PR locally. Whatever issue I had with the assert is gone.

PR #200, opened 2018-10-04
Propagate known parameter types to derived functions

### LINKED ISSUE #194: Instance function parameter types should propagate to descendants in type hierarchy
The rationale is similar to the rationale for return types in #11 and #115.

If `A: B` and on an variable `b` of type `B` we call `b.foo(42)`, that means, that `A.foo` must also be able to accept `int`, because `b` can actually hold a value of type `A`.

Also related to: #131, #185

lostmsu:
Adds a set of derived functions to `FunctionInfo`. Fills the set when `DDG` encounters a `FunctionDefinition`.

Whenever a call to some function is analyzed, it looks at the list of derived functions, and passes the call info along to them.

Also added some styles to .editorconfig

Fixes #194 .
Seems to reduce the total number of failing test cases (local run shows 32->29).

I could not understand if I needed to also requeue anything for analysis. Left a comment about it in the code.

The change might be incomplete, as the new code was not added to the case when `_callsWithClosure != null` in `FunctionInfo.Call`. I could not quickly understand that code.

### DIFF
--- a/src/.editorconfig
+++ b/src/.editorconfig
@@ -76,3 +76,7 @@ csharp_new_line_before_finally = false
 csharp_new_line_before_members_in_anonymous_types = false
 
 csharp_new_line_before_members_in_object_initializers = false
+
+dotnet_style_qualification_for_field = false:suggestion
+dotnet_style_qualification_for_property = false:none
+dotnet_style_qualification_for_method = false:none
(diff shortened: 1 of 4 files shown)


src/.editorconfig
```diff
@@ -76,3 +76,7 @@ csharp_new_line_before_finally = false
 csharp_new_line_before_members_in_anonymous_types = false
 
 csharp_new_line_before_members_in_object_initializers = false
+
+dotnet_style_qualification_for_field = false:suggestion
+dotnet_style_qualification_for_property = false:none
```
MikhailArkhipov: I'd rather keep these as they were. We are slowly converting code to C# 7+ so we do want to see suggestion/information for `=>` style so we can fix them as we go.

MikhailArkhipov:
I'll merge and fix the editorconfig

PR #206, opened 2018-10-05
Make sure tuple protocol name does not ignore unknown parameter types

MikhailArkhipov:
Additional fix for https://github.com/Microsoft/python-language-server/issues/173 since 'SelectMany' drops empty sequences and tuples lose arguments of unknown type making them look like they have fewer arguments.

### DIFF
--- a/src/Analysis/Engine/Impl/Values/Protocols.cs
+++ b/src/Analysis/Engine/Impl/Values/Protocols.cs
@@ -410,17 +410,12 @@ namespace Microsoft.PythonTools.Analysis.Values {
                 return;
             }
 
-            if (sets.Length == 1) {
-                sb.Append(sets[0] is IHasQualifiedName qn ? qn.FullyQualifiedName : sets[0].ShortDescription);
-                return;
-            }
-
-            sb.Append('[');
+            sb.AppendIf(sets.Length > 1, "[");
             for (var i = 0; i < sets.Length; i++) {
                 sb.AppendIf(i > 0, ", ");
                 sb.Append(sets[i] is IHasQualifiedName qn ? qn.FullyQualifiedName : sets[i].ShortDescription);
             }
-            sb.Append(']');
+            sb.AppendIf(sets.Length > 1, "]");
         }
 
         protected override void EnsureMembers(IDictionary<string, IAnalysisSet> members) {
--- a/src/Analysis/Engine/Impl/Values/Protocols.cs
+++ b/src/Analysis/Engine/Impl/Values/Protocols.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using Microsoft.PythonTools.Analysis.Infrastructure;
 using Microsoft.PythonTools.Interpreter;
 using Microsoft.PythonTools.Parsing;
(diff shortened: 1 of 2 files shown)


src/Analysis/Engine/Impl/Values/Protocols.cs
```diff
@@ -389,9 +390,41 @@ class TupleProtocol : IterableProtocol {
...
+                if (i > 0) {
+                    sb.Append(", ");
+                }
+                sb.Append(GetParameterString(_values[i].ToArray()));
+            }
+            sb.Append(']');
+            return sb.ToString();
+        }
+
+        private string GetParameterString(AnalysisValue[] sets) {
```
AlexanderSher: you can reuse the same `StringBuilder` here.

src/Analysis/Engine/Impl/Values/Protocols.cs
```diff
@@ -389,9 +390,41 @@ class TupleProtocol : IterableProtocol {
...
+            if (sets.Length == 0) {
+                return "?";
+            }
+            var sb = new StringBuilder();
+            if (sets.Length > 1) {
```
AlexanderSher: This can be simplified:
```cs
If (sets.Length == 1) {
    sb.Append(sets[0] is IHasQualifiedName qn ? qn.FullyQualifiedName : sets[i].ShortDescription)
} else {
    // remaining code
}
```

src/Analysis/Engine/Impl/Values/Protocols.cs
```diff
@@ -389,9 +390,41 @@ class TupleProtocol : IterableProtocol {
...
+                sb.Append('[');
+            }
+            for (var i = 0; i < sets.Length; i++) {
+                if (i > 0) {
+                    sb.Append(", ");
```
AlexanderSher: `", "` or `","`?

src/Analysis/Engine/Impl/Values/Protocols.cs
```diff
@@ -389,9 +390,41 @@ class TupleProtocol : IterableProtocol {
...
+            Name = GetNameFromValues();
+        }
 
-            var argTypes = _values.SelectMany(e => e.Select(v => v is IHasQualifiedName qn ? qn.FullyQualifiedName : v.ShortDescription));
-            Name = "tuple[{0}]".FormatInvariant(string.Join(", ", argTypes));
+        private string GetNameFromValues() {
+            // Enumerate manually since SelectMany drops empty/unknown values
+            var sb = new StringBuilder("tuple[");
+            for (var i = 0; i < _values.Length; i++) {
+                if (i > 0) {
```
AlexanderSher: We can have an extension method for this:
https://github.com/Microsoft/RTVS/blob/master/src/Common/Core/Impl/Extensions/StringBuilderExtensions.cs

src/Analysis/Engine/Impl/Values/Protocols.cs
MikhailArkhipov: Good point

src/Analysis/Engine/Impl/Values/Protocols.cs
MikhailArkhipov: ", " more readable

Issue #165, opened 2018-09-27
Implement on-type block formatting


jakebailey:
The VS Code extension provides an on-type formatter that triggers on `:` which allows for block formatting. For example, you can hit colon after the `else` in the following code:

```python
if x:
    pass
    else
```

Which will then be reformatted into:

```python
if x:
    pass
else:
```

This functionality should be implemented in language server for feature parity. The language server may even be able to handle more cases correctly than the extension's use of regex, since we have access to an AST.

Since on-type formatting has only one entrypoint (`DocumentOnTypeFormatting`), that function would need to map trigger characters to the right handler. A nicety would be to have a Dictionary of handlers, then produce the `DocumentOnTypeFormattingOptions`'s trigger character members from that dictionary rather than needing to remember to change the characters in two different files.

Related to Microsoft/vscode-python#2714.

MikhailArkhipov:
Also this https://github.com/Microsoft/vscode-python/issues/481

PR #217, opened 2018-10-09
Port regex-based block formatting from vscode-python

jakebailey:
Updates #165.

This is a 1:1 port of the `:` triggered block formatter that's running in the extension, including its tests. See here for the current implementation:
- https://github.com/Microsoft/vscode-python/blob/master/src/client/typeFormatters/codeBlockFormatProvider.ts
- https://github.com/Microsoft/vscode-python/blob/master/src/client/typeFormatters/blockFormatProvider.ts
- https://github.com/Microsoft/vscode-python/blob/master/src/client/typeFormatters/contracts.ts

This may not be the most efficient or most correct way to do this sort of formatting, but not having any block formatting in PLS means that we can't also ship the line formatting code. This will at least behave the same way and let us ship the new line formatter without regressing the editor in terms of features. This can be replaced later with something that uses the tokenizer/parser to act much more generally.

### DIFF
--- a/src/Analysis/Engine/Test/BlockFormatterTests.cs
+++ b/src/Analysis/Engine/Test/BlockFormatterTests.cs
@@ -1,0 +1,305 @@
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABLITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using FluentAssertions;
+using Microsoft.Python.LanguageServer;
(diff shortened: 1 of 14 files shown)


src/Analysis/Engine/Test/BlockFormatterTests.cs
```diff
@@ -0,0 +1,305 @@
...
+
+        [TestMethod, Priority(0)]
+        public void NullReader() {
+            Func<Task<TextEdit[]>> func = () => BlockFormatter.ProvideEdits(null, new Position(), new FormattingOptions());
+            func.Should().Throw<ArgumentNullException>().And.ParamName.Should().Be("reader");
+        }
+
+        [TestMethod, Priority(0)]
+        public async Task FirstLine() {
+            using (var reader = new StringReader("")) {
```
AlexanderSher: You don't need using for `StringReader` when it is used explicitly. It doesn't hold any resources.

src/Analysis/Engine/Test/LanguageServerTests.cs
```diff
@@ -1006,6 +1006,13 @@ class C:
                 edits = await s.SendDocumentOnTypeFormatting(uri, new SourceLocation(4, 1), "\n");
                 edits.Should().OnlyHaveTextEdit("x += 1", (2, 4, 2, 9));
             }
+
+            using (var s = await CreateServer()) {
```
AlexanderSher: Can we split it in two tests?

src/Analysis/Engine/Test/BlockFormatterTests.cs
jakebailey: I can certainly remove it, but I'm a little bit wary of not disposing of it just because we know it doesn't do anything given that a future .NET release could easily implement it differently.

src/Analysis/Engine/Test/LanguageServerTests.cs
jakebailey: Done.

src/Analysis/Engine/Test/BlockFormatterTests.cs
jakebailey: I'll fix that `string.Empty` thing though. Muscle memory from other languages which don't allocate an empty string...

PR #227, opened 2018-10-10
Fix MutatingReferences and couple more things

MikhailArkhipov:
- Queue dependent modules, but don't walk imports endlessly
- Add Start/Stop methods to the analysis queue so we can delay analysis until all files are loaded


src/Analysis/Engine/Impl/Infrastructure/Extensions/HashSetExtensions.cs
```diff
@@ -0,0 +1,27 @@
...
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System.Collections.Generic;
+
+namespace Microsoft.PythonTools.Analysis.Infrastructure.Extensions {
+    public static class HashSetExtensions {
+        public static void Add<T>(this HashSet<T> set, IEnumerable<T> items) {
+            foreach(var e in items) {
```
jakebailey: Super nitpick, but I think the formatting rules should have added a space after `foreach`, so hopefully VS hasn't undone your settings (it does it to me all the time...).

MikhailArkhipov:
Will reopen after merge

PR #240, opened 2018-10-11
Fix function override completion test, improve behavior

### LINKED ISSUE #236: CompletionForOverrideArgs incorrect behavior between Python 2/3
The test case introduced in #161 is incorrect. `arg` should have been `self`, as it's a class method that's being overridden. The test should be modified to be:

```python
class A(object):
    def bar(self, arg=None):
        pass

class B(A):
    def bar(self, arg=None):
        return super().bar()
```

However, that's only valid for Python 3, where arguments to `super` are optional. For Python 2, the `Bar` class needs to be completed out to:

```python
class B(object):
    def bar(self, arg=None):
        return super(A, self).bar()
```

So the test needs to have two cases via checking `this is LanguageServerTests_V2`.

See:
- https://docs.python.org/2/library/functions.html#super
- https://docs.python.org/3/library/functions.html#super

Additionally, it may be more correct to not suggest this sort of thing for non-"new-style" classes in Python 2, but I'm not sure how feasible that is or if it's worth the effort to detect.

jakebailey:
Fixes #236.

This also improves the completed override function by not repeating default values, instead referring to parameters directly.

Old behavior:

old

New behavior:

new

This also removes the "safe" check for variable names, as I think all it seems to actually do is prevent non-ASCII identifiers from working (which are [allowed in Python 3](https://docs.python.org/3/reference/lexical_analysis.html#identifiers)).

### DIFF
--- a/src/Analysis/Engine/Test/LanguageServerTests.cs
+++ b/src/Analysis/Engine/Test/LanguageServerTests.cs
@@ -442,7 +442,7 @@ class B(A):
     def f";
 
             using (var s = await CreateServer()) {
-                var u = await AddModule(s, code);
+                var u = await s.OpenDefaultDocumentAndGetUriAsync(code);
 
                 await AssertNoCompletion(s, u, new SourceLocation(3, 9));
 
--- a/src/Analysis/Engine/Test/LanguageServerTests.cs
+++ b/src/Analysis/Engine/Test/LanguageServerTests.cs
@@ -433,14 +433,30 @@ namespace AnalysisTests {
 
         [TestMethod, Priority(0)]
         public async Task CompletionForOverrideArgs() {
-            using (var s = await CreateServer()) {
-                var u = await AddModule(s, "class A:\n    def bar(arg=None): pass\n\nclass B(A):\n    def b");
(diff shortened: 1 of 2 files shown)


src/Analysis/Engine/Test/LanguageServerTests.cs
```diff
@@ -433,14 +433,30 @@ public class LanguageServerTests {
...
+class A(object):
+    def foo(self, a, b=None, *args, **kwargs):
+        pass
+
+class B(A):
+    def f";
+
             using (var s = await CreateServer()) {
-                var u = await AddModule(s, "class A:\n    def bar(arg=None): pass\n\nclass B(A):\n    def b");
+                var u = await AddModule(s, code);
```
MikhailArkhipov: To do what actually product does, you can do

```cs
    var u = await server.OpenDefaultDocumentAndGetUriAsync(code);
```

jakebailey: Done, though if that's the new "correct" way to do it, then it should probably also be replaced everywhere else.

PR #243, opened 2018-10-11
Fix Goto Definition with class methods

### LINKED ISSUE #208: Intellisense is wrong with staticmethod
https://github.com/Microsoft/vscode-python/issues/182

image

Intellisense in `foo` method is wrong, `formmater`. need show `str`'s method, like `format()`, but it will show `Test`'s method, like `foo()`

### LINKED ISSUE #225: Goto Definition does not work with class methods when modules reference each other
Works in a single file, but not cross-files, specifically with circular references. Works if method is not a class method.

`module1`:
```python
from .module2 import Derived1

class Base(object):
    @classmethod
    def fob_base(cls): 
        pass

class Derived2(Derived1):
    pass

Derived2.fob_base()
```
`module2`:
```python
from module1 import *

class Derived1(Base):
    pass
```

MikhailArkhipov:
Fixes #225
Fixes #208 

- `GetDefinition(name, types)` on bases does not need to push type since it calls main `GetDefinition(name)` which will push it. Pushing in the inner function prevents the outer from finding bases.
- Add related tests for cases around class method navigation and various types of imports.

### DIFF
--- a/src/Analysis/Engine/Impl/Values/ClassInfo.cs
+++ b/src/Analysis/Engine/Impl/Values/ClassInfo.cs
@@ -636,19 +636,8 @@ namespace Microsoft.PythonTools.Analysis.Values {
             return result;
         }
 
-        private IEnumerable<IReferenceable> GetDefinitions(string name, IEnumerable<AnalysisValue> nses) {
-            var result = new List<IReferenceable>();
-            foreach (var subType in nses) {
-                if (subType.Push()) {
-                    IReferenceableContainer container = subType as IReferenceableContainer;
-                    if (container != null) {
-                        result.AddRange(container.GetDefinitions(name));
-                    }
-                    subType.Pop();
-                }
-            }
-            return result;
-        }
+        private IEnumerable<IReferenceable> GetDefinitions(string name, IEnumerable<AnalysisValue> nses)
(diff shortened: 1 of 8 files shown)


src/Analysis/Engine/Impl/Values/ClassInfo.cs
```diff
@@ -636,19 +636,8 @@ internal class ClassInfo : AnalysisValue, IClassInfo, IHasRichDescription, IHasQ
...
-                    if (container != null) {
-                        result.AddRange(container.GetDefinitions(name));
-                    }
-                    subType.Pop();
-                }
-            }
-            return result;
-        }
+        private IEnumerable<IReferenceable> GetDefinitions(string name, IEnumerable<AnalysisValue> nses)
+            => nses.OfType<IReferenceableContainer>().SelectMany(c => c.GetDefinitions(name));
```
AlexanderSher: What those `subType.Push()` and `subType.Pop()` has been doing?

MikhailArkhipov: There is static hashset that tracks that the type is already being processed. I guess to prevent stack overflows with circular references. In this case, however, `GetDefinition` will add the type, otherwise, if added here, `GetDefinition` will return nothing since it will think type is processed up the call stack.

PR #249, opened 2018-10-12
Fix #241: addBrackets adds parens to function override completions

### LINKED ISSUE #241: addBrackets adds parens to function override completions
Example:

bad_bracket

Note the extra parens added after the completed function.

Since this involves override completion, it may be better to wait to fix this until #240 is merged (to prevent conflicts).

AlexanderSher:
This PR also contains big refactoring of `CompletionAnalysis`:
- Non-intersecting code paths are separated
- Control flow is separated from actual completion items
- Nesting reduced

### DIFF
--- a/src/Analysis/Engine/Impl/Infrastructure/Check.cs
+++ b/src/Analysis/Engine/Impl/Infrastructure/Check.cs
@@ -3,20 +3,37 @@
 
 using System;
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 using static System.FormattableString;
 
 namespace Microsoft.PythonTools.Analysis.Infrastructure {
     public static class Check {
         [DebuggerStepThrough]
+        public static void FieldType<T>(string fieldName, object fieldValue) {
+            if (!(fieldValue is T)) {
+                throw new InvalidOperationException($"Field {fieldName} must be of type {fieldValue}");
+            }
+        }
+
+        [DebuggerStepThrough]
+        public static void ArgumentOfType<T>(string argumentName, object argument, [CallerMemberName] string callerName = null) {
+            ArgumentNull(argumentName, argument);
+
+            if (!(argument is T)) {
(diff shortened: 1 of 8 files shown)


src/Analysis/Engine/Impl/Infrastructure/Check.cs
```diff
@@ -3,20 +3,37 @@
...
-            if (argument == null) {
+            if (argument is null) {
                 throw new ArgumentNullException(argumentName);
             }
         }
 
         [DebuggerStepThrough]
-        public static void ArgumentStringNullOrEmpty(string argumentName, string argument) {
-            Check.ArgumentNull(argumentName, argument);
+        public static void ArgumentNotNullOrEmpty(string argumentName, string argument) {
```
jakebailey: 👍 So happy to see this function name change to actually reflect what it does.

src/Analysis/Engine/Test/CompletionTest.cs
```diff
@@ -410,5 +410,23 @@ class Class1(list):
...
+
+class B(A):
+    def f";
+
+            await server.SendDidChangeConfiguration(new ServerSettings.PythonCompletionOptions { addBrackets = true });
+            var uri = await server.OpenDefaultDocumentAndGetUriAsync(code);
+            var completion = await server.SendCompletion(uri, 6, 9);
+
+            completion.Should().HaveItem("foo")
+                .Which.Should().HaveInsertText("foo(self):\r\n    return super().foo()");
```
jakebailey: Just to be clear, these tests run under Python 3, correct? Otherwise, it'd be completed out to `super(A, self).foo()`.

MikhailArkhipov: May be worth handling both?

AlexanderSher: Done

PR #251, opened 2018-10-13
Navigation and tooltip on self

### LINKED ISSUE #22: Go to definition on self should navigate to the class definition
https://github.com/Microsoft/vscode-python/issues/1637

MikhailArkhipov:
Fixes #22 

- Provide tooltip on `self` that is from the class definition
- Goto definition on `self` navigates to the class definition
- Move hover tests to a separate class
- Fix issues in various tests that did not actually load files into the server. Opening is not enough if more than a single file is involved.


src/Analysis/Engine/Test/FindReferencesTests.cs
```diff
@@ -878,6 +885,9 @@ class bcd(abc):
             var fobUri = TestData.GetTempPathUri("fob.py");
             var oarUri = TestData.GetTempPathUri("oar.py");
             using (var server = await CreateServerAsync(PythonVersions.LatestAvailable3X)) {
+                await server.LoadFileAsync(fobUri);
```
AlexanderSher: `Server.DidOpenTextDocument` calls the same `AddFileAsync` as `LoadFileAsync` (`ProjectFiles.GetEntry` is a simple dictionary lookup):
https://github.com/Microsoft/python-language-server/blob/39a17b1a8aefed9ee18d904ceff2d3c447f0f060/src/LanguageServer/Impl/Implementation/Server.cs#L192

So if these loads change anything in test results, it is not because of the codeflow, but because of the timings.

AlexanderSher: Or, maybe calling `EnqueueItem` twice really changes something, but that isn't expected either.

MikhailArkhipov: There is known problem with files being analyzed while other files are still being added. I am going to get my `AnalysisQueue.Start/Stop` change and see if that works better.

MikhailArkhipov: AnalysisQueue.Start/Stop works OK, no need to double load/open

src/Analysis/Engine/Impl/Intellisense/AnalysisQueue.cs
```diff
@@ -34,13 +34,16 @@ internal sealed class AnalysisQueue : IDisposable {
...
         private readonly Task _consumerTask;
+        private readonly ManualResetEventSlim _queueEnabled = new ManualResetEventSlim(true);
 
         public event EventHandler AnalysisStarted;
         public event EventHandler AnalysisComplete;
         public event EventHandler AnalysisAborted;
         public event EventHandler<UnhandledExceptionEventArgs> UnhandledException;
 
         public int Count => _ppc.Count;
+        public void Start() => _queueEnabled.Set();
```
AlexanderSher: It looks like we can replace pair of `Start` and `Stop` with `IDisposable Pause` using `CountdownDisposable` from RTVS: https://github.com/Microsoft/RTVS/blob/master/src/Common/Core/Impl/Disposables/CountdownDisposable.cs

src/Analysis/Engine/Test/AnalysisTest.cs
```diff
@@ -1280,8 +1286,7 @@ yield 1
 
 
         [DataRow(PythonLanguageVersion.V27)]
-        [DataRow(PythonLanguageVersion.V31)]
-        [DataRow(PythonLanguageVersion.V33)]
+        [DataRow(PythonLanguageVersion.V36)]
```
AlexanderSher: Do we stop supporting `PythonLanguageVersion.V31` and `PythonLanguageVersion.V33`?

src/Analysis/Engine/Test/AnalysisTest.cs
```diff
@@ -703,9 +703,12 @@ import mod1
...
                 var uri1 = TestData.GetTempPathUri("mod1.py");
                 var uri2 = TestData.GetTempPathUri("mod2.py");
-                await server.SendDidOpenTextDocument(uri1, text1);
-                await server.SendDidOpenTextDocument(uri2, text2);
+                using (FileLoading()) {
+                    await server.SendDidOpenTextDocument(uri1, text1);
+                    await server.SendDidOpenTextDocument(uri2, text2);
+                }
 
+                await server.WaitForCompleteAnalysisAsync(CancellationToken.None);
```
AlexanderSher: Ok, these changes make tests pass, but when we make changes to `LanguageServer` so that it doesn't load all the files in the root directory and subdirectories, these tests will become irrelevant to the product behavior, cause there will be no waiting.

src/Analysis/Engine/Impl/Intellisense/AnalysisQueue.cs
MikhailArkhipov: OK. I did it in tests, will move to main line.

src/Analysis/Engine/Test/AnalysisTest.cs
MikhailArkhipov: I am a bit on a fence here. I think test was added back when 3.3 was new. Now 3.7 is actually the main line. I think actually we should be using latest... Using 3.3 here requires its installation and if it is not installed, the test does not run which is bad.

src/Analysis/Engine/Test/AnalysisTest.cs
MikhailArkhipov: Even then the analysis will still have to be completed fully on the loaded set and importing mechanism will be loading files that are referenced, even if they are not open.

src/Analysis/Engine/Impl/Intellisense/AnalysisQueue.cs
```diff
@@ -55,6 +58,7 @@ private void Enqueue(object key, Func<CancellationToken, Task> handler, Analysis
             RaiseEventOnThreadPool(AnalysisStarted);
             while (!_ppc.IsDisposed) {
                 try {
+                    _queueEnabled.Wait();
```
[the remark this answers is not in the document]
MikhailArkhipov: Done. Not sure about the `CancellationToken` since `CountdownDisposable` is not using it. If `ppc` is canceled, it doesn't matter if queue is stopped.

src/Analysis/Engine/Impl/ModuleAnalysis.cs
```diff
@@ -327,6 +336,22 @@ private class VariableScopePair : Tuple<IAnalysisVariable, int> {
...
+            }
+            var cd = instanceInfo.ClassInfo.ClassDefinition;
+            var cdLocation = unit.Tree.IndexToLocation(cd.StartIndex);
+            var cdLocInfo = new LocationInfo(selfDefinition.Location.FilePath, selfDefinition.Location.DocumentUri, cdLocation.Line, cdLocation.Column);
+
+            var classDefinition = new AnalysisVariable(selfDefinition.Variable.Variable, VariableType.Definition, cdLocInfo);
+            var reference = new AnalysisVariable(selfDefinition.Variable.Variable, VariableType.Reference, selfDefinition.Location);
+            var otherRefs = definitions.Where(d => d.VariableType == VariableType.Reference).Select(d => d.Variable);
+
+            return new VariablesResult(new[] { classDefinition }.Concat(new[] { reference }).Concat(otherRefs), unit.Tree);
```
AlexanderSher: `new[] { classDefinition, reference }`

src/Analysis/Engine/Test/AnalysisTest.cs
AlexanderSher: We can have both 3.3 and 3.7

MikhailArkhipov: OK

PR #255, opened 2018-10-13
Make sure enumerables dont change

### LINKED ISSUE #253: CrossThreadAnalysisCalls occasionally fails
```
Message: Test method AnalysisTests.ThreadingTest.CrossThreadAnalysisCalls threw exception: 
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
```

MikhailArkhipov:
Fixes #253

### DIFF
--- a/src/Analysis/Engine/Impl/Values/ClassInfo.cs
+++ b/src/Analysis/Engine/Impl/Values/ClassInfo.cs
@@ -318,7 +318,7 @@ namespace Microsoft.PythonTools.Analysis.Values {
             );
         }
 
-        public IEnumerable<IAnalysisSet> Bases => _bases;
+        public IEnumerable<IAnalysisSet> Bases => _bases.AsLockedEnumerable();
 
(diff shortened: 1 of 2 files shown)


src/Analysis/Engine/Impl/Values/ClassInfo.cs
```diff
@@ -318,7 +318,7 @@ internal class ClassInfo : AnalysisValue, IClassInfo, IHasRichDescription, IHasQ
             );
         }
 
-        public IEnumerable<IAnalysisSet> Bases => _bases;
+        public IEnumerable<IAnalysisSet> Bases => _bases.AsLockedEnumerable();
```
AlexanderSher: `LockedEnumerableObject` only takes the lock for enumerator. Pretty much it does nothing. What you need here is probably change the type of the `_bases` to array (or `ImmutableArray`) and inside `SetBases` simply replace the field value. `SetBase` can be removed, it isn't used anywhere.

PR #271, opened 2018-10-17
Clean Python versions in tests

### LINKED ISSUE #268: Clean up tests with regarding Python versions
We should use latest installed (upcoming 3.8) as `latest` and continue testing on 2.7, 3.5 and 3.6 and 3.7. We may choose to deprecate handling of 3.1/3.2/3.3 in analysis tests

MikhailArkhipov:
Fixes #268

- Add 3.7 and 3.8 to available versions
- Add 3.7 and 3.8 to the full std library test
- Remove tests dedicated to 3.1 - 3.4

### DIFF
--- a/src/Analysis/Engine/Impl/Parsing/PythonLanguageVersion.cs
+++ b/src/Analysis/Engine/Impl/Parsing/PythonLanguageVersion.cs
@@ -36,7 +36,8 @@ namespace Microsoft.PythonTools.Parsing {
         V34 = 0x0304,
         V35 = 0x0305,
         V36 = 0x0306,
-        V37 = 0x0307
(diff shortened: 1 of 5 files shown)


jakebailey:
3.4 might still be good to keep around. I found the page that lists what the Python devs are actually supporting: https://devguide.python.org/#status-of-python-branches

jakebailey:
Although they do note that security releases don't come with binary builds, which means it's a bit more difficult for us to actually test them at their latest versions, so maybe dropping at least 3.4 isn't such a big deal.

MikhailArkhipov:
@jakebailey - it is not necessarily what is supported but rather how many users are on it balanced with the length of the test run.

Issue #273, opened 2018-10-17
Analysis may be incorrect depending on timing/module load order


MikhailArkhipov:
When analyzing `mod1` that imports `mod2` code in
```cs
            var reanalyzeEntries = aliases.SelectMany(a => Analyzer.GetEntriesThatImportModule(a, true)).ToArray();
            foreach (var entryRef in reanalyzeEntries) {
                await EnqueueItemAsync(entryRef as IDocument, AnalysisPriority.Low, parse: false);
            }
```
assumes that analysis of other modules is complete and `GetEntriesThatImportModule` will return the correct set. However, if analysis of dependencies hasn't been completed yet, the set may be incomplete. So in find reference tests code that loads dependency first like
```cs
                var uri2 = await TestData.CreateTestSpecificFileAsync("mod2.py", text2);
                using (server.AnalysisQueue.Pause()) {
                    await server.LoadFileAsync(uri2);
                    uri1 = await server.OpenDocumentAndGetUriAsync("mod1.py", text1);
                }
```
will succeed because mod2 will be placed in the queue first, but code that loads dependency last
```cs
                var uri2 = await TestData.CreateTestSpecificFileAsync("mod2.py", text2);
                using (server.AnalysisQueue.Pause()) {
                    uri1 = await server.OpenDocumentAndGetUriAsync("mod1.py", text1);
                    await server.LoadFileAsync(uri2);
                }
```
will lead to incomplete analysis with references missing

MikhailArkhipov:
This requires fair amount of work to build dependency graph and re-analyze from there.

MikhailArkhipov:
#135

PR #274, opened 2018-10-17
Fix MutatingReferences test

### LINKED ISSUE #47: MutatingReferences test fails
Looks like we are only looking into variables in a single file. We are not inspecting all files that may be referencing types from the file the user is is working with. I.e. FindReferences request comes with specific URI but we don't look into other files.

MikhailArkhipov:
Fixed #47

Note:  https://github.com/Microsoft/python-language-server/issues/273 is related, but requires different fix.

- Added calculation of dependent set dropping cycles
- Added test


src/Analysis/Engine/Impl/Infrastructure/Extensions/HashSetExtensions.cs
```diff
@@ -0,0 +1,29 @@
...
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.PythonTools.Analysis.Infrastructure.Extensions {
+    public static class HashSetExtensions {
+        public static void Add<T>(this HashSet<T> set, IEnumerable<T> items) {
```
jakebailey: Isn't this just `HashSet.UnionWith`?

MikhailArkhipov: Good point

MikhailArkhipov:
Actually, we need to resolve https://github.com/Microsoft/python-language-server/issues/273 first

Issue #278, opened 2018-10-18
Huge memory usage when analysis 9 lines python code


jxww:
Hi everyone,

I am writing some python code using vscode with python-language-server extension and just find out  `Microsoft.Python.LanguageServer.exe` eat all of my memory (over 8g) when I open a specific file. I finnally reduced the problematic file to the following 9 lines and the language server still use over 2.6g memory. 

Python code:

    class A():
        def test(self, test):
            return test
    class B():
        def test(self, test):
            return test
    class C():
        def test(self, test):
            return test

The interesting thing is that if I remove any of the class/method or change any of the `test` token in the code, then the language server will run normally with only 100+m memory cost.

My environment is as follow:

- vscode 1.28.2
- python extension 2018.9.0 with language server 0.1.42
- windows 10 pro 64 bit
- python 3.6

![image](https://user-images.githubusercontent.com/26917526/47133809-f6cbf580-d2dc-11e8-95e5-8bdcdf08f428.png)
![image](https://user-images.githubusercontent.com/26917526/47133818-021f2100-d2dd-11e8-9f5a-cdf9e39fbe44.png)

I think the may be something wrong in code analysis. Anyone can fix it? Thanks very much.

jakebailey:
I can reproduce this against both the released version (on my personal machine) and master (on my dev machine). Sometimes I can cut and paste the code a few times to make it go past the ~2.5GB of usage, or even drop down to ~40MB.

I did a memory snapshot when usage is high, and it seems like most of the memory is storing JSON-related objects (`Newtonsoft.Json.Linq.Property`, `JValue`, `JObject`, dicts/arrays of those objects). AFAIK the only place we're using JSON is for RPC.

Actually looking at some of the objects shows that they're cyclic. Maybe they're not getting GC'd properly.

jakebailey:
I'll look at this more tomorrow (so it's not midnight), but after following the "path to root" for many of these objects, they all seem to lead up to Actions in StreamJsonRpc which are holding onto JsonRpcMessages with their status set to "RanToCompletion".

jakebailey:
#280 fixes it. Your example now sits at a nice 26MB on my system. 😄 

Thanks for providing a concise way to reproduce the issue.

jxww:
Thank you for all of your nice and quick response. 
I updated my python extension to 2018.9.1(with python language server 0.1.42) just now and still found this problem. Is this fix merged to the new release version?

[...]

Issue #109, opened 2018-09-18
Reloading modules leaks memory


MikhailArkhipov:
Clone [dups](https://github.com/linuxwhatelse/dups).
Open it
Open few files, do some random go to definitions.
Leave product inactive

Observe RAM growth and some odd module reloading

![image](https://user-images.githubusercontent.com/12820357/45721234-10eaaa80-bb5c-11e8-89bd-1c1bba988e83.png)

heyakyra:
Found this bug via https://github.com/Microsoft/vscode-python/issues/2594

Is there a workaround for this? I'm basically blocked at work because all 16G ram + 8G swap gets completely consumed after half an hour of editing

MikhailArkhipov:
@heyakyra - you are probably hitting https://github.com/Microsoft/python-language-server/issues/193. This one is not too bad.

MikhailArkhipov:
Depends on https://github.com/Microsoft/python-language-server/issues/209

jakebailey:
@GinRyan Unless you can see in your output "reloading modules" repeated while memory usage grows, your issue is likely unrelated to the original bug reported here. If you believe you have an issue we can reproduce, I'd suggest making a new issue with more information instead of commenting on older issue like this one and #109.

PR #280, opened 2018-10-18
Several bug fixes

AlexanderSher:
- Fix issue #95: `from ... import` shows incorrect completions. `TryGetCompletionsInFromImport` didn't use information from `ModuleInfo._referencedModules`, trying to find module directly inside AST.
- Fix issue #109: Reloading modules leaks memory. `ModuleInfo.Clear` didn't remove back references from `ModuleReference` instances
- Fix issue #278: Huge memory usage when analysis 9 lines python code. Issue caused by Json.Net serializing a cycle in `DocumentSymbol` tree. `Server.ToDocumentSymbols` uses `Dictionary<IMemberResult, List<IMemberResult>>` to make a tree, and `MemberResult`'s `GetHashCode` and `Equals` didn't take care of the scope. As a result, all results with equal parent name were collapsed into one level, which created a loop in a resulting tree where Json.Net has been hanging.
- Some code clean-up (dead code paths, auto-properties, pattern matching, etc.)


jakebailey:
I can confirm that this fixes #278. Nice!

src/Analysis/Engine/Impl/Parsing/Ast/DottedName.cs
```diff
@@ -22,16 +22,14 @@ namespace Microsoft.PythonTools.Parsing.Ast {
     public class DottedName : Node {
         private readonly NameExpression[] _names;
 
-        public DottedName(NameExpression[] names) {
+        public DottedName(NameExpression[]/*!*/ names) {
```
jakebailey: What's the meaning of `/*!*/` introduced in a few of these files?

src/Analysis/Engine/Impl/Values/ModuleInfo.cs
```diff
@@ -172,22 +161,27 @@ internal class ModuleInfo : AnalysisValue, IReferenceableContainer, IModuleInfo
...
+        public void AddModuleReference(string moduleName, ModuleReference moduleReference) {
+            Check.ArgumentNotNull(nameof(moduleName), moduleName);
+            Check.ArgumentNotNull(nameof(moduleReference), moduleReference);
+
+            _referencedModules[moduleName] = moduleReference;
+            moduleReference.AddReference(this);
+        }
+
+        public void ClearReferencedModules() {
+            foreach (var moduleReference in _referencedModules.Values.Distinct()) {
```
MikhailArkhipov: Not sure if Values can be null (when module is missing) - perhaps `MaybeEnumerate`?

src/Analysis/Engine/Impl/Parsing/Ast/DottedName.cs
AlexanderSher: Just a comment that I've used to mark parameter that can't be null. I can add `Check` calls instead.

src/Analysis/Engine/Impl/Values/ModuleInfo.cs
AlexanderSher: `_referencedModules` is a dictionary, it is always initialized and it is readonly.

src/Analysis/Engine/Impl/Parsing/Ast/DottedName.cs
jakebailey: You don't have to change it, I just hadn't seen that before and was curious. Grepping through the project shows its use a number of other times as well (so I just hadn't noticed).

AlexanderSher: To be clear - this is just a comment. It isn't handled by any analyzers. I really expect to see non-nullable reference types in C# 8: https://github.com/dotnet/csharplang/wiki/Nullable-Reference-Types-Preview

jakebailey: Yep, I know. I actually went searching to see if that feature existed yet when you mentioned it. 🙂

Issue #261, opened 2018-10-15
Better distinguish override completions in tooltip


jakebailey:
Right now, completions for function overrides look like:

![image](https://user-images.githubusercontent.com/5341706/46974060-d3ecd780-d077-11e8-8426-c5ff2318c06b.png)

I don't think that it's very clear that selecting this completion will actually fill out a full function rather than just completing the name for you (which is what jedi does with the same exact hint name). I think it'd be better to include some extra text in the tooltip, like:

![image](https://user-images.githubusercontent.com/5341706/46974449-dc91dd80-d078-11e8-8cf7-a7e5a1435b6d.png)

To better show what the completion actually does.

jakebailey:
FWIW, the above is just:

```csharp
private CompletionItem ToOverrideCompletionItem(IOverloadResult o, ClassDefinition cd, string indent) {
    return new CompletionItem {
        label = o.Name + " (override)",
        insertText = MakeOverrideCompletionString(indent, o, cd.Name),
        kind = CompletionItemKind.Method
    };
}
```

PR #283, opened 2018-10-18
Added an option to propagate parameter types to base methods

lostmsu:
It is similar #194, in that the same types parameter types might be OK for base methods. It could happen when we just don't have any examples in the current code, that would give that kind of information to analyzer.

By default set to `false` to keep conservative behavior.

```python
class Base:
  def foo(x):
    ...

class Derived(Base):
  def foo(x):
    ...

Derived().foo(42)

# Base().foo(42) is actually also called, but not in this project
```

### DIFF
--- a/src/Analysis/Engine/Impl/AnalysisLimits.cs
+++ b/src/Analysis/Engine/Impl/AnalysisLimits.cs
@@ -181,6 +181,12 @@ namespace Microsoft.PythonTools.Analysis {
         public bool ProcessCustomDecorators { get; set; }
 
         /// <summary>
+        /// <c>True</c> to propagate parameter types to base methods.
+        /// Parameter types always propagate to derived methods.
+        /// </summary>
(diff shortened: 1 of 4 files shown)


src/Analysis/Engine/Impl/AnalysisLimits.cs
```diff
@@ -180,6 +180,12 @@ public sealed class AnalysisLimits {
         /// </summary>
         public bool ProcessCustomDecorators { get; set; }
 
+        /// <summary>
+        /// <c>True</c> to propagate parameter types to base methods.
+        /// Parameter types always propagate to derived methods.
+        /// </summary>
+        public bool PropagateParameterTypeToBaseMethods { get; set; }
```
MikhailArkhipov: Is there case when we don't want this?

MikhailArkhipov: I mean, it is not something that user sets so it it typically set by the application one way or another. #261 may be related.

PR #284, opened 2018-10-18
Fix ApplicableSpan calculation

### LINKED ISSUE #23: Exception `Start and End must be well ordered`

MikhailArkhipov:
Fixes #23 

- With whitespace surrounding dot span of the expression can be to the right of the dot. Adjust span accordingly.
- Add test for the case (reproes only when completion context is supplied)

### DIFF
--- a/src/Analysis/Engine/Test/CompletionTests.cs
+++ b/src/Analysis/Engine/Test/CompletionTests.cs
@@ -1003,6 +1003,15 @@ class Simple(unittest.TestCase):
             }
         }
 
+        [TestMethod, Priority(0)]
+        public async Task WithWhitespaceAroundDot() {
+            using (var s = await CreateServerAsync()) {
+                var u = await s.OpenDefaultDocumentAndGetUriAsync("import sys\nsys  .  version\n");
+                await AssertCompletion(s, u, new[] { "argv" }, null, new SourceLocation(2, 7), 
+                    new CompletionContext { triggerCharacter = ".", triggerKind = CompletionTriggerKind.TriggerCharacter });
+            }
+        }
+
         private static async Task AssertCompletion(Server s, Uri uri, IReadOnlyCollection<string> contains, IReadOnlyCollection<string> excludes, Position? position = null, CompletionContext? context = null, Func<CompletionItem, string> cmpKey = null, string expr = null, Range? applicableSpan = null) {
             await s.WaitForCompleteAnalysisAsync(CancellationToken.None);
             var res = await s.Completion(new CompletionParams {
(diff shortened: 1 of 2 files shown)


src/LanguageServer/Impl/Implementation/Server.Completion.cs
```diff
@@ -38,7 +40,7 @@ public sealed partial class Server {
             var analysis = entry != null ? await entry.GetAnalysisAsync(50, cancellationToken) : null;
             if (analysis == null) {
                 TraceMessage($"No analysis found for {uri}");
-                return new CompletionList();
+                return EmptyCompletion;
```
AlexanderSher: It is not safe to return shared instance cause `CompletionList` isn't immutable,

MikhailArkhipov: OK

MikhailArkhipov: We do return hover though

AlexanderSher: Considering amount of objects created by Json.Net, another instance won't make a big difference.

src/Analysis/Engine/Test/CompletionTests.cs
```diff
@@ -1003,6 +1003,15 @@ class Simple(unittest.TestCase):
             }
         }
 
+        [TestMethod, Priority(0)]
+        public async Task WithWhitespaceAroundDot() {
+            using (var s = await CreateServerAsync()) {
+                var u = await s.OpenDefaultDocumentAndGetUriAsync("import sys\nsys  .  version\n");
```
AlexanderSher: Do we care about `\r\n`?

src/LanguageServer/Impl/Implementation/Server.Completion.cs
```diff
@@ -70,39 +69,50 @@ public sealed partial class Server {
...
+                return res;
+            }
+
+            await InvokeExtensionsAsync((ext, token)
+                => (ext as ICompletionExtension)?.HandleCompletionAsync(uri, analysis, tree, @params.position, res, cancellationToken), cancellationToken);
+
+            return res;
+        }
+
+        private SourceSpan? GetApplicableSpan(CompletionAnalysis ca, CompletionParams @params, PythonAst tree) {
```
AlexanderSher: Does any of it affects `textEdit` field in the individual `CompletionItem`?

src/Analysis/Engine/Test/CompletionTests.cs
MikhailArkhipov: Not here, I believe there are parser tests for the line endings

src/LanguageServer/Impl/Implementation/Server.Completion.cs
MikhailArkhipov: Technically they should all have the same applicable span. The span is actually used only by VS to figure out what to delete and replace with the completion item insertion text.

PR #289, opened 2018-10-19
Improve diagnostic generation/handling, enable diagnostics with setting

### LINKED ISSUE #242: Show warning for undeclared variables
Also variables used before declaration

jakebailey:
Fixes #242.

This does the following:
- Fixes diagnostics when settings are changed at runtime. Previously, only the mapping from `code` to `severity` was updated, not the actual listing, so diagnostics were never re-run even when settings were changed.
- Moves all severity decision and diagnostic filtering upward into the LS. This means all generated diagnostics will be published as-is, and any future additions (i.e. import code actions) can be generated when new diagnostics are published.
- Adds a setting called `python.analysis.diagnostics` which is true when diagnostics should be published altogether (and also triggers an update when changed). ~The default behavior remains the same; users will not see any diagnostics until VSC adds the setting to its `package.json`. Devs (us) can set this in the JSON just like we set `downloadLanguageServer` even though it's undocumented.~ Default will be enabled, as per PR comments.

Looking for feedback on:
- The naming of `python.analysis.diagnostics`.
- Whether or not I should be removing `EnableDiagnostics` now that the toggle is focused on deciding whether or not to publish diagnostics, not whether or not to collect them.

### DIFF
--- a/src/LanguageServer/Impl/Implementation/Server.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.cs
@@ -696,12 +696,8 @@ namespace Microsoft.Python.LanguageServer.Implementation {
                 return true;
             }
 
-            if (newSettings.analysis.openFilesOnly != oldSettings.analysis.openFilesOnly) {
-                _editorFiles.UpdateDiagnostics();
-                return false;
-            }
-
-            if (!newSettings.analysis.errors.SetEquals(oldSettings.analysis.errors) ||
+            if (newSettings.analysis.openFilesOnly != oldSettings.analysis.openFilesOnly ||
+                !newSettings.analysis.errors.SetEquals(oldSettings.analysis.errors) ||
                 !newSettings.analysis.warnings.SetEquals(oldSettings.analysis.warnings) ||
                 !newSettings.analysis.information.SetEquals(oldSettings.analysis.information) ||
                 !newSettings.analysis.disabled.SetEquals(oldSettings.analysis.disabled)) {
--- a/src/LanguageServer/Impl/Implementation/Server.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.cs
@@ -250,6 +250,7 @@ namespace Microsoft.Python.LanguageServer.Implementation {
             var reanalyze = true;
             if (@params.settings != null) {
                 if (@params.settings is ServerSettings settings) {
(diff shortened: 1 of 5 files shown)


src/LanguageServer/Impl/LanguageServer.cs
```diff
@@ -145,6 +145,7 @@ private void OnUnregisterCapability(object sender, UnregisterCapabilityEventArgs
                 settings.diagnosticPublishDelay = GetSetting(analysis, "diagnosticPublishDelay", 1000);
                 settings.symbolsHierarchyDepthLimit = GetSetting(analysis, "symbolsHierarchyDepthLimit", 10);
                 settings.symbolsHierarchyMaxSymbols = GetSetting(analysis, "symbolsHierarchyMaxSymbols", 1000);
+                settings.analysis.diagnostics = GetSetting(analysis, "diagnostics", false);
```
MikhailArkhipov: Default should be `true` as the setting is not in the `package.json` yet

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -250,6 +250,7 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
             var reanalyze = true;
             if (@params.settings != null) {
                 if (@params.settings is ServerSettings settings) {
+                    Analyzer.EnableDiagnostics = settings.analysis.diagnostics;
```
MikhailArkhipov: The setting needs to be added to the `package.json` in the extension defaulting to `true`

jakebailey: Yep, I've done that locally for now (until I PR it), since I wasn't 100% sure on the naming.

src/LanguageServer/Impl/LanguageServer.cs
jakebailey: Okay, though that will mean that VSC will double-up on a couple of diagnostics since they still run pylint for everything.

src/LanguageServer/Impl/Implementation/Server.cs
MikhailArkhipov: @DonJayamanne can suggest name of the setting

jakebailey:
I'm going to close this in favor of #298. Adding this setting complicates things in terms of how to deal with parsing diagnostics that I'm not confident this change can properly handle.

PR #295, opened 2018-10-22
[WIP] Use lazy members in AST walks

### LINKED ISSUE #294: Analysis of return values does not work with forward references
`Properties.py`:
```python
class A(object):
    def methodA(self):
        return 's'

class B(object):
    def getA(self):
        return self.propA

    @property
    def propA(self):
        return A()
```
Code:
```python
x = B().getA().methodA()
```

x is nothing. This is b/c after AST walker processes class it processes functions in queue order and property `propA` is not yet known when return type of `getA` is analyzed.

Code needs to use delayed resolution or relocation tables.

MikhailArkhipov:
Fixes #294.

Use lazy members so types are resolved in the main analyzer while AST walker primarily creates lazy types.

### DIFF
--- a/src/Analysis/Engine/Impl/Interpreter/Ast/AstAnalysisFunctionWalker.cs
+++ b/src/Analysis/Engine/Impl/Interpreter/Ast/AstAnalysisFunctionWalker.cs
@@ -37,7 +37,6 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
             _overload = overload ?? throw new ArgumentNullException(nameof(overload));
         }
 
-        public IList<IPythonType> ReturnTypes => _overload.ReturnTypes;
         public IPythonFunctionOverload Overload => _overload;
 
         private void GetMethodType(FunctionDefinition node, out bool classmethod, out bool staticmethod) {
@@ -52,7 +51,7 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
             var classmethodObj = _scope.Interpreter.GetBuiltinType(BuiltinTypeId.ClassMethod);
             var staticmethodObj = _scope.Interpreter.GetBuiltinType(BuiltinTypeId.StaticMethod);
             foreach (var d in (_target.Decorators?.Decorators).MaybeEnumerate().ExcludeDefault()) {
-                var m = _scope.GetValueFromExpression(d);
+                var m = _scope.GetValueFromExpression(d).ResolveType();
                 if (m == classmethodObj) {
                     classmethod = true;
                 } else if (m == staticmethodObj) {
@@ -65,7 +64,7 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
             var self = GetSelf();
             _selfType = (self as AstPythonConstant)?.Type as AstPythonType;
 
-            _overload.ReturnTypes.AddRange(_scope.GetTypesFromAnnotation(_target.ReturnAnnotation).ExcludeDefault());
+            _overload.AddReturnTypes(_scope.GetTypesFromAnnotation(_target.ReturnAnnotation).ExcludeDefault().OfType<IMember>());
 
             _scope.PushScope();
             if (self != null) {
@@ -163,9 +162,8 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
         }
 
         public override bool Walk(ReturnStatement node) {
-            foreach (var type in _scope.GetTypesFromValue(_scope.GetValueFromExpression(node.Expression)).ExcludeDefault()) {
-                _overload.ReturnTypes.Add(type);
-            }
+            // Must use lazy type since actual type may not be known until all classes and methods are walked.
+            _overload.AddReturnType(_scope.Clone().GetValueFromExpression(node.Expression));
             return true; // We want to evaluate all code so all private variables in __new__ get defined
         }
 
--- a/src/Analysis/Engine/Impl/Interpreter/Ast/AstAnalysisFunctionWalker.cs
+++ b/src/Analysis/Engine/Impl/Interpreter/Ast/AstAnalysisFunctionWalker.cs
@@ -64,7 +64,7 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
             var self = GetSelf();
             _selfType = (self as AstPythonConstant)?.Type as AstPythonType;
 
-            _overload.AddReturnTypes(_scope.GetTypesFromAnnotation(_target.ReturnAnnotation).ExcludeDefault().OfType<IMember>());
+            _overload.AddReturnTypes(_scope.GetTypesFromAnnotation(_target.ReturnAnnotation).OfType<IMember>());
 
             _scope.PushScope();
             if (self != null) {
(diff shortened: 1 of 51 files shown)


src/Analysis/Engine/Impl/Interpreter/Ast/AstLazyMember.cs
```diff
@@ -0,0 +1,36 @@
...
+        [DebuggerStepThrough]
+        public AstLazyMember(Func<IMember> getter) {
+            _getter = getter;
+        }
+        public PythonMemberType MemberType => PythonMemberType.Unknown;
+
+        public string Name => "lazy";
+
+        [DebuggerStepThrough]
+        public IMember Get() => _value ?? (_value = _getter());
```
AlexanderSher: Should we check if `_getter` is called inside `AnalysisQueue`?

src/Analysis/Engine/Impl/Interpreter/Ast/AstScope.cs
```diff
@@ -299,74 +325,85 @@ sealed class NameLookupContext {
...
 
-            var m = GetValueFromExpression(expr.Target);
-            if (m is IPythonType type) {
-                if (type.TypeId == BuiltinTypeId.Type && type == Interpreter.GetBuiltinType(BuiltinTypeId.Type) && expr.Args.Count >= 1) {
-                    var aType = GetTypeFromValue(GetValueFromExpression(expr.Args[0].Expression));
-                    if (aType != null) {
-                        return aType;
+            return new AstLazyMember(() => {
+                var m = GetValueFromExpression(expr.Target).ResolveType();
+                if (m is IPythonType type) {
```
AlexanderSher: `switch`

src/Analysis/Engine/Impl/Interpreter/Ast/AstScope.cs
```diff
@@ -525,17 +562,14 @@ sealed class NameLookupContext {
                 return null;
             }
 
-            IMember obj;
-            if ((scope ?? _scopes.Peek()).TryGetValue(name, out obj)) {
+            if ((scope ?? _scopes.Peek()).TryGetValue(name, out var obj)) {
```
AlexanderSher: Can `_scopes` be empty?

src/Analysis/Engine/Impl/Interpreter/Ast/AstAnalysisFunctionWalker.cs
```diff
@@ -52,7 +51,7 @@ AstPythonFunctionOverload overload
             var classmethodObj = _scope.Interpreter.GetBuiltinType(BuiltinTypeId.ClassMethod);
             var staticmethodObj = _scope.Interpreter.GetBuiltinType(BuiltinTypeId.StaticMethod);
             foreach (var d in (_target.Decorators?.Decorators).MaybeEnumerate().ExcludeDefault()) {
-                var m = _scope.GetValueFromExpression(d);
+                var m = _scope.GetValueFromExpression(d).ResolveType();
```
AlexanderSher: So what is a criteria of calling `ResolveType`?

src/Analysis/Engine/Impl/Interpreter/Ast/AstAnalysisFunctionWalker.cs
```diff
@@ -65,7 +64,7 @@ AstPythonFunctionOverload overload
             var self = GetSelf();
             _selfType = (self as AstPythonConstant)?.Type as AstPythonType;
 
-            _overload.ReturnTypes.AddRange(_scope.GetTypesFromAnnotation(_target.ReturnAnnotation).ExcludeDefault());
+            _overload.AddReturnTypes(_scope.GetTypesFromAnnotation(_target.ReturnAnnotation).ExcludeDefault().OfType<IMember>());
```
AlexanderSher: You don't need `ExcludeDefault()` - `OfType<IMember>()` will remove nulls.

src/Analysis/Engine/Impl/Interpreter/Ast/AstAnalysisFunctionWalker.cs
MikhailArkhipov: Generally in AST walk types should be lazy to the very end unless it is trivial to resolve. Otherwise, ResolveType should be called

- in delayed actions which are actually invoked to resolve type
- when code needs actual type, as the snippet above
- After AST walk I am forcibly resolving types in all scopes (pending change)

src/Analysis/Engine/Impl/Interpreter/Ast/AstLazyMember.cs
MikhailArkhipov: To make sure it is resolved?

AlexanderSher: And to avoid unexpected concurrent calls. `DDG` is executed only inside `AnalysisQueue`. Do we expect `ILazyMember.Get()` to be called not from analyzer?

MikhailArkhipov: Yes, it can be called at the end of AST walks and also is called in tests without analyzer present

src/Analysis/Engine/Impl/Interpreter/Ast/AstScope.cs
```diff
@@ -25,14 +25,15 @@
 using Microsoft.PythonTools.Parsing.Ast;
 
 namespace Microsoft.PythonTools.Interpreter.Ast {
-    sealed class NameLookupContext {
+    sealed class AstScope {
         private readonly Stack<Dictionary<string, IMember>> _scopes = new Stack<Dictionary<string, IMember>>();
+        private readonly List<Dictionary<string, IMember>> _allScopes = new List<Dictionary<string, IMember>>();
```
AlexanderSher: Need better name for the field

src/Analysis/Engine/Impl/Interpreter/Ast/AstScope.cs
```diff
@@ -71,8 +72,8 @@ sealed class NameLookupContext {
         public LookupOptions DefaultLookupOptions { get; set; }
         public bool SuppressBuiltinLookup { get; set; }
 
-        public NameLookupContext Clone(bool copyScopeContents = false) {
-            var ctxt = new NameLookupContext(
+        public AstScope Clone(bool copyScopeContents = false) {
+            var ctxt = new AstScope(
```
AlexanderSher: Should we transfer `_allScopes` during `Clone`?

src/Analysis/Engine/Impl/Interpreter/Ast/AstScope.cs
```diff
@@ -215,7 +225,7 @@ sealed class NameLookupContext {
                 return null;
             }
 
-            IMember existing = LookupNameInScopes(expr.Name, options);
+            var existing = LookupNameInScopes(expr.Name, options);
```
AlexanderSher: While you're here, can you check if we can implement `GetValueFromExpression(Expression expr, LookupOptions options)` using pattern matching instead of `??` flow:
```cs
switch (expr) {
    case NameExpression name: return GetValueFromName(expr as NameExpression, options) ;
/*...*/
    case null: return null;
}
```

It is much easier to debug and we can save some of the null checks inside `GetValueFrom*` methods

PR #297, opened 2018-10-22
Add telemetry event structure

MikhailArkhipov:
Matching https://github.com/Microsoft/vscode-python/pull/2978

### DIFF
--- a/src/LanguageServer/Impl/Definitions/ITelemetryService.cs
+++ b/src/LanguageServer/Impl/Definitions/ITelemetryService.cs
@@ -14,10 +14,17 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace Microsoft.Python.LanguageServer {
     public interface ITelemetryService {
-        Task SendTelemetry(object o);
+        Task SendTelemetry(TelemetryEvent telemetryEvent);
+    }
+
+    public sealed class TelemetryEvent {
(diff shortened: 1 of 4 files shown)


src/LanguageServer/Impl/Definitions/ITelemetryService.cs
```diff
@@ -14,10 +14,17 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 namespace Microsoft.Python.LanguageServer {
     public interface ITelemetryService {
-        Task SendTelemetry(object o);
+        Task SendTelemetry(TelemetryEvent telemetryEvent);
```
AlexanderSher: What about implementation?

MikhailArkhipov: Doh

src/LanguageServer/Impl/LanguageServer.cs
```diff
@@ -99,7 +97,8 @@ private class PublishDiagnosticsParams {
         }
 
         #region Events
-        private void OnTelemetry(object sender, TelemetryEventArgs e) => _telemetry.SendTelemetry(e.value);
+        [Obsolete]
+        private void OnTelemetry(object sender, TelemetryEventArgs e) { }
```
AlexanderSher: You can simply remove it - it is private.

MikhailArkhipov: Yep

PR #298, opened 2018-10-22
Only enable diagnostics and fix settings change detection

### LINKED ISSUE #242: Show warning for undeclared variables

jakebailey:
Fixes #242.

#289 causes a few issues to do with parser-generated diagnostics, and there are more diagnostics I wasn't aware of coming from ErrorSink. This is a simplified PR that does the minimum to enable diagnostics (ignoring "Unspecified" at publish time) and fixes the setting change bug.

We'll have to revisit where diagnostics should be stored to access via code actions in a different PR. There are only two diagnostics that actually come out of the server which can be disabled with `python.analysis.disabled` for now.

I'll close #289 in favor of this one.

### DIFF
--- a/src/Analysis/Engine/Test/LanguageServerTests.cs
+++ b/src/Analysis/Engine/Test/LanguageServerTests.cs
@@ -28,6 +28,7 @@ using Microsoft.Python.LanguageServer.Implementation;
 using Microsoft.Python.Tests.Utilities.FluentAssertions;
 using Microsoft.PythonTools;
 using Microsoft.PythonTools.Analysis;
+using Microsoft.PythonTools.Analysis.Analyzer;
 using Microsoft.PythonTools.Analysis.FluentAssertions;
 using Microsoft.PythonTools.Analysis.Infrastructure;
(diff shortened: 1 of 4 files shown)


MikhailArkhipov:
Test would be good. Like declare a variable, observer squiggly, change file, squiggly goes away, change back, it reappears. @AlexanderSher - do we have something like this for unresolved imports?

jakebailey:
Okay, maybe not as simple as I thought. The test helper `SendDidChangeConfiguration` only accepts completion options, then reimplements part of `DidChangeConfiguration`, specifically the part that I'm trying to test... So I'll have to do some extra work to actually give the server a proper config change message to run the real code.

jakebailey:
I've added a test which changes the settings and re-checks the diagnostics. The overall diagnostic testing code and settings editing code is really strange, though, and could probably be improved.

All of the "work" for changing settings is actually done in `ServerBase` at the RPC method level on the JToken. It'd be nice if all of that hand-written JToken decoding was done automatically so that the tests could actually call it...

PR #303, opened 2018-10-23
Upgrade StreamJsonRpc library to 2.0 beta for error message modification

jakebailey:
Alternative to #276, related to #246.

This allows for the error message's data to be swapped out for the stack trace.

To do telemetry, we'd need to use `rpc.TraceSource` to track error events, which is not in this PR.

I figured I'd make a WIP PR instead of continuing to link to a diff.


src/LanguageServer/Impl/Microsoft.Python.LanguageServer.ruleset
```diff
@@ -0,0 +1,92 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
AlexanderSher: I think we can make one global ruleset in `src` folder. @MikhailArkhipov , what do you think?

MikhailArkhipov: Makes sense

jakebailey: I think some of these messages may actually be useful in some cases, so I'll update this with the ruleset move and change these to informational instead of completely disabled.

jakebailey:
I've moved the ruleset up to src, and also enabled all of the messages as Info except for the Async suffix one (because it adds 500+ messages because so many functions don't do this, especially tests). We can modify this to be more granular in the future if we want to respect the Async suffix guideline (and then disable for tests only).

PR #304, opened 2018-10-24
Fix processing of forward references

### LINKED ISSUE #294: Analysis of return values does not work with forward references

MikhailArkhipov:
Fixes #294 

- Implement ordered processing of functions and properties. If return type is not known, process walker for that function which, in turn, can process more.
- Fix some comments and clean up code.


src/Analysis/Engine/Impl/Interpreter/Ast/AstAnalysisFunctionWalkerSet.cs
```diff
@@ -0,0 +1,56 @@
...
+using Microsoft.PythonTools.Parsing.Ast;
+
+namespace Microsoft.PythonTools.Interpreter.Ast {
+    /// <summary>
+    /// Represents set of function body walkers. Functions are walked after
+    /// all classes are collected. If function or property return type is unknown,
+    /// it can be walked, and so on recursively, until return type is determined
+    /// or there is nothing left to walk.
+    /// </summary>
+    class AstAnalysisFunctionWalkerSet {
```
AlexanderSher: Do we care about thread safety of this type?

MikhailArkhipov: Module analysis is a single thread as far as I can tell, begins with `new AstPythonModule`. Existing code didn't seem to use locks.

src/Analysis/Engine/Impl/Interpreter/Ast/NameLookupContext.cs
```diff
@@ -346,27 +354,61 @@ sealed class NameLookupContext {
...
+            if (value == null) {
+                _log?.Log(TraceLevel.Verbose, "UnknownCallable", expr.Target.ToCodeString(Ast).Trim());
+                return new AstPythonConstant(_unknownType, GetLoc(expr));
+            }
+            return value;
+        }
+
+        private IMember GetValueFromPropertyOrFunction(IMember fn, Expression expr) {
+            switch (fn) {
+                case IPythonBoundFunction bf:
```
AlexanderSher: If we move type check into `case`:
```cs
case IPythonBoundFunction bf when bf.Function as IPythonFunction2 f: 
    GetValueFromFunction(f, expr);
```
we can get rid of null check in the `GetValueFromFunction`.

src/Analysis/Engine/Impl/Interpreter/Ast/NameLookupContext.cs
```diff
@@ -346,27 +354,61 @@ sealed class NameLookupContext {
...
+                    return GetValueFromFunction(bf.Function as IPythonFunction2, expr);
+                case IPythonFunction2 f:
+                    return GetValueFromFunction(f, expr);
+                case IBuiltinProperty2 p:
+                    return GetPropertyReturnType(p, expr);
+            }
+            return null;
```
AlexanderSher: Shouldn't we return `AstPythonConstant` here?

src/Analysis/Engine/Impl/Interpreter/Ast/NameLookupContext.cs
```diff
@@ -346,27 +354,61 @@ sealed class NameLookupContext {
...
 
             var m = GetValueFromExpression(expr.Target);
-            if (m is IPythonType type) {
-                if (type.TypeId == BuiltinTypeId.Type && type == Interpreter.GetBuiltinType(BuiltinTypeId.Type) && expr.Args.Count >= 1) {
-                    var aType = GetTypeFromValue(GetValueFromExpression(expr.Args[0].Expression));
-                    if (aType != null) {
-                        return aType;
+            IMember value = null;
+            switch (m) {
+                case IPythonType type:
```
AlexanderSher: Can save one level of nesting:
```cs
case IPythonType type when type == Interpreter.GetBuiltinType(BuiltinTypeId.Type) && expr.Args.Count >= 1:
    value = GetTypeFromValue(GetValueFromExpression(expr.Args[0].Expression, options));
    break;
case IPythonType type:
    return new AstPythonConstant(type, GetLoc(expr));
```

src/Analysis/Engine/Impl/Interpreter/Ast/NameLookupContext.cs
```diff
@@ -228,28 +224,40 @@ sealed class NameLookupContext {
...
+                        .Select(x => x.GetMember(Context, expr.Name))
+                        .ExcludeDefault()
+                        .FirstOrDefault();
+                    break;
+                default:
+                    value = GetValueFromPropertyOrFunction(e, expr);
+                    break;
+            }
+            if(value is IBuiltinProperty2 p) {
+                value = GetPropertyReturnType(p, expr);
```
AlexanderSher: `GetPropertyReturnType` never returns null, so `value == null` can be tested earlier.

src/Analysis/Engine/Impl/Interpreter/Ast/NameLookupContext.cs
MikhailArkhipov: Sure

src/Analysis/Engine/Impl/Interpreter/Ast/NameLookupContext.cs
MikhailArkhipov: `GetValueFromMember` and `GetValueFromCallable` and `GetValueFromExpression` expect nulls for logging purposes I guess. However, I don't know if there is much value in it.

MikhailArkhipov: I think point of returning nulls is to reduce amount of 'unknowns' objects. Ex if we return unknown constant it gets stuck in return types while null is skipped by `ExcludeDefault` clause

src/Analysis/Engine/Impl/Interpreter/Ast/AstAnalysisFunctionWalkerSet.cs
```diff
@@ -0,0 +1,56 @@
...
+            // Do not use foreach since walker list is dynamically
+            // modified and walkers are removed as they are done.
+            while (_functionWalkers.Count > 0) {
+                var walker = _functionWalkers.First().Value;
+                ProcessWalker(walker);
+            }
+        }
+
+        public void ProcessFunction(FunctionDefinition fn) {
+            if (_functionWalkers.TryGetValue(fn, out var w)) {
```
AlexanderSher: Should we remove walker from the dictionary if it was used already?

src/Analysis/Engine/Impl/Interpreter/Ast/NameLookupContext.cs
MikhailArkhipov: Perhaps we should stick with nulls to save some memory?

src/Analysis/Engine/Impl/Interpreter/Ast/AstAnalysisFunctionWalkerSet.cs
MikhailArkhipov: `ProcessWalker` removes it

src/Analysis/Engine/Impl/Interpreter/Ast/NameLookupContext.cs
MikhailArkhipov: I am actually cleaning up and returning more nulls now.

AlexanderSher: Ah, so that is intentional. Then ok.

src/Analysis/Engine/Impl/Interpreter/Ast/AstAnalysisFunctionWalkerSet.cs
```diff
@@ -0,0 +1,56 @@
...
+        private readonly Dictionary<FunctionDefinition, AstAnalysisFunctionWalker> _functionWalkers
+            = new Dictionary<FunctionDefinition, AstAnalysisFunctionWalker>();
+
+        public void Add(AstAnalysisFunctionWalker walker)
+            => _functionWalkers[walker.Target] = walker;
+
+        public void ProcessSet() {
+            // Do not use foreach since walker list is dynamically
+            // modified and walkers are removed as they are done.
+            while (_functionWalkers.Count > 0) {
```
AlexanderSher:
```cs
var walkers = _functionWalkers.Values.ToArray();
_functionWalkers.Clear();
foreach (var walker in walkers) {
    walker.Walk();
}
```

src/Analysis/Engine/Impl/Interpreter/Ast/AstAnalysisFunctionWalkerSet.cs
AlexanderSher: Ok

src/Analysis/Engine/Impl/Interpreter/Ast/NameLookupContext.cs
MikhailArkhipov: Well, it got worse. `AstPythonMultipleMembers.Combine` does not accept two null while it is actually possible. For safety reasons I will stick with unknowns for now.

PR #314, opened 2018-10-26
Add Exception telemetry, upgrade telemetry object to ITelemetryService2

### LINKED ISSUE #246: Add telemetry with the exception call stack

jakebailey:
Fixes #246.

This adds a TraceListener which will respond to the events that StreamJsonRpc sends through its TraceSource. It checks for the specific trace event ID that the RPC library uses, and checks to make sure the arguments are sane. Adding more listeners or changing the trace level should be safe and won't break the listener.

Property naming is up to us, since it's a new event, but I mirrored some of what I saw in the extension, that is, camelCase names. The event name is the same as what the extension is using for its language server telemetry, but with a `/` afterward so we don't have any name conflicts and can do our own thing. I picked `rpc.exception` for this event, but it can be changed if wanted.

### DIFF
--- a/src/LanguageServer/Impl/LanguageServer.cs
+++ b/src/LanguageServer/Impl/LanguageServer.cs
@@ -14,17 +14,17 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
-using System;
-using System.Collections.Generic;
-using System.Linq;
-using System.Threading;
-using System.Threading.Tasks;
 using Microsoft.Python.LanguageServer.Services;
 using Microsoft.PythonTools.Analysis;
 using Microsoft.PythonTools.Analysis.Infrastructure;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using StreamJsonRpc;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
 
 namespace Microsoft.Python.LanguageServer.Implementation {
     /// <summary>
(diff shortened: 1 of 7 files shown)


src/LanguageServer/Impl/LanguageServer.cs
```diff
@@ -14,17 +14,17 @@
...
-using System.Linq;
-using System.Threading;
-using System.Threading.Tasks;
 using Microsoft.Python.LanguageServer.Services;
 using Microsoft.PythonTools.Analysis;
 using Microsoft.PythonTools.Analysis.Infrastructure;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 using StreamJsonRpc;
+using System;
```
MikhailArkhipov: Reverse sorting (set it in VS Tools | Options | Text Editor | C# | Advanced)

src/LanguageServer/Impl/TelemetryRpcTraceListener.cs
```diff
@@ -0,0 +1,112 @@
...
+            }
+
+            // See JsonRpc.StripExceptionToInnerException.
+            if (exception is TargetInvocationException || (exception is AggregateException && exception.InnerException != null)) {
+                exception = exception.InnerException;
+            }
+
+            // Exceptions we expect to throw and pass back over RPC, and therefore should not record.
+            switch (exception) {
+                case InvalidOperationException _:
```
MikhailArkhipov: In `Rename` I'd change `InvalidOperationException` to `RenameException: NonFatalLanguageServerException` where `NonFatalLanguageServerException: LanguageServerException` and then filter out `NonFatalLanguageServerException`. This way actual bugs that cause `InvalidOperationException` will be logged.

src/LanguageServer/Impl/LanguageServer.cs
jakebailey: Bah, VS reset my settings again...

src/LanguageServer/Impl/TelemetryRpcTraceListener.cs
jakebailey: I've introduced a new exception and replaced Rename's use of InvalidOperationException with it.

PR #317, opened 2018-10-27
Fix TopLevelCompletions test on Mac

### LINKED ISSUE #310: TopLevelCompletions test hangs on Mac

MikhailArkhipov:
Fixes #310
Somehow backslashes creeped back in.


GrumpyR:
I didn't debug it deeply, but issue was that analysis was never completing and since backslash caused malformed URLs most probably process was continuously searching for modules and could never find any.

MikhailArkhipov:
Analysis was never completing and since backslash caused malformed URLs. Most probably process was continuously searching for modules and could never find any.

MikhailArkhipov:
Yes, I have too many changes in Q in local branches so I forked off master.

PR #321, opened 2018-10-29
Fix handling of generics and type declarations

### LINKED ISSUE #254: NewType does not subclass correctly
## Summary
Types created by `NewType` do not inherit any methods from their parents
## To Reproduce
```python
from typing import NewType

Foo = NewType("Foo", dict)

foo: Foo = Foo({})
foo.
```
## Expected Output
image
## Actual Output
image

## More Info
[The documentation for NewType](https://docs.python.org/3.6/library/typing.html#newtype) states

> The static type checker will treat the new type as if it were a subclass of the original type.

which is how Mypy behaves, but the language server treats `NewType` types as ...well, new types.
I noticed this because [TypedDict](https://mypy.readthedocs.io/en/latest/more_types.html#typeddict) doesn't work either, and while writing that issue I discovered this.

### LINKED ISSUE #260: Fix List[T] not inheriting list/List methods
Microsoft/vscode-python#2446

E.g.
```python
from typing import List

def func(a: List[str]):
    pass
```
Currently `a` has no completions instead of having completions for a list.

See [generics from PEP 484](https://www.python.org/dev/peps/pep-0484/#generics) for some details.

MikhailArkhipov:
Fixes #254 
Fixes #260

#254 - Declare protocol that delegates to actual type fully.
#260 - Declare ListProtocol and DictProtocol delegating *member* fetching to actual underlying types but leaving actual work to `Iterable` and `Mapping` protocols respectively as it used to be.
- Added tests for analysis and completion for the cases

Will split Protocols files into separate classes before merging


jakebailey:
You might want to add some tests for `TypeVar`, since you mention it in the comments. I think that'll be required for dealing with some of the typeshed generic types. Maybe something like this (modified from https://docs.python.org/3/library/typing.html#generics):

```python

from typing import Sequence, TypeVar

T = TypeVar('T')      # Declare type variable

def first(l: Sequence[T]) -> T:   # Generic function
    return l[0]

arr = [1, 2, 3]
x = first(arr)  # should be int
```

src/Analysis/Engine/Test/AnalysisTest.cs
```diff
@@ -4182,7 +4182,64 @@ class test():
...
+            var code = @"
+from typing import Dict
+
+def func(a: Dict[int, str]):
+    pass
+";
+
+            using (var server = await CreateServerAsync(PythonVersions.LatestAvailable3X)) {
+                var analysis = await server.OpenDefaultDocumentAndGetAnalysisAsync(code);
+                analysis.Scope.Children
```
AlexanderSher: I think `analysis.Should().HaveFunction("func")` should work here

jakebailey:
(Forgive me, I started writing this comment and it blew up to be larger than expected...)

That example I gave does seem to work, but when I actually run it in the editor, I see more than just `int`:

image

I think the fact that "The most base type" is getting added means that `object`'s docs are getting added which I find to be a little awkward.

Also, I can add the following:

```python
arr2 = ["a", "b", "c"]
y = first(arr2)
```

And hovering over `first` will show:

image

Which I find to be a little goofy because:

- `first` isn't returning `int` in this particular location, it's returning `str`, and `y` seems to have been given the correct-ish typeset of `{str, T}`.
- `list` isn't _technically_ `Sequence` (but I suppose we also do this when it's `List`).

I guess this is due to the fact that we're combining sets of types, and `T` is just another on top of what's known to be possible. Maybe a future change can improve this.

Comparing this behavior to what's already in `master`, it seems like this PR doesn't change how `TypeVar` behaves, but is certainly an improvement for `NewType` in any case.

(I'd also hoped that the hover would show something like `first(l: list[int]) -> int` or `first(l: list[str]) -> str` depending on the invocation, but I think that's probably something that hover will have to deal with some other time.)

MikhailArkhipov:
It is by design as all types are currently stored, so yet, it is `int` and `T`. Same as in `def func(a: str)` `a` is `str` and `type` - the latter is `function argument` since it is not type checked and actually does can be anything.

Changing type merging for visual display would be a different task and large amount of work.

jakebailey:
I understand that's the current implementation, and I'm not saying this PR should be fixing anything of that sort, but what about the inconsistency when it comes to how `first` is having its return types displayed? For example, I can swap the order of the invocations and get a different result:

image

src/Analysis/Engine/Impl/Values/IterableInfo.cs
```diff
@@ -255,6 +255,7 @@ public IterableValue(VariableDef[] indexTypes, BuiltinClassInfo seqType, Node no
                 }
 
                 var pi = new ProtocolInfo(DeclaringModule, ProjectState);
+                var actualType = unit.State.GetAnalysisValueFromObjects(ClassInfo.TypeId);
```
MikhailArkhipov: Not used, will remove

src/Analysis/Engine/Impl/Values/Protocols.cs
```diff
@@ -731,4 +731,113 @@ class InstanceProtocol : CallableProtocol {
...
+    /// <summary>
+    /// Generic protocol that delegates work to its inner class.
+    /// Typically used in NewType and TypeVar scenarios delegating 
+    /// calls to the actual class serving as new type.
+    /// </summary>
+    class TypeDelegateProtocol : Protocol {
+        protected readonly AnalysisValue _actualType;
+
+        /// <summary>
+        /// Creates protocol that delegates to base type, such as generic list
```
MikhailArkhipov: Will fix comment since protocol is for NewType and TypeVar

Issue #288, opened 2018-10-19
Add support for Implicit Namespace Packages


AlexanderSher:
- Specification: https://www.python.org/dev/peps/pep-0420/
- General information about packages: https://docs.python.org/3/tutorial/modules.html#packages
- Import on Case-Insensitive Platforms: https://www.python.org/dev/peps/pep-0235/

AlexanderSher:
**First key difference:** namespace packages can be presented in several directories. Consider the following folder structure:
```
├─ package1
│  └─ parent
│     └─ child
│        └─ one.py
├─ package2
│  └─ parent
│     └─ child
│        └─ two.py
└─ root
   └─ app.py
```
If `app.py` contains the following code:
```python
import sys
sys.path += ['path_to_project\\package1\\', 'path_to_project\\package2\\']
import parent.child.one
import parent.child.two

print(parent.child.__path__)
try: 
    print(parent.child.__file__)
except AttributeError:
    print("parent.child have no __file__ attribute")
print(parent.child.one.__file__)
print(parent.child.two.__file__)
```
executing `python root\app.py` will show
```
_NamespacePath(['path_to_project\\package1\\parent\\child', 'path_to_project\\package2\\parent\\child'])
parent.child have no __file__ attribute
path_to_project\package1\parent\child\one.py
path_to_project\package2\parent\child\two.py
```

AlexanderSher:
**Second key difference**: `from package import *` doesn't import anything unless other imports provided information for the namespace. If `app.py` contains the following code:
```python
import sys
sys.path += ['path_to_project\\package1\\', 'path_to_project\\package2\\']
from parent import *

print(child.__path__)
```
executing `python root\app.py` will produce an error:
```
  File "root\app.py", line 8, in <module>
    print(child.__path__)
NameError: name 'child' is not defined
```
However, changing it to 
```python
import sys
sys.path += ['path_to_project\\package1\\', 'path_to_project\\package2\\']
from parent.child.one import *
from parent import *

print(child.__path__)
```
works normally:
```
_NamespacePath(['d:\\Work\\Temp\\PEP0420\\package1\\parent\\child', 'd:\\Work\\Temp\\PEP0420\\package2\\parent\\child'])
```

AlexanderSher:
**Conflict resolution:** if both regular and namespace packages are presented in search paths, namespace packages will be ignored:
```
├─ package1
│  └─ parent
│     └─ child
│        └─ one.py
├─ package2
[...]

PR #327, opened 2018-10-31
[WIP] Support Namespace Packages. Part 1: path resolver

AlexanderSher:
This PR adds a structure that holds references to all python modules in a form of a tree that can be searched for packages and names that are available for each import. This **is not** equivalent to exact set of names, which is affected by previous imports (see, for instance, https://github.com/Microsoft/python-language-server/issues/288#issuecomment-432759994).


AlexanderSher:
Test changes will follow

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.Node.cs
```diff
@@ -0,0 +1,105 @@
...
+            public Node(string name, Node child) {
+                Name = name;
+                _children = new[] {child};
+                ModulePath = null;
+            }
+
+            public Node ToModule(string modulePath) => new Node(Name, _children, modulePath);
+            public Node ToPackage() => new Node(Name, _children, null);
+            public Node AddChild(Node child) => new Node(Name, _children.ImmutableAdd(child), ModulePath);
+            public Node ReplaceChildAt(Node child, int index) => new Node(Name, _children.ImmutableReplaceAt(child, index), ModulePath);
```
MikhailArkhipov: Looks like IList, no?

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -0,0 +1,280 @@
...
+
+        private PathResolverSnapshot(PythonLanguageVersion pythonLanguageVersion, Node[] roots, int version) {
+            _pythonLanguageVersion = pythonLanguageVersion;
+            _roots = roots;
+            Version = version;
+        }
+
+        public IAvailableImports GetAvailableImportsFromAbsolutePath(in string modulePath, in IEnumerable<string> path) {
+            if (!TryFindModule(modulePath, out var lastArc, out _, out _)) {
+                return default;
```
MikhailArkhipov: Does it return null? Perhaps make static empty instance of AvaiableImports returning empty objects?

src/Analysis/Engine/Impl/DependencyResolution/Arc.cs
```diff
@@ -0,0 +1,92 @@
...
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System.Collections.Generic;
+
+namespace Microsoft.PythonTools.Analysis.DependencyResolution {
+    internal partial struct PathResolverSnapshot {
+        /// <summary>
+        /// Represents arc between two nodes in the tree
```
MikhailArkhipov: I think it is called `edge` in https://en.wikipedia.org/wiki/Glossary_of_graph_theory_terms#edge

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -0,0 +1,280 @@
...
+
+            var relativeParentArc = lastArc.GetPrevious(parentCount);
+            if (!TryAppendPackages(relativeParentArc, relativePath, out lastArc)) {
+                return default;
+            }
+            
+            return GetAvailableImports(lastArc);
+        }
+
+        private IAvailableImports GetAvailableImports(in Arc lastArc, in bool requireInitPy = true) {
```
MikhailArkhipov: Do we need `requireInitPy`? The object appears to use specific language version that defines the rule.

src/Analysis/Engine/Impl/Infrastructure/Extensions/ArrayExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+
+namespace Microsoft.PythonTools.Analysis.Infrastructure {
+    internal static class ArrayExtensions {
```
jakebailey: I think `System.Collections.Immutable` has all of this stuff. Would there be any benefit to using that (specifically `ImmutableArray`) instead of our own custom extensions to the builtin Array? I'd think it might help make clear what actually has to be immutable (and what doesn't), and it claims to optimize things to prevent unneeded copies when possible.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.Node.cs
AlexanderSher: No, it's all immutable, so we can't use `IList` interface.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.Node.cs
```diff
@@ -0,0 +1,105 @@
...
+            public Node RemoveChildAt(int index) => new Node(Name, _children.ImmutableRemoveAt(index), ModulePath);
+
+            public Node GetChildAt(int index) => _children[index];
+            public Node GetChild(string childName) {
+                var index = GetChildIndex(childName);
+                return index == -1 ? default : _children[index];
+            }
+
+            public int GetChildIndex(string childName) {
+                for (var index = 0; index < _children.Length; index++) {
```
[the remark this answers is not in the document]
AlexanderSher: Good point, I'll port it.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -0,0 +1,280 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Runtime.InteropServices;
+using Microsoft.PythonTools.Analysis.Infrastructure;
+using Microsoft.PythonTools.Parsing;
+
+namespace Microsoft.PythonTools.Analysis.DependencyResolution {
+    internal partial struct PathResolverSnapshot {
+        private static readonly bool IgnoreCase = RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
```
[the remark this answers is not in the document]
AlexanderSher: Looks like it can be more complicated: https://www.howtogeek.com/354220/how-to-enable-case-sensitive-folders-on-windows-10/
Or we can simply assume the "80%" case.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -0,0 +1,280 @@
...
+        private readonly PythonLanguageVersion _pythonLanguageVersion;
+        private readonly Node[] _roots;
+        public int Version { get; }
+
+        public PathResolverSnapshot(PythonLanguageVersion pythonLanguageVersion) {
```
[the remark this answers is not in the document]
AlexanderSher: Yep

src/Analysis/Engine/Impl/DependencyResolution/IAvailableImports.cs
```diff
@@ -0,0 +1,21 @@
...
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+namespace Microsoft.PythonTools.Analysis.DependencyResolution {
+    internal interface IAvailableImports {
+        string Name { get; }
```
MikhailArkhipov: Only `Name`? What about `Modules` and `Packages` in `AvailablePackageImports`?

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
AlexanderSher: Yep, not required.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
AlexanderSher: Not sure about that, empty instance will require some default values. Need to get to the next step when it will be used by walker.

src/Analysis/Engine/Impl/DependencyResolution/Arc.cs
```diff
@@ -0,0 +1,92 @@
...
+        /// </summary>
+        private readonly struct Arc {
+            // List of vertexes.
+            // To provide immutability, it must be changed only by calling List<T>.Add
+            // Better to be replaced with some immutable structure
+            private readonly List<(int index, Node node)> _vertexes;
```
MikhailArkhipov: `vertices`

src/Analysis/Engine/Impl/DependencyResolution/Arc.cs
AlexanderSher: image
but I can rename it if it is easier to understand.

jakebailey: Also in the same vein of naming changes would be vertexes -> vertices, but I think "arc" is much less common relative to "edge" than "vertexes" is to "vertices".

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -0,0 +1,280 @@
...
+        }
+
+        private bool TryFindModule(string modulePath, out Arc lastArc, out string normalizedPath, out (int start, int length) unmatchedPathSpan) {
+            if (!Path.IsPathRooted(modulePath)) {
+                throw new InvalidOperationException("Module path should be rooted");
+            }
+
+            normalizedPath = PathUtils.NormalizePath(modulePath);
+            var rootIndex = 0;
+            while (rootIndex < _roots.Length && !normalizedPath.StartsWithOrdinal(_roots[rootIndex].Name)) {
```
MikhailArkhipov: OS-specific case sensitivity comparison?

src/Analysis/Engine/Impl/DependencyResolution/IAvailableImports.cs
AlexanderSher: `AvailableModuleImports` doesn't have `Modules` and `Packages`. It only has `ModulePath` which should be used to list variables that are defined in that module (if AST is available).

src/Analysis/Engine/Impl/DependencyResolution/Arc.cs
AlexanderSher: Good point.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -0,0 +1,280 @@
...
+                var newStart = lastArc.Start.ReplaceChildAt(newEnd, lastArc.EndIndex);
+                lastArc = lastArc.Previous;
+                newEnd = newStart;
+            }
+
+            return newEnd;
+        }
+
+        private bool TryFindModule(string modulePath, out Arc lastArc, out string normalizedPath, out (int start, int length) unmatchedPathSpan) {
+            if (!Path.IsPathRooted(modulePath)) {
```
MikhailArkhipov: Is it time to introduce `IFileSystem` for testability along with service container?... So we can run Mac path tests on Windows and the other way around?

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
AlexanderSher: Python is case sensitive in terms of names, need to check how it works with root locations

src/Analysis/Engine/Test/CompletionTests.cs
```diff
@@ -443,19 +443,19 @@ class B(A):
 
         [TestMethod, Priority(0)]
         public async Task TopLevelCompletions() {
-            using (var s = await CreateServerAsync(PythonVersions.LatestAvailable3X)) {
-                var uri = GetDocument(Path.Combine("TestData", "AstAnalysis", "TopLevelCompletions.py"));
-                await s.LoadFileAsync(uri);
+            using (var server = await CreateServerAsync(PythonVersions.LatestAvailable3X)) {
+                var uri = GetDocument(@"TestData\AstAnalysis\TopLevelCompletions.py");
```
jakebailey: I don't think this path should be getting changed back to `\`.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
AlexanderSher: Yes it is. Do you want me to add it to this [WIP]?

src/Analysis/Engine/Test/CompletionTests.cs
AlexanderSher: Sorry, merge conflict.

src/LanguageServer/Impl/LanguageServer.cs
```diff
@@ -44,7 +45,6 @@ public sealed partial class LanguageServer : IDisposable {
 
         private IServiceContainer _services;
         private IUIService _ui;
-        private ITelemetryService2 _telemetry;
```
jakebailey: Is there a reason why you're removing this? The RPC trace listener isn't going to be the only thing that needs to access telemetry (and I don't think it has anything to do with this PR's intent).

src/Analysis/Engine/Impl/Infrastructure/Extensions/ArrayExtensions.cs
```diff
@@ -0,0 +1,52 @@
...
+        public static T[] ImmutableAdd<T>(this T[] array, in T item) {
+            if (array.Length == 0) {
+                return new[] {item};
+            }
+
+            var copy = new T[array.Length + 1];
+            Array.Copy(array, copy, array.Length);
```
MikhailArkhipov: Out of curiosity, is copying cheaper than List?

src/LanguageServer/Impl/LanguageServer.cs
AlexanderSher: Well, I don't like fields that are just hanging around.

src/Analysis/Engine/Impl/DependencyResolution/Arc.cs
MikhailArkhipov: I am fine either way, just that more commonly used term is easier to understand for newcomers.

src/Analysis/Engine/Impl/DependencyResolution/IAvailableImports.cs
MikhailArkhipov: I guess I am confused by plural `Imports` while there is only single `Name`

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
MikhailArkhipov: Sure, prob will need for tests anyway.

src/Analysis/Engine/Impl/Infrastructure/Extensions/ArrayExtensions.cs
AlexanderSher: Yes, `ImmutableArray` is almost free (it is a thin struct wrapper), but other types are rather slow end expensive. Referencing new package for a single type is an overkill, especially for a reference-free assembly. Also `ImmutableArray` stops being cheap when any custom change operation is required.
We can grab the sources of the `ImmutableArray`.

src/Analysis/Engine/Impl/DependencyResolution/Arc.cs
AlexanderSher: Renamed

src/Analysis/Engine/Impl/DependencyResolution/Arc.cs
MikhailArkhipov: Following offline discussion - comment may come handy what is that `index` represents and how it is used since structure is a bit different from typical node with children.

src/Analysis/Engine/Impl/Infrastructure/Extensions/ArrayExtensions.cs
AlexanderSher: In terms of memory? No, `List<T>` is cheaper, but it isn't immutable. Benchmark that created a 5-level tree with 5 children for each node (15626 nodes total) by adding each node one by one produced about 130 kilobytes of garbage.

src/Analysis/Engine/Impl/DependencyResolution/Arc.cs
AlexanderSher: Yep, will add it. And add a `DebuggerDisplay` as well.

src/Analysis/Engine/Impl/Infrastructure/Extensions/ArrayExtensions.cs
jakebailey: I guess I'm a bit ignorant of the implications of adding a reference when we currently have none. I'm personally less satisfied by copying the source code of another library rather than using it the normal way (especially since its code seems to be non-trivial), or even leaving it how it is with some indication that uses shouldn't modify it.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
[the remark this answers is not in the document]
MikhailArkhipov: I guess on Windows it is more like 0.01% since setting appears to be per folder.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
MikhailArkhipov: https://www.python.org/dev/peps/pep-0235/ specifically https://www.python.org/dev/peps/pep-0235/#id4

src/LanguageServer/Impl/LanguageServer.cs
MikhailArkhipov: I think Jake is now using it ;-)

jakebailey: He moved the `GetService` down into the constructor call for the class I added, so it'll still function correctly. My statement was more about how if/when we start sending more telemetry than just for RPC method exceptions, then this change will just be undone anyway.

AlexanderSher: Not exactly. If all the usages of `ITelemetryService2` instance in `LanguageServer` is limited to passing it to another constructors, we won't need a field as well. We will need a field only if `LanguageServer` actually uses the `ITelemetryService2`, but for now it isn't used.

src/Analysis/Engine/Impl/Infrastructure/Extensions/ArrayExtensions.cs
AlexanderSher: The problem may happen if several versions of assembly has to referenced through different code paths. This particular assembly (`System.Collections.Immutable`) has caused problems in RTVS because of the threading package that required special version of assembly that we couldn't use in our extension. 
So being dependent only on .net standard is better for our clients.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
AlexanderSher: PEP 235 doesn't say anything about root locations. Experiment shows that it depends on file system, so with our approximation we can use OS.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
MikhailArkhipov: OK

src/LanguageServer/Impl/Program.cs
```diff
@@ -29,6 +29,7 @@ namespace Microsoft.Python.LanguageServer.Server {
     internal static class Program {
         public static void Main(string[] args) {
             CheckDebugMode();
+            Debugger.Launch();
```
MikhailArkhipov: Prob not needed

AlexanderSher: Sure :)

Issue #296, opened 2018-10-22
Is it possible to use .pyi files alongside analized code?


lostmsu:
I am analyzing a project, that does not have embedded typings, and but there are separate typings for it in .pyi files.

How can I make PythonAnalyzer use those typings?

What I've tried so far:
1. just parse and add typings to analysis queue
2. put typings into a separate directory `Typings` like `Typings\mymodule\mysubmodule.pyi`, and use `PythonAnalysis.SetTypeStubPaths` to `Typings`.

In both cases, when walking `mymodule\mysubmodule.py` file after analysis, and trying to `GetValues(Expression, ...)` I get the same values as if there's no matching typings file.

MikhailArkhipov:
Technically they are used, however, there may be bugs in parsing pyi. 
Related
https://github.com/Microsoft/python-language-server/issues/301
https://github.com/Microsoft/python-language-server/issues/300
https://github.com/Microsoft/python-language-server/issues/254

Try using pyi with simple syntax first, to be sure basic scenario works.

lostmsu:
@MikhailArkhipov but which one is the correct way? And how do I ensure, that it is even loaded?

lostmsu:
Started working on this as a feature here: https://github.com/losttech/PLS/tree/features/AlongsideAnnotations

So far, added parallel annotations AST to `AnalysisUnit`s, and `ProjectEntry`. With this change, it can take annotations file in `AddModule`, and let user supply its Tree to `IPythonParse`. The tree will even be used in multiple places in `FunctionInfo` and `FunctionAnalysisUnit`, however, all evaluation will happen within the context of the main file, which basically means, that all imports in .pyi will be ignored.

I'll try to have analyzer walk over annotations, and let `DDG` have a separate evaluator for them.

MikhailArkhipov:
Correct, they are only used in library analysis since it is too expensive to run full analyzer on such large bodies of code and stub files bring clear benefit.

MikhailArkhipov:
We are doing major rework of the analysis engine and https://github.com/Microsoft/python-language-server/pull/546 supports them.

PR #331, opened 2018-11-01
Allows to use .pyi file annotations supplement to .py files in analysis core

lostmsu:
`PythonAnalyzer.AddModuleAnnotations` allows to add a special kind of module, that is external type annotation. In this change, internally those are tracked as ordinary modules with a special name suffix `__pyi__` (not `.__pyi__`! That is required for relative imports to work).

These modules are analyzed just like any other module.

For every normal module `DDG` creates a reference to corresponding `.pyi`, if one exists, to create a dependency.

Now that both are analyzed, `FunctionAnalysisUnit` needs to find its counterpart from supplemental annotations. For this purpose, `AnalysisUnit` implements `GetExternalAnnotationAnalysisUnit`. For module, it looks for the annotations module in `ProjectState`. For class and function, unit simply descends into the scope by class/function name.

Once `FunctionAnalysisUnit` has it annotations counterpart, it can read analyzed annotations from it, if the local annotations are missing.

This change only applies to the core analysis engine. Language server will need to be updated separately to take advantage of it.

### DIFF
--- a/src/Analysis/Engine/Impl/AnalysisUnit.cs
+++ b/src/Analysis/Engine/Impl/AnalysisUnit.cs
@@ -351,8 +351,9 @@ namespace Microsoft.PythonTools.Analysis {
 
         private AnalysisUnit GetExternalAnalysisUnitImpl() {
             string analysisModuleName = ProjectEntry.ModuleName + PythonAnalyzer.AnnotationsModuleSuffix;
-            if (!ProjectEntry.ProjectState.Modules.TryImport(analysisModuleName, out var analysisModuleReference))
+            if (!ProjectEntry.ProjectState.Modules.TryImport(analysisModuleName, out var analysisModuleReference)) {
                 return null;
+            }
 
             return analysisModuleReference.AnalysisModule?.AnalysisUnit;
         }
@@ -479,15 +480,19 @@ namespace Microsoft.PythonTools.Analysis {
 
         protected internal override AnalysisUnit GetExternalAnnotationAnalysisUnit() {
             var parentAnnotation = _outerUnit.GetExternalAnnotationAnalysisUnit();
-            if (parentAnnotation == null)
+            if (parentAnnotation == null) {
                 return null;
-            if (!parentAnnotation.Scope.TryGetVariable(Ast.Name, out var annotationVariable))
+            }
+
+            if (!parentAnnotation.Scope.TryGetVariable(Ast.Name, out var annotationVariable)) {
                 return null;
-            if (annotationVariable.Types is ClassInfo classAnnotation)
+            }
+            
+            if (annotationVariable.Types is ClassInfo classAnnotation) {
                 return classAnnotation.AnalysisUnit;
-            if (annotationVariable.Types.OnlyOneOrDefault() is ClassInfo nestedAnnotation)
-                return nestedAnnotation.AnalysisUnit;
-            return null;
+            }
+
+            return (annotationVariable.Types.OnlyOneOrDefault() as ClassInfo)?.AnalysisUnit;
         }
     }
 
--- a/src/Analysis/Engine/Impl/AnalysisUnit.cs
+++ b/src/Analysis/Engine/Impl/AnalysisUnit.cs
@@ -59,8 +59,9 @@ namespace Microsoft.PythonTools.Analysis {
         internal AnalysisUnit(Node ast, PythonAst tree, IScope scope, bool forEval) {
             Ast = ast;
             Tree = tree;
-            if (ast == tree)
+            if (ast == tree) {
                 _externalAnnotationAnalysisUnit = new Lazy<AnalysisUnit>(GetExternalAnalysisUnitImpl);
+            }
             _scope = scope;
             ForEval = forEval;
         }
@@ -346,14 +347,9 @@ namespace Microsoft.PythonTools.Analysis {
         ILocationResolver ILocationResolver.GetAlternateResolver() => AlternateResolver;
 
         readonly Lazy<AnalysisUnit> _externalAnnotationAnalysisUnit;
-        protected internal virtual AnalysisUnit GetExternalAnnotationAnalysisUnit() {
-            if (Ast != Tree)
-                return null;
+        protected internal virtual AnalysisUnit GetExternalAnnotationAnalysisUnit() => _externalAnnotationAnalysisUnit?.Value;
 
-            return _externalAnnotationAnalysisUnit.Value;
-        }
-
-        AnalysisUnit GetExternalAnalysisUnitImpl() {
+        private AnalysisUnit GetExternalAnalysisUnitImpl() {
             string analysisModuleName = ProjectEntry.ModuleName + PythonAnalyzer.AnnotationsModuleSuffix;
             if (!ProjectEntry.ProjectState.Modules.TryImport(analysisModuleName, out var analysisModuleReference))
                 return null;
--- a/src/Analysis/Engine/Impl/AnalysisUnit.cs
+++ b/src/Analysis/Engine/Impl/AnalysisUnit.cs
@@ -59,6 +59,8 @@ namespace Microsoft.PythonTools.Analysis {
         internal AnalysisUnit(Node ast, PythonAst tree, IScope scope, bool forEval) {
             Ast = ast;
             Tree = tree;
+            if (ast == tree)
+                _externalAnnotationAnalysisUnit = new Lazy<AnalysisUnit>(GetExternalAnalysisUnitImpl);
             _scope = scope;
             ForEval = forEval;
         }
@@ -343,10 +345,15 @@ namespace Microsoft.PythonTools.Analysis {
 
         ILocationResolver ILocationResolver.GetAlternateResolver() => AlternateResolver;
 
+        readonly Lazy<AnalysisUnit> _externalAnnotationAnalysisUnit;
         protected internal virtual AnalysisUnit GetExternalAnnotationAnalysisUnit() {
             if (Ast != Tree)
                 return null;
 
+            return _externalAnnotationAnalysisUnit.Value;
+        }
+
+        AnalysisUnit GetExternalAnalysisUnitImpl() {
             string analysisModuleName = ProjectEntry.ModuleName + PythonAnalyzer.AnnotationsModuleSuffix;
             if (!ProjectEntry.ProjectState.Modules.TryImport(analysisModuleName, out var analysisModuleReference))
                 return null;
--- a/src/Analysis/Engine/Impl/AnalysisUnit.cs
+++ b/src/Analysis/Engine/Impl/AnalysisUnit.cs
@@ -342,6 +342,17 @@ namespace Microsoft.PythonTools.Analysis {
         internal virtual ILocationResolver AlternateResolver => null;
 
         ILocationResolver ILocationResolver.GetAlternateResolver() => AlternateResolver;
+
+        protected internal virtual AnalysisUnit GetExternalAnnotationAnalysisUnit() {
+            if (Ast != Tree)
+                return null;
+
+            string analysisModuleName = ProjectEntry.ModuleName + PythonAnalyzer.AnnotationsModuleSuffix;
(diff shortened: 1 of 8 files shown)


lostmsu:
Should resolve #296.

lostmsu:
This change only works for scenarios, where the annotations module does not have any internal references.

From the analysis standpoint, `class A` defined in `.py` file, and then annotated in `.pyi` files are two different classes, so if some function `foo` in `A` is annotated to return type `A` in `.pyi`, then analysis will not equate it to returning `A` from `.py`.

src/Analysis/Engine/Impl/AnalysisUnit.cs
```diff
@@ -59,6 +59,8 @@ internal AnalysisUnit(ScopeStatement ast, IScope scope)
         internal AnalysisUnit(Node ast, PythonAst tree, IScope scope, bool forEval) {
             Ast = ast;
             Tree = tree;
+            if (ast == tree)
```
MikhailArkhipov: Braces

src/Analysis/Engine/Impl/AnalysisUnit.cs
```diff
@@ -342,6 +344,22 @@ internal AnalysisUnit(ScopeStatement ast, IScope scope)
         internal virtual ILocationResolver AlternateResolver => null;
 
         ILocationResolver ILocationResolver.GetAlternateResolver() => AlternateResolver;
+
+        readonly Lazy<AnalysisUnit> _externalAnnotationAnalysisUnit;
+        protected internal virtual AnalysisUnit GetExternalAnnotationAnalysisUnit() {
+            if (Ast != Tree)
```
MikhailArkhipov: Since `_externalAnnotationAnalysisUnit` is only created when `ast == tree` it looks like this can be simplified to `=> _externalAnnotationAnalysisUnit != null ? _externalAnnotationAnalysisUnit.Value : null;`

src/Analysis/Engine/Impl/AnalysisUnit.cs
```diff
@@ -342,6 +344,22 @@ internal AnalysisUnit(ScopeStatement ast, IScope scope)
...
+                return null;
+
+            return _externalAnnotationAnalysisUnit.Value;
+        }
+
+        AnalysisUnit GetExternalAnalysisUnitImpl() {
```
MikhailArkhipov: `private`

src/Analysis/Engine/Impl/AnalysisUnit.cs
```diff
@@ -342,6 +344,22 @@ internal AnalysisUnit(ScopeStatement ast, IScope scope)
...
+            string analysisModuleName = ProjectEntry.ModuleName + PythonAnalyzer.AnnotationsModuleSuffix;
+            if (!ProjectEntry.ProjectState.Modules.TryImport(analysisModuleName, out var analysisModuleReference))
+                return null;
+
+            return analysisModuleReference.AnalysisModule?.AnalysisUnit;
```
MikhailArkhipov: Imported non-user files only get through AST analysis, do you actually get analysis unit here?

lostmsu: I am not sure I understand this comment.

PythonAnalyzer.AddModuleAnnotations adds annotations module to its Modules section. Just like for normal modules, you'd have to call .Analyze on the IPythonProjectEntry that is returned by AddModuleAnnotations.

Is there some other kind of analysis, that I am missing?

src/Analysis/Engine/Impl/AnalysisUnit.cs
lostmsu: fixed

src/Analysis/Engine/Impl/AnalysisUnit.cs
lostmsu: accepted this suggestion

src/Analysis/Engine/Impl/AnalysisUnit.cs
lostmsu: added

src/Analysis/Engine/Impl/AnalysisUnit.cs
lostmsu: @jakebailey fixed, also appropriately modified .editorconfig

src/Analysis/Engine/Impl/PythonAnalyzer.cs
```diff
@@ -216,6 +217,34 @@ public static PythonAnalyzer Create(IPythonInterpreterFactory factory, IPythonIn
...
+        /// <param name="filePath">The path to the file on disk</param>
+        /// <param name="cookie">An application-specific identifier for the module</param>
+        /// <returns>The project entry for the new module.</returns>
+        public IPythonProjectEntry AddModuleAnnotations(string moduleName, string filePath, Uri documentUri = null, IAnalysisCookie cookie = null) {
+            if (moduleName == null) {
+                throw new ArgumentNullException(nameof(moduleName));
+            }
+
+            moduleName += AnnotationsModuleSuffix;
+            var entry = new ProjectEntry(this, moduleName, filePath, documentUri, cookie);
```
AlexanderSher: It looks like the body of the method is similar to `AddModule`:
```cs
public IPythonProjectEntry AddModuleAnnotations(string moduleName, string filePath, Uri documentUri = null, IAnalysisCookie cookie = null) {
    Check.ArgumentNotNull(nameof(moduleName), moduleName);
    return AddModule(moduleName + AnnotationsModuleSuffix, filePath, documentUri, cookie);
}
```

src/Analysis/Engine/Impl/Infrastructure/Extensions/EnumerableExtensions.cs
```diff
@@ -60,6 +60,24 @@ public static bool IsNullOrEmpty<T>(this IEnumerable<T> source)
         public static IEnumerable<T> Keys<T, U>(this IEnumerable<KeyValuePair<T, U>> source) => source.Select(GetKey);
         public static IEnumerable<T> ExcludeDefault<T>(this IEnumerable<T> source) => source.Where(i => !Equals(i, default(T)));
 
+        /// <summary>
+        /// Returns the only element of sequence, or the default value of <typeparamref name="T"/>, if sequence has &lt;&gt; 1 elements.
+        /// </summary>
+        public static T OnlyOneOrDefault<T>(this IEnumerable<T> source) {
```
AlexanderSher: There is `Enumerable.SingleOrDefault<T>` method.

src/Analysis/Engine/Impl/ProjectEntry.cs
```diff
@@ -46,7 +46,7 @@ internal sealed class ProjectEntry : IPythonProjectEntry, IAggregateableProjectE
         private readonly HashSet<AggregateProjectEntry> _aggregates = new HashSet<AggregateProjectEntry>();
 
         private TaskCompletionSource<IModuleAnalysis> _analysisTcs = new TaskCompletionSource<IModuleAnalysis>();
-        private AnalysisUnit _unit;
+        internal AnalysisUnit _unit;
```
AlexanderSher: Why do you need it to be `internal`?

src/Analysis/Engine/Impl/Infrastructure/Extensions/EnumerableExtensions.cs
lostmsu: It has a different semantics. SingleOrDefault throws if there's more than 1 element. This one returns default.

src/Analysis/Engine/Impl/ProjectEntry.cs
lostmsu: It's used in ModuleInfo.AnalysisUnit.

src/Analysis/Engine/Impl/PythonAnalyzer.cs
lostmsu: Makes sense

src/Analysis/Engine/Impl/ProjectEntry.cs
AlexanderSher: With some exceptions, fields should be private, they should be wrapped with properties or access methods. But in this case, it is not safe to provide `AnalysisUnit` from `ProjectEntry` for public access. It may be changed while analysis is in progress and that would create unpredictable results.

lostmsu: I could add a `AnalysisUnit` to `ModuleInfo` in a similar way it exists in `ClassInfo`. Then `ProjectEntry` will have control over when to change it. However, if you are talking about potentially multithreaded analysis, they can still be out of sync.

AlexanderSher: Analysis is single threaded, but it is in parallel with parsing. Unfortunately, you can't do it like with `ClassInfo`, because there is only one instance of `ModuleInfo` per `ProjectEntry` - it isn't recreated when new analysis arrives.

lostmsu: Is there a reason for `ClassInfo`s to be recreated and and `ModuleInfo`s not?

lostmsu: It looks like `_unit` reassignment happens just before clearing `ModuleInfo` contents anyway.

AlexanderSher: > Is there a reason for `ClassInfo` to be recreated and `ModuleInfo` not?

`ModuleReference` (https://github.com/Microsoft/python-language-server/blob/2935897bfd8244dbe4296101a80021400ff3c88e/src/Analysis/Engine/Impl/ModuleReference.cs#L37) and some other parts of module dependency analysis work through instance references.

lostmsu: I am open for suggestions. Actually, at this moment I don't understand how can parsing be completed in parallel to analysis without race conditions. Here's the relevant code excerpt from `ProjectEntry.Parse`:

```csharp
_unit = new AnalysisUnit(tree, MyScope.Scope);
AnalysisLog.NewUnit(_unit);

MyScope.Scope.Children = new List<InterpreterScope>();
MyScope.Scope.ClearNodeScopes();
MyScope.Scope.ClearNodeValues();
MyScope.Scope.ClearLinkedVariables();
MyScope.Scope.ClearVariables();
MyScope.ClearReferencedModules();
MyScope.ClearUnresolvedModules();
_unit.State.ClearDiagnostics(this);
```

There does not seem to be anything, that would prevent modification of MyScope.* if there might be an analysis of it going on at the same time.

src/Analysis/Engine/Impl/Infrastructure/Extensions/EnumerableExtensions.cs
lostmsu: The reasoning here is one would use OnlyOneOrDefault when they want to make a decision based on unambiguous analysis result, if there is no expectation of lack of ambiguity.

AlexanderSher:
@lostmsu , can you create a minimal repro project (or provide content of the files) for the issue that you want to fix with this PR? I'm fixing *.pyi handling inside #361, will be useful for me.

lostmsu:
@AlexanderSher 
the scenario this change addresses is: suppose you open a directory in VS Code with nothing but the following two files in it:
`alongside_annotated.py`
```python
class Annotated:
    def foo(self, bar):
        bar
```
and `alongside_annotated.pyi`
```python
class Annotated:
    def foo(self, bar: int): ...
```

The goal is to have VS Code tell you, that `bar` in `alongside_annotated.py` is `int` when you hover over it, and get appropriate completions.

This change only affects analysis part: it enables one to add .pyi files to `PythonAnalyzer` by calling `PythonAnalyzer.AddModuleAnnotations`. These are then used for function parameters and return types.

lostmsu:
@AlexanderSher did that work for you? Does your work cover this use case?

MikhailArkhipov:
We are doing major rework of the analysis engine and https://github.com/Microsoft/python-language-server/pull/546 supports them. DDG analysis is no longer actively supported.

PR #339, opened 2018-11-02
Handle named tuples

### LINKED ISSUE #15: No completions on namedtuple
https://github.com/Microsoft/PTVS/issues/731
https://github.com/Microsoft/vscode-python/issues/2377

```python
from collections import namedtuple
RequestStartLine = namedtuple("RequestStartLine", "method path version".split())
start_line = RequestStartLine("GET", "/query", "HTTP/1.1")
print(start_line)
start_line.  # <- no suggestion
```
This needs new specializations

MikhailArkhipov:
Fixes #15

- Requires #304 so it is included here
- Adds handling of generic `Type[T]` from `typing` to `typeinfo` and the AST analysis
- Adds handling of `namedtuple` from `collections` via typeshed annotations
- Adds fetching of tuple members and handling assignment of tuple to sequence
- Adds specific 'class' flag to the Python type telling that function returns class rather than instance and hence calling it creates instance of the class rather than invoking on instance.
- Filters function overloads preferring annotated definitions to plain ones
- Relevant tests

Implement wrapper over tuple so we can expose declared members in completion - separate item in https://github.com/Microsoft/python-language-server/issues/343


src/Analysis/Engine/Impl/Interpreter/Ast/AstAnalysisFunctionWalkerSet.cs
```diff
@@ -0,0 +1,68 @@
...
+        public void ProcessFunction(FunctionDefinition fn) {
+            if (_functionWalkers.TryGetValue(fn, out var w)) {
+                ProcessWalker(w);
+            }
+        }
+
+        private void ProcessWalker(AstAnalysisFunctionWalker walker) {
+            // Remove walker before processing as to prevent reentrancy.
+            _functionWalkers.Remove(walker.Target);
+            var z = walker.Target.Name == "day";
```
MikhailArkhipov: Debug code, will remove

src/Analysis/Engine/Test/AstAnalysisTests.cs
```diff
@@ -212,6 +191,50 @@ private static Task<Server> CreateServerAsync(InterpreterConfiguration configura
...
+                var analysis = await server.OpenDefaultDocumentAndGetAnalysisAsync(@"
+from ForwardRefFunc1 import *
+x = B().getA().methodA()
+");
+                analysis.Should().HaveVariable("x").OfTypes(BuiltinTypeId.Str);
+            }
+        }
+
+        [TestMethod, Priority(0)]
+        public async Task AstForwardRefFunction2() {
```
AlexanderSher: Can be a parameterized test.

src/Analysis/Engine/Impl/Values/BuiltinFunctionInfo.cs
```diff
@@ -121,26 +122,10 @@ public override IAnalysisSet Call(Node node, AnalysisUnit unit, IAnalysisSet[] a
...
+        public override IEnumerable<OverloadResult> Overloads
+            => GetFunctionOverloads().Select(o => new BuiltinFunctionOverloadResult(ProjectState, Function.Name, o, 0, () => Description));
 
-        public override string Documentation {
-            get {
-                _doc = _doc ?? Utils.StripDocumentation(Function.Documentation);
-                return _doc;
-            }
-        }
+        public override string Documentation => _doc = _doc ?? (_doc = Utils.StripDocumentation(Function.Documentation));
```
AlexanderSher: `public override string Documentation => _doc ?? (_doc = Utils.StripDocumentation(Function.Documentation));` should be enough

src/Analysis/Engine/Impl/Values/BuiltinFunctionInfo.cs
```diff
@@ -25,27 +23,30 @@
...
         public override IPythonType PythonType => _type;
 
-        public override bool IsOfType(IAnalysisSet klass) {
-            return klass.Contains(ProjectState.ClassInfos[BuiltinTypeId.Function]) ||
-                klass.Contains(ProjectState.ClassInfos[BuiltinTypeId.BuiltinFunction]);
+        public override bool IsOfType(IAnalysisSet klass) 
+            => klass.Contains(ProjectState.ClassInfos[BuiltinTypeId.Function]) || klass.Contains(ProjectState.ClassInfos[BuiltinTypeId.BuiltinFunction]);
+
+        public override IAnalysisSet Call(Node node, AnalysisUnit unit, IAnalysisSet[] args, NameExpression[] keywordArgNames) {
+            var returnTypes = GetFunctionOverloads().Where(o => o.ReturnType != null).SelectMany(o => o.ReturnType);
```
AlexanderSher: Do we need `.Distinct()` here?

src/Analysis/Engine/Impl/Values/BuiltinClassInfo.cs
```diff
@@ -158,7 +159,22 @@ public BuiltinClassInfo(IPythonType classObj, PythonAnalyzer projectState)
         }
 
         public override IAnalysisSet GetIndex(Node node, AnalysisUnit unit, IAnalysisSet index) {
-            // TODO: Needs to actually do indexing on type
+            try {
+                var indices = index.Select(x => x.GetConstantValue()).ExcludeDefault().OfType<int>().ToArray();
```
AlexanderSher: `OfType<int>()` will remove defaults, so `ExcludeDefault()` isn't needed.

src/Analysis/Engine/Impl/Values/BuiltinClassInfo.cs
```diff
@@ -158,7 +159,22 @@ public BuiltinClassInfo(IPythonType classObj, PythonAnalyzer projectState)
...
+                    var values= indices.Where(i => i >= 0 && i < a.Length).Select(i => unit.State.GetAnalysisValueFromObjects(a[i]));
+                    return AnalysisSet.UnionAll(values);
+                }
+                if (this is IBuiltinSequenceClassInfo bseqt) {
+                    var a = bseqt.IndexTypes.ToArray();
+                    var values = indices.Where(i => i >= 0 && i < a.Length).Select(i => a[i]);
+                    return AnalysisSet.UnionAll(values);
+                }
+            } catch {
+                return AnalysisSet.Empty;
```
AlexanderSher: What are we catching here?

src/Analysis/Engine/Impl/Values/BuiltinClassInfo.cs
```diff
@@ -158,7 +159,22 @@ public BuiltinClassInfo(IPythonType classObj, PythonAnalyzer projectState)
         }
 
         public override IAnalysisSet GetIndex(Node node, AnalysisUnit unit, IAnalysisSet index) {
-            // TODO: Needs to actually do indexing on type
+            try {
+                var indices = index.Select(x => x.GetConstantValue()).ExcludeDefault().OfType<int>().ToArray();
+                if (this is IPythonSequenceType seqt) {
```
AlexanderSher: Is it possible? What type is derived from `BuiltinClassInfo` and implements `IPythonSequenceType`?

src/Analysis/Engine/Impl/Values/BuiltinClassInfo.cs
```diff
@@ -158,7 +159,22 @@ public BuiltinClassInfo(IPythonType classObj, PythonAnalyzer projectState)
...
+                    var a = seqt.IndexTypes.ToArray();
+                    var values= indices.Where(i => i >= 0 && i < a.Length).Select(i => unit.State.GetAnalysisValueFromObjects(a[i]));
+                    return AnalysisSet.UnionAll(values);
+                }
+                if (this is IBuiltinSequenceClassInfo bseqt) {
+                    var a = bseqt.IndexTypes.ToArray();
```
AlexanderSher: Shouldn't we move this implementation into `GetIndex` override in `SequenceBuiltinClassInfo`?

src/Analysis/Engine/Impl/Interpreter/Ast/AstTypeAnnotationConverter.cs
```diff
@@ -196,6 +192,11 @@ class AstTypeAnnotationConverter : TypeAnnotationConverter<IPythonType> {
             return res;
         }
 
+        private IPythonType MakeGenericClassType(IPythonType typeArg)
+            => typeArg.IsBuiltin
```
AlexanderSher: `UnionType` always return `IsBuiltin => false`. Is that ok?

src/Analysis/Engine/Impl/Interpreter/Ast/AstTypeAnnotationConverter.cs
```diff
@@ -86,21 +86,17 @@ class AstTypeAnnotationConverter : TypeAnnotationConverter<IPythonType> {
...
-        }
+        public override IPythonType MakeUnion(IReadOnlyList<IPythonType> types) => new UnionType(types);
 
-        public override IReadOnlyList<IPythonType> GetUnionTypes(IPythonType unionType) {
-            return (unionType as UnionType)?.Types ??
-                   (unionType as IPythonMultipleMembers)?.Members.OfType<IPythonType>().ToArray();
-        }
+        public override IReadOnlyList<IPythonType> GetUnionTypes(IPythonType unionType) =>
+            (unionType as UnionType)?.Types ??
+            (unionType as IPythonMultipleMembers)?.Members.OfType<IPythonType>().ToArray();
```
AlexanderSher: A bit hard to read. What would you say about changing this to:
```cs
type is UnionType unionType
    ? unionType.Types:
    : type is IPythonMultipleMembers multipleMembers
        ? multipleMembers.Members.OfType<IPythonType>().ToArray()
        : null;
```

src/Analysis/Engine/Impl/Interpreter/Ast/AstPythonType.cs
```diff
@@ -232,8 +236,8 @@ ILocationInfo loc
 
         public IEnumerable<string> GetMemberNames(IModuleContext moduleContext) {
             var names = new HashSet<string>();
-            lock (_members) {
-                names.UnionWith(_members.Keys);
+            lock (Members) {
```
AlexanderSher: I really don't like an idea of taking locks on mutable, non-private field. `AstPythonBuiltinType` needs it only inside `IsHidden`, which can be fixed by adding
```cs
protected bool ContainsMember(string name) {
    lock (_members) {
        return Members.ContainsKey(name);
    }
}
```

src/Analysis/Engine/Impl/Interpreter/Ast/AstPythonProperty.cs
```diff
@@ -5,34 +5,31 @@
...
 
 namespace Microsoft.PythonTools.Interpreter.Ast {
-    class AstPythonProperty : IBuiltinProperty, ILocatedMember {
+    class AstPythonProperty : IBuiltinProperty2, ILocatedMember {
         private IPythonFunctionOverload _getter;
 
         public AstPythonProperty(
             PythonAst ast,
-            FunctionDefinition getter,
+            FunctionDefinition definition,
```
AlexanderSher: 👍

src/Analysis/Engine/Impl/Values/BuiltinClassInfo.cs
MikhailArkhipov: Looks like no one

src/Analysis/Engine/Impl/Interpreter/Ast/AstTypeAnnotationConverter.cs
MikhailArkhipov: I changed implementation, will update soon

src/Analysis/Engine/Impl/Interpreter/Ast/AstPythonType.cs
MikhailArkhipov: OK

src/Analysis/Engine/Impl/Values/BuiltinClassInfo.cs
MikhailArkhipov: Yes

src/Analysis/Engine/Impl/AnalysisValueSetExtensions.cs
```diff
@@ -140,8 +135,7 @@ public static class AnalysisValueSetExtensions {
         /// Performs a delete index operation propagating the index types into
         /// the provided object.
         /// </summary>
-        public static void DeleteIndex(this IAnalysisSet self, Node node, AnalysisUnit unit, IAnalysisSet index) {
-        }
+        public static void DeleteIndex(this IAnalysisSet self, Node node, AnalysisUnit unit, IAnalysisSet index) { }
```
AlexanderSher: Empty static method?

MikhailArkhipov: Apparently it is called from DDG... Let's leave this alone until we decide fate of the DDG.

PR #344, opened 2018-11-05
Fix invalid insertTextFormat in CompletionItem instances

### LINKED ISSUE #342: Completion item insertTextFormat property should be only 1 or 2
Completion items are returning with insertTextFormat properties set to 0. For example:

            {
              "label": "and",
              "kind": 14,
              "detail": null,
              "documentation": null,
              "sortText": "1",
              "filterText": null,
              "preselect": null,
              "insertText": "and",
              "insertTextFormat": 0,
              "textEdit": null,
              "additionalTextEdits": null,
              "commitCharacters": null,
              "command": null,
              "data": null,
              "_kind": "keyword",
              "_values": null
            }

The spec is [here](https://microsoft.github.io/language-server-protocol/specification) (Search for `InsertTextFormat`). According to the spec, `insertTextFormat `be either `1` or `2`.

I believe this is the last item that should get completion working in vim/neovim via https://github.com/autozimu/LanguageClient-neovim.

Thank you for this tool!

jakebailey:
Fixes #342.

All places where `insertText` is set, `insertTextFormat` is now also set. Tests which check `insertText` directly also now check `insertTextFormat`. Most tests use AssertCompletion, which I have modified to also check that every matching completion has a specific format. VS complains about the tuple because I used `Item#` instead of an inferred name to access it, which I'd like to do but cannot because this project seems to only be on C# 7.0, not 7.1.

I think it would be better for the classes that use these sorts of enums to explicitly define the default value (especially given so many LSP enums do not contain 0), but there are quite a few enums and a load of duplication in the two Structures.cs files that I'm not going to try to attempt it now.

### DIFF
--- a/src/LanguageServer/Impl/Implementation/Server.Completion.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.Completion.cs
@@ -146,6 +146,7 @@ namespace Microsoft.Python.LanguageServer.Implementation {
                     filterText = x.filterText,
                     preselect = x.preselect,
                     insertText = x.insertText,
+                    insertTextFormat = (PythonTools.Analysis.LanguageServer.InsertTextFormat)x.insertTextFormat,
                 }).ToArray()
             };
 
@@ -169,6 +170,7 @@ namespace Microsoft.Python.LanguageServer.Implementation {
                 filterText = x.filterText,
                 preselect = x.preselect,
                 insertText = x.insertText,
+                insertTextFormat = (InsertTextFormat)x.insertTextFormat,
                 textEdit = x.textEdit.HasValue
                     ? new TextEdit {
                         range = new Range {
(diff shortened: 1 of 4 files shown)


src/LanguageServer/Impl/Implementation/Server.Completion.cs
```diff
@@ -146,6 +146,7 @@ await InvokeExtensionsAsync((ext, token)
                     filterText = x.filterText,
                     preselect = x.preselect,
                     insertText = x.insertText,
+                    insertTextFormat = (PythonTools.Analysis.LanguageServer.InsertTextFormat)x.insertTextFormat,
```
MikhailArkhipov: using? Definitions that are in the engine are obsolete and are kept around for backward compat. POR to remove them in a month or so.

jakebailey: I'm not sure if any of that is used, but I wanted to be thorough and make sure it's set anywhere it can be. Once all of the old types go away this'll probably also be deleted (and we can probably start setting sane defaults too).

Issue #347, opened 2018-11-05
Functions which return tuples show duplicated return types


jakebailey:
Start with:

![image](https://user-images.githubusercontent.com/5341706/48030697-be168380-e106-11e8-83e6-c4580d89cbd0.png)

Which is correct-ish, ignoring the missing types (which would be better shown as `Any`, but that's a different issue). Now, add a call, which (in the current implementation) propagates types to the function:

![image](https://user-images.githubusercontent.com/5341706/48030769-f322d600-e106-11e8-8eae-447a5605b5aa.png)

The function does in fact return a tuple with three ints, but it's being shown as `tuple[int, int, int][int, int, int]` instead of just `tuple[int, int, int]`.

Note that if I go and annotate the parameters myself with `int` (restricting the possible types from `Any`), this goes away:

![image](https://user-images.githubusercontent.com/5341706/48030922-60cf0200-e107-11e8-9b94-8333f2c26657.png)

jakebailey:
Also:

![image](https://user-images.githubusercontent.com/5341706/48031252-6bd66200-e108-11e8-848c-45f47f085f7b.png)

The first one shows `?` for unknown types, while the second just shows an empty string. Might help indicate where the extra info is coming from.

MikhailArkhipov:
Let me look at it as it is related heavily to #50

PR #350, opened 2018-11-06
Fix max length of short descriptions and tuple names

### LINKED ISSUE #50: RecursiveTuples test fails

MikhailArkhipov:
Fixes #50

- Recursive tuples analysis produced huge description string. Limit that.
- Reduce number of stored types (we don't really need to store various combinations of arguments and return types at all, but that's a different issue).

### DIFF
--- a/src/Analysis/Engine/Impl/AnalysisValue.cs
+++ b/src/Analysis/Engine/Impl/AnalysisValue.cs
@@ -18,6 +18,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Text;
 using Microsoft.PythonTools.Analysis.Analyzer;
 using Microsoft.PythonTools.Analysis.Values;
 using Microsoft.PythonTools.Interpreter;
@@ -30,8 +31,11 @@ namespace Microsoft.PythonTools.Analysis {
     /// analysis values include top-level code, classes, and functions.
     /// </summary>
     public class AnalysisValue : IAnalysisValueOperations {
+        public const int MaxShortDescriptionLength = 128;
+
         [ThreadStatic]
         private static HashSet<AnalysisValue> _processing;
+        private string _shortDescription;
 
         protected AnalysisValue() { }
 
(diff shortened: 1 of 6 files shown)


src/Analysis/Engine/Impl/AnalysisValue.cs
```diff
@@ -67,23 +71,25 @@ public class AnalysisValue : IAnalysisValueOperations {
...
         public virtual string ShortDescription {
             get {
-                var hrd = this as IHasRichDescription;
-                if (hrd != null) {
-                    return string.Join("", hrd.GetRichDescription().TakeWhile(kv => kv.Key != WellKnownRichDescriptionKinds.EndOfDeclaration).Select(kv => kv.Value));
+                if (string.IsNullOrEmpty(_shortDescription) && this is IHasRichDescription hrd) {
+                    var sb = new StringBuilder();
+                    foreach (var item in hrd.GetRichDescription().TakeWhile(kv => kv.Key != WellKnownRichDescriptionKinds.EndOfDeclaration).Select(kv => kv.Value)) {
+                        if (sb.Length >= MaxShortDescriptionLength) {
+                            sb.Append("...");
```
AlexanderSher: Is it better than unicode ellipsis (https://www.fileformat.info/info/unicode/char/2026/index.htm)?

jakebailey: IMO probably, since it's also the same string for Python's ellipsis, and clients will typically render `...` how they see fit depending on the context (though in our case, this will come out to be monospaced).

MikhailArkhipov: Existing code just does plain string - I think because doc text from Python uses plain string as well so appearance of `...` in tooltips etc will be consistent since otherwise some`...` will be different from others in UI appearance.

jakebailey:
Do you have any tests for this new character limiting behavior? (Other than uncommenting the broken one.)

jakebailey:
Though it may not matter since we're going to be changing some of this "protocol" stuff in the future anyway...

MikhailArkhipov:
@jakebailey - I am planning to add the test in the upcoming PR for #347 since that PR will change the description text anyway.

Issue #163, opened 2018-09-27
Optimize 'open files only' case


MikhailArkhipov:
There is no need to analyze the entire workspace. We need to analyze tree that is rooted at the open file.

MikhailArkhipov:
https://github.com/Microsoft/python-language-server/pull/465

PR #361, opened 2018-11-08
Use PathResolverSnapshot to find modules.

AlexanderSher:


src/Analysis/Engine/Impl/DependencyResolution/AvailablePackageImports.cs
```diff
@@ -20,13 +20,15 @@
 namespace Microsoft.PythonTools.Analysis.DependencyResolution {
     internal class AvailablePackageImports : IAvailableImports {
         public string Name { get; }
-        public IReadOnlyDictionary<string, string> Modules { get; }
+        public string RootPath { get; }
+        public AvailableModuleImports[] Modules { get; }
         public string[] Packages { get; }
 
-        public AvailablePackageImports(string name, IReadOnlyDictionary<string, string> modules, string[] packages) {
+        public AvailablePackageImports(string name, string rootPath, AvailableModuleImports[] modules, string[] packages) {
```
MikhailArkhipov: `IAvailableImports[]` or you need a real thing?

src/Analysis/Engine/Impl/DependencyResolution/IAvailableImports.cs
```diff
@@ -17,5 +17,6 @@
 namespace Microsoft.PythonTools.Analysis.DependencyResolution {
     internal interface IAvailableImports {
         string Name { get; }
```
MikhailArkhipov: Comment would be helpful what does `name` means. Interface name implies is plural but data returned seem to imply a single entity

src/Analysis/Engine/Impl/Analyzer/DDG.cs
```diff
@@ -297,160 +302,121 @@ internal class DDG : PythonWalker {
...
-
-            if (!TryImportModule(modName, node.ForceAbsolute, out var modRef, out var bits)) {
-                _unit.DeclaringModule.AddUnresolvedModule(modName, node.ForceAbsolute);
-                return false;
+            var rootNames = node.Root.Names.Select(n => n.Name);
+            var availableImports = node.Root is RelativeModuleName relativeName
+                ? _pathResolver.GetAvailableImportsFromRelativePath(_filePath, relativeName.DotCount, rootNames)
+                : _pathResolver.GetAvailableImportsFromAbsolutePath(_filePath, rootNames);
+
+            switch (availableImports) {
```
MikhailArkhipov: Does it resolve to user files, i.e. `import A` looks for A.py in the same location?

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.Node.cs
```diff
@@ -28,32 +28,39 @@ private class Node {
...
                 ModulePath = null;
+                FullModuleName = null;
             }
 
-            public Node ToModule(string modulePath) => new Node(Name, _children, modulePath);
-            public Node ToPackage() => new Node(Name, _children, null);
-            public Node AddChild(Node child) => new Node(Name, _children.ImmutableAdd(child), ModulePath);
-            public Node ReplaceChildAt(Node child, int index) => new Node(Name, _children.ImmutableReplaceAt(child, index), ModulePath);
-            public Node RemoveChildAt(int index) => new Node(Name, _children.ImmutableRemoveAt(index), ModulePath);
+            public Node ToModule(string modulePath, string fullModuleName) => new Node(Name, _children, modulePath, fullModuleName);
```
MikhailArkhipov: Sounds more like `FromModule` since it takes module data and returns `Node`

src/Analysis/Engine/Impl/DependencyResolution/AvailablePackageImports.cs
AlexanderSher: This collection contains only module imports. I can add an interface for `AvailableModuleImports`.

src/Analysis/Engine/Impl/DependencyResolution/IAvailableImports.cs
AlexanderSher: Yeah, I actually think I need better name for the interface as well.

src/Analysis/Engine/Impl/Analyzer/DDG.cs
AlexanderSher: This logic applies only to python 2. In python 3, `import a` always means that `a` is in the root folder.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.Node.cs
AlexanderSher: I can call it `ToModuleNode`

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -14,31 +14,59 @@
...
 using System.Runtime.InteropServices;
-using Microsoft.PythonTools.Analysis.Infrastructure;
-using Microsoft.PythonTools.Parsing;
+using System.Text;
 
 namespace Microsoft.PythonTools.Analysis.DependencyResolution {
     internal partial struct PathResolverSnapshot {
-        private static readonly Node NullRoot = Node.CreateRoot("$");
+        private static readonly Node NullRoot = Node.CreateNullRoot();
         private static readonly bool IgnoreCaseInPaths = RuntimeInformation.IsOSPlatform(OSPlatform.Windows) || RuntimeInformation.IsOSPlatform(OSPlatform.OSX);
```
MikhailArkhipov: `IgnoreCaseInPaths ` - might be worth making something common in `Infrastructure`, prob will be useful in other places.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -91,48 +112,146 @@ public PathResolverSnapshot(PythonLanguageVersion pythonLanguageVersion)
...
+
+            var newRoots = CreateRoots(normalizedRootDirectory, _userSearchPaths, _interpreterSearchPaths);
+            return new PathResolverSnapshot(_pythonLanguageVersion, normalizedRootDirectory, _userSearchPaths, _interpreterSearchPaths, newRoots, Version + 1);
         }
 
-        public PathResolverSnapshot SetSearchPaths(in IEnumerable<string> searchPaths) {
-            var rootNodes = searchPaths.Where(Path.IsPathRooted).Select(Node.CreateRoot).Prepend(_roots[0]).ToArray();
-            return new PathResolverSnapshot(_pythonLanguageVersion, rootNodes, Version + 1);
+        public PathResolverSnapshot SetUserSearchPaths(in IEnumerable<string> searchPaths) {
+            var userSearchPaths = searchPaths.ToArray();
```
MikhailArkhipov: User search paths can contain anything, like `.` for example

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -228,19 +347,23 @@ public PathResolverSnapshot(PythonLanguageVersion pythonLanguageVersion)
...
 
             if (unmatchedPathSpan.start == GetModuleNameStart(modulePath)) {
                 // Module is added to existing package
-                return Node.CreateModule(modulePath.Substring(unmatchedPathSpan.start, unmatchedPathSpan.length), modulePath);
+                var name = modulePath.Substring(unmatchedPathSpan.start, unmatchedPathSpan.length);
+                fullModuleName = GetFullModuleName(lastEdge, name);
+                return Node.CreateModule(name, modulePath, fullModuleName);
             }
 
             var names = modulePath.Split(Path.DirectorySeparatorChar, unmatchedPathSpan.start, unmatchedPathSpan.length);
```
MikhailArkhipov: Does it handle UNC ie `\\machine\share\...`?

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -249,6 +372,75 @@ public PathResolverSnapshot(PythonLanguageVersion pythonLanguageVersion)
...
+            return sb.ToString();
+        }
+
+        private static string GetFullModuleName(in Edge lastEdge, string[] names) {
+            var sb = GetFullModuleNameBuilder(lastEdge);
+            if (!lastEdge.IsFirst) {
+                AppendName(sb, lastEdge.End.Name);
+                sb.Append('.').Append(".", names, 0, names.Length - 1);
+            } else {
+                sb.Append(".", names, 0, names.Length - 1);
```
MikhailArkhipov: Seems to be identical to above ie
```cs
            if (!lastEdge.IsFirst) {
                AppendName(sb, lastEdge.End.Name);
                sb.Append('.');
            } 
            sb.Append(".", names, 0, names.Length - 1);
```

src/Analysis/Engine/Impl/Infrastructure/PathUtils.cs
```diff
@@ -18,15 +18,41 @@
...
+
+        internal static readonly char[] DirectorySeparators = {
             Path.DirectorySeparatorChar,
             Path.AltDirectorySeparatorChar
         };
 
+        public static bool IsValidWindowsDriveChar(char value) => (value >= 'A' && value <= 'Z') || (value >= 'a' && value <= 'z');
+
+        public static bool IsValidFileNameCharacter(char character) {
+            if (character < 32) {
```
MikhailArkhipov: `Path.InvalidPathChars` ?

src/Analysis/Engine/Impl/Infrastructure/PathUtils.cs
```diff
@@ -382,6 +408,46 @@ static class PathUtils {
...
+            } else if (IsWindows && path.Length >= 2 && IsValidWindowsDriveChar(path[0]) && path[1] == ':') {
+                offset = 2;
+            } else {
+                return false;
+            }
+
+            var impossibleName = false;
+            for (var i = offset; i < path.Length; i++) {
+                var character = path[i];
+                if (character == Path.AltDirectorySeparatorChar) {
```
MikhailArkhipov: `Path.AltDirectorySeparatorChar == Path.DirectorySeparatorChar` on Unix

src/Analysis/Engine/Impl/Infrastructure/PathUtils.cs
```diff
@@ -382,6 +408,46 @@ static class PathUtils {
             return null;
         }
 
+        public static bool IsNormalizedPath(string path) {
```
MikhailArkhipov: I wonder if we can just `PathUtil.NormalizePath(path) == path`?

src/Analysis/Engine/Impl/Infrastructure/PathUtils.cs
MikhailArkhipov: There is some fun in `PathEqualityComparerTests`

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
AlexanderSher: PathResolver never accesses file system, so there can't be any IOExceptions. Search paths are preserved in their form cause when `_rootDirectory` is changed, we can try to convert relative path to absolute.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
AlexanderSher: No. Do we have to support it?

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
AlexanderSher: Fixed

src/Analysis/Engine/Impl/Infrastructure/PathUtils.cs
AlexanderSher: Ok

src/Analysis/Engine/Impl/Infrastructure/PathUtils.cs
AlexanderSher: > `Path.AltDirectorySeparatorChar == Path.DirectorySeparatorChar` on Unix

Didn't know. Thanks!

src/Analysis/Engine/Impl/Infrastructure/PathUtils.cs
AlexanderSher: `PathUtil.NormalizePath(path)` produces enormous amount of garbage. During initial load, some of that garbage even ends up in Gen2. So the goal was the opposite - don't create new path if the original one is normalized already.

src/Analysis/Engine/Impl/DependencyResolution/IAvailableImports.cs
AlexanderSher: Renamed

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
AlexanderSher: Doesn't work right now because it always uses `Path.DirectorySeparatorChar`, but it is easy to fix.

AlexanderSher:
Unfortunately, `AstAnalysisWalker` doesn't have a lot to share with `DDG`, cause they work differently:
- `DDG` immediately tries to load module while walking through `ImportStatement` and `FromImportStatement` and assign all imported values to the variables. If module can't be loaded immediately because of circular dependency, it will use a sentinel, but because of that dependency module will be reanalyzed, and sentinel will be (most likely) replaced with the real module.
- `AstAnalysisWalker` doesn't load anything (with exception for `from module import *` case). It creates `AstNestedPythonModule` for imported modules and `AstNestedPythonModuleMember` instance for imported members, which later resolve dependencies through lazy loading when variable is accessed. If module isn't imported at the moment of resolution, `AstNestedPythonModule` preserves sentinel module, and it will never be replaced with a real one. It is probably done intentionally to avoid complexity.

So the only thing I can share is `PathResolver` itself. I'll submit changes to `AstAnalysisWalker` in the next commit to this [WIP].

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -14,132 +14,344 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
+using Microsoft.PythonTools.Analysis.Infrastructure;
+using Microsoft.PythonTools.Parsing;
```
MikhailArkhipov: Sorting

src/LanguageServer/Impl/Program.cs
```diff
@@ -29,6 +29,7 @@ namespace Microsoft.Python.LanguageServer.Server {
     internal static class Program {
         public static void Main(string[] args) {
             CheckDebugMode();
+            Debugger.Launch();
```
MikhailArkhipov: Is it needed?

MikhailArkhipov:
This needs tons of tests :-)

src/Analysis/Engine/Impl/Analyzer/DDG.cs
MikhailArkhipov: I meant, in general, for #163 which currently does not work since legacy import resolution works on libraries while user files have to be loaded when the workspace opens.

src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
MikhailArkhipov: I think `Path.Combine(rootDirectory, p)` etc also throw when path parts are invalid, but that is `ArgumentException`

MikhailArkhipov:
I assume we merge it post Nov release?

AlexanderSher:
Ok, I've removed the WIP prefix. There are some known bugs, but they can be fixed separately.

src/Analysis/Engine/Impl/Analyzer/DDG.cs
AlexanderSher: It doesn't rely on the file being open, yes. File will be analyzed by `AstAnalysisWalker`. There may be problems though when file is open.

jakebailey:
Given that this dependency resolver is independent from the analyzer and really only makes use of the AST, would it make more sense to move this into one of the new assemblies Mikhail is planning on creating in a future PR?

The only `using` from `Analysis` is `Microsoft.PythonTools.Analysis.Infrastructure`, which I would expect to be moved/reworked to a more common location anyway since it contains a lot of useful extensions.

MikhailArkhipov:
@jakebailey - I will move the resolver to the proper assembly with my PR so new code can access it. Currently there is no separate assembly yet.

PR #372, opened 2018-11-08
Remove call to path resolver until it's finished

jakebailey:
It's causing some InvalidOperationExceptions to be thrown. Disable this for now, since we're not yet using this system.

### DIFF
--- a/src/Analysis/Engine/Impl/PythonAnalyzer.cs
+++ b/src/Analysis/Engine/Impl/PythonAnalyzer.cs
@@ -206,7 +206,7 @@ namespace Microsoft.PythonTools.Analysis {
         /// <returns>The project entry for the new module.</returns>
         public IPythonProjectEntry AddModule(string moduleName, string filePath, Uri documentUri = null, IAnalysisCookie cookie = null) {
             if (documentUri == null || documentUri.Scheme != "python") {
-                _pathResolver.AddModulePath(filePath);
+                // _pathResolver.AddModulePath(filePath);
             }
 
             var entry = new ProjectEntry(this, moduleName, filePath, documentUri, cookie);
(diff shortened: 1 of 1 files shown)


src/Analysis/Engine/Impl/PythonAnalyzer.cs
```diff
@@ -206,7 +206,7 @@ public static PythonAnalyzer Create(IPythonInterpreterFactory factory, IPythonIn
         /// <returns>The project entry for the new module.</returns>
         public IPythonProjectEntry AddModule(string moduleName, string filePath, Uri documentUri = null, IAnalysisCookie cookie = null) {
             if (documentUri == null || documentUri.Scheme != "python") {
-                _pathResolver.AddModulePath(filePath);
+                // _pathResolver.AddModulePath(filePath);
```
MikhailArkhipov: Do we need to comment out `SetRoot` as well?

jakebailey: SetRoot above, you mean? My VS claims it has no references.

jakebailey: The other option is that I can just comment out every reference to `_pathResolver` (declaration of the class field, construction, etc), but I thought that'd make merging for Alex more annoying and I didn't want to do that.

MikhailArkhipov: `internal void SetRoot(string rootDir) => _pathResolver.SetRoot(rootDir);` called from

```
Analyzer.SetRoot(_rootDir);
```

in `DoInitializeAsync`

jakebailey: Fixed.

PR #389, opened 2018-11-13
Include assembly version number in telemetry event properties, standardize TelemetryEvent creation

### LINKED ISSUE #373: Send version string in telemetry event
The version number would be very helpful in looking at and filtering events. The extension forwards our telemetry as-is, so we should add the version to the telemetry events before sending.

Right now, only the `rpc.exception` event is being sent, so we can add an extra property with the version string. In the future when we send more than just the one, we'll probably have some standardized way of producing the event (pulling out the eventName prefix and so on too).

jakebailey:
Fixes #373.

The "default" version for the assembly is 1.0.0.0 (which shows up when running a locally built copy), so I've added the suffix `-debug` when `DEBUG` is set in order to distinguish events if we reach a 1.0.0.0 release.

### DIFF
--- a/src/LanguageServer/Impl/Telemetry.cs
+++ b/src/LanguageServer/Impl/Telemetry.cs
@@ -23,26 +23,10 @@ using StreamJsonRpc;
 namespace Microsoft.Python.LanguageServer.Implementation {
     internal class Telemetry {
         private const string EventPrefix = "python_language_server/";
-        private static readonly string PLSVersion = GetPLSVersion();
 
-        public static TelemetryEvent CreateEvent(string eventName) {
-            var e = new TelemetryEvent {
-                EventName = EventPrefix + eventName,
-            };
-            e.Properties["plsVersion"] = PLSVersion;
-
-            return e;
-        }
-
-        private static string GetPLSVersion() {
-            var version = Assembly.GetExecutingAssembly().GetName().Version.ToString();
-
-#if DEBUG
-            version += "-debug";  // Add a suffix so we can more easily ignore non-release versions.
-#endif
-
-            return version;
-        }
+        public static TelemetryEvent CreateEvent(string eventName) => new TelemetryEvent {
+            EventName = EventPrefix + eventName,
+        };
     }
 
     internal class TelemetryRpcTraceListener : TraceListener {
--- a/src/LanguageServer/Impl/Telemetry.cs
+++ b/src/LanguageServer/Impl/Telemetry.cs
@@ -21,6 +21,30 @@ using Microsoft.PythonTools.Analysis.Infrastructure;
 using StreamJsonRpc;
 
 namespace Microsoft.Python.LanguageServer.Implementation {
+    internal class Telemetry {
+        private const string EventPrefix = "python_language_server/";
+        private static readonly string PLSVersion = GetPLSVersion();
+
+        public static TelemetryEvent CreateEvent(string eventName) {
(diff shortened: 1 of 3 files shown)


jakebailey:
Marking this as WIP (do not merge). I may take this chance to pull out the creation of a common TelemetryEvent out into Infrastructure, since we're going to want to have the EventName prefix for everything and the version string too.

jakebailey:
Updated to move things into a telemetry file, and add a standardized helper to prepopulate an event to be added to.

Technically, I could have made the helper also accept a dictionary for properties/measurements, but the usage would have been more annoying because of how C# dictionary literals look.

src/LanguageServer/Impl/Telemetry.cs
```diff
@@ -21,6 +21,30 @@
...
 namespace Microsoft.Python.LanguageServer.Implementation {
+    internal class Telemetry {
+        private const string EventPrefix = "python_language_server/";
+        private static readonly string PLSVersion = GetPLSVersion();
+
+        public static TelemetryEvent CreateEvent(string eventName) {
+            var e = new TelemetryEvent {
+                EventName = EventPrefix + eventName,
+            };
+            e.Properties["plsVersion"] = PLSVersion;
```
AlexanderSher: Shouldn't this be in `TelemetryService.SendTelemetry`? I.e.:
```cs
public Task SendTelemetry(TelemetryEvent telemetryEvent) {
    telemetryEvent.Properties["plsVersion"] = PLSVersion;
    _rpc.NotifyWithParameterObjectAsync("telemetry/event", telemetryEvent);
}
```

jakebailey: I'm not sure, since it's also being used by Intellicode to send their own telemetry and I'm not sure I want to add properties to their events.

jakebailey: Though, including the language server version wholesale would probably not be a bad idea.

AlexanderSher: Intellicode depends on LS, so having information about LS version is also important.

jakebailey: Alright, I guess I'll move this out into the service. Should I just overwrite the dictionary's `plsVersion` value (as your suggested code does) or check before setting? (I can't imagine they'd ever try to set it differently, but...)

I'll probably also have to null check too, since currently a null event is allowed to go through, as useless as that is. Maybe it'd be better to just have the service ignore any null events outright.

AlexanderSher: I think we shouldn't allow overriding of our own event values by plugins, this information is important. As for null ref, I'd prefer to fail. As you said, null events are meaningless, and most likely it is a bug.

jakebailey:
One side effect of that change I just thought of: this code will modify the event it's given, which may be unexpected behavior. Should I do a copy of the dictionary and add the extra key into that rather than modifying the event given?

AlexanderSher:
Up to you. `TelemetryEvent` has public read/write dictionaries, meaning anyone can change it by design.

jakebailey:
Unfortunately the way the TelemetryEvent class was written differs from the other data classes we send over RPC. I can't just make an event with the specified measurements and properties dict, I have to copy from one to the other by hand as it creates its own dicts at construction time. I'll just leave this as-is for now, and if it's a problem we can come back to it later. These events are going to be fire-and-forget anyway.

PR #393, opened 2018-11-13
Improve line formatter behavior with unmatched close brackets, log warning

### LINKED ISSUE #388: Ignore "close bracket has no matching open" exception in telemetry
This exception gets thrown when the line formatter gets run on code with unbalanced braces/parens/brackets. Since this is a user code error, we should not be recording the event, as it's just noise.

My initial implementation just threw `new Exception(...)`, and should probably be switched to a more specific subclass.

jakebailey:
Fixes #388.

Instead of throwing a generic exception, keep track of the unmatched token for later use, and just ignore it. When the line formatting is done, DocumentOnTypeFormatting can check to see if an unmatched token occurred before or in the line being formatted and report to the user a warning about the bad behavior.

I also fixed the comment strings that claimed that the line formatter worked with one-indexed lines, which is no longer true after Alex's changes to this code.

### DIFF
--- a/src/LanguageServer/Impl/Implementation/Server.OnTypeFormatting.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.OnTypeFormatting.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Python.LanguageServer.Implementation {
 
                 if (unmatchedToken != null) {
                     var message = Resources.LineFormatter_UnmatchedToken.FormatInvariant(unmatchedToken.Value.token, unmatchedToken.Value.line + 1);
-                    ShowMessage(MessageType.Warning, message);
+                    LogMessage(MessageType.Warning, message);
                 }
 
                 return edits;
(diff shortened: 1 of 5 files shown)


src/LanguageServer/Impl/Implementation/Server.OnTypeFormatting.cs
```diff
@@ -49,7 +50,15 @@ public sealed partial class Server {
                 }
 
                 var lineFormatter = new LineFormatter(reader, Analyzer.LanguageVersion);
-                return lineFormatter.FormatLine(targetLine);
+                var edits = lineFormatter.FormatLine(targetLine);
+                var unmatchedToken = lineFormatter.UnmatchedToken(targetLine);
+
+                if (unmatchedToken != null) {
+                    var message = Resources.LineFormatter_UnmatchedToken.FormatInvariant(unmatchedToken.Value.token, unmatchedToken.Value.line + 1);
+                    ShowMessage(MessageType.Warning, message);
```
MikhailArkhipov: This will pop up message box (modal in VS) that has to be dismissed. UI designers consider popups intrusive and the practice is discouraged. Non-intrusive console output (via logger) is  typically sufficient.

jakebailey: I switched it out for `LogMessage`.

PR #429, opened 2018-11-21
Special case known good names for use-before-def

### LINKED ISSUE #391: True/False being reported as undefined
Microsoft/vscode-python#3274

This has been reported by at least 3 separate people (see the duplicate issue to the one above).

jakebailey:
Fixes #391.

Sort of a hack, but these hardcoded names are pulled from the list of atoms in the Python grammar. I've still been unable to reproduce True/False being use-before-def (other than once without a debugger attached), but it shouldn't occur given that I now explicitly ignore them.

Unfortunately, this may mask a bug in `ProjectState.BuiltinModule.GetMember`, since I don't think there's a good reason why `None` should fail to be found...

### DIFF
--- a/src/Analysis/Engine/Impl/Analyzer/ExpressionEvaluator.cs
+++ b/src/Analysis/Engine/Impl/Analyzer/ExpressionEvaluator.cs
@@ -150,7 +150,18 @@ namespace Microsoft.PythonTools.Analysis.Analyzer {
                         res = refs.Types;
                     } else {
                         // ... warn the user
-                        warn = !(node is ConstantExpression); // Don't warn when None.
+                        // "atom" in Python grammar.
+                        switch (name) {
+                            case "None":
+                            case "False":
+                            case "True":
+                            case "...":
+                            case string _ when node is ConstantExpression:
+                                break;
+                            default:
+                                warn = true;
+                                break;
+                        }
                     }
                 }
             }
--- a/src/Analysis/Engine/Test/LanguageServerTests.cs
+++ b/src/Analysis/Engine/Test/LanguageServerTests.cs
@@ -363,17 +363,14 @@ def f(a = 2, b): pass
             }
         }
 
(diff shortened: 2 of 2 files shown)


MikhailArkhipov:
This can happen when builtins module is not loaded/scraped, which would be a big problem in many other cases. Members are assigned in the `AstBuiltinsPythonModule`
```cs
            if (boolType != null) {
                _members["True"] = _members["False"] = new AstPythonConstant(boolType);
            }
```
This can be a threading issue such as code analyzed before builtins are loaded

jakebailey:
Well, if you think there's a better way to approach this, I'd be curious to see what you think about the test case for `None` presented in the most recent comment in #391.

My previous fix to ignore `ConstantExpression` doesn't really work because the node being inspected is some sort of index expression. I think that fix was actually also masking this same problem, because `None` should have shown up in the builtins.

MikhailArkhipov:
Might be falling to fallback in
```cs
            var moduleRef = await Modules.TryImportAsync(_builtinName, token).ConfigureAwait(false);
            if (moduleRef != null) {
                _builtinModule = (BuiltinModule)moduleRef.Module;
            } else {
                _builtinModule = new BuiltinModule(fallback, this);
                Modules.SetModule(_builtinName, BuiltinModule);
            }
```

MikhailArkhipov:
On the `None` - try adding it to members as we do with `True` and `False`

```
                    if (typeId == BuiltinTypeId.NoneType) {
                        noneType = m as IPythonType;
                    }
...
_members["None"] = new AstPythonConstant(noneType);
```

jakebailey:
Doing that does actually fix that Generator test (and is way less hacky), but I'm not exactly sure where how to approach finding how the analysis for builtins doesn't happen before the rest of the diagnostics happens.

jakebailey:
Closing in favor of #447 for the `None` part. The True/False issue remains unsolved and will likely need to be approached in another way.

PR #439, opened 2018-11-29
Clean up type system

### LINKED ISSUE #404: Clean up type system
Clean interfaces: 
- IPythonType, IPythonClass, IPythonModule,.... 
- Remove obsolete methods.
- Derive implementation appropriately (i.e. function is a type). 
- Drop .NET/IronPython specific code.
- Make variables store `IPythonType` rather than `IMember`. 
- Implement MemberId as function of the BuiltInTypeId so they appear consistent.
- Implement IsBuiltIn as a function of declaring module.

MikhailArkhipov:
Fixes #404

- Make type derive properly, i.e. `IPythonType` => `IPythonFunction`, make module a type, etc.
- Move class-specific methods to `IPythonClass`
- Simplify implementation by using proper base classes like `AstPythonTypeWrapper`.
- Reduce type complexity with functions/methods/bounded such as if function has declaring type, it is then a method, etc.
- Fix couple tests with typings related to protocol derivation.
- In most cases `MemberId` is now function of `BuiltInTypeId` for consistency.
- Update few tests baselines with changed tooltips/information.

I will wait for https://github.com/Microsoft/python-language-server/pull/361 to go in first.


src/Analysis/Engine/Test/FluentAssertions/MemberContainerAssertionsExtensions.cs
```diff
@@ -27,9 +27,9 @@ internal static class MemberContainerAssertionsExtensions {
 
             Execute.Assertion.ForCondition(constraint.Which.MemberType == memberType)
                 .BecauseOf(because, reasonArgs)
-                .FailWith($"Expected {AssertionsUtilities.GetQuotedName(constraint.Which)} to have type '{memberType}', but found '{memberType}'");
+                .FailWith($"Expected {AssertionsUtilities.GetQuotedName(constraint.Which)} to have type '{memberType}', but found '{constraint.Which.MemberType}'");
```
AlexanderSher: 👍

src/Analysis/Engine/Test/AstAnalysisTests.cs
```diff
@@ -942,19 +945,13 @@ class BankAccount(object):
                 var analysis = await server.OpenDefaultDocumentAndGetAnalysisAsync(code);
 
                 var v0 = analysis.Should().HaveVariable("scanner").WithValueAt<IBuiltinInstanceInfo>(0);
```
AlexanderSher: you probably don't need this variable anymore.

src/Analysis/Engine/Impl/Values/Protocols.cs
```diff
@@ -71,7 +71,7 @@ abstract class Protocol : AnalysisValue, IHasRichDescription {
         public override IAnalysisSet GetDescriptor(PythonAnalyzer projectState, AnalysisValue instance, AnalysisValue context) => AnalysisSet.Empty;
         public override IAnalysisSet GetEnumeratorTypes(Node node, AnalysisUnit unit) => AnalysisSet.Empty;
         public override IAnalysisSet GetIndex(Node node, AnalysisUnit unit, IAnalysisSet index) => AnalysisSet.Empty;
-        public override IAnalysisSet GetInstanceType() => AnalysisSet.Empty;
+        public override IAnalysisSet GetInstanceType() => null;
```
AlexanderSher: Looks dangerous. I think we have a lot of places that expect `IAnalysisSet` to be not null.

src/Analysis/Engine/Impl/Values/BuiltinInstanceInfo.cs
```diff
@@ -26,15 +26,15 @@ internal class BuiltinInstanceInfo : BuiltinNamespace<IPythonType>, IBuiltinInst
         private readonly BuiltinClassInfo _klass;
 
         public BuiltinInstanceInfo(BuiltinClassInfo klass)
-            : base(klass?._type, klass?.ProjectState) {
+            : base(klass?.Type, klass?.ProjectState) {
```
AlexanderSher: Do you mind if we rename `klass` into `classInfo`?

src/Analysis/Engine/Impl/Values/BuiltinFunctionInfo.cs
```diff
@@ -93,7 +93,7 @@ public override bool IsOfType(IAnalysisSet klass)
                 return name;
             }
             name = type.Name + "." + name;
-            if (type.IsBuiltin || type.DeclaringModule == null) {
+            if (type.IsBuiltIn || type.DeclaringModule == null) {
```
AlexanderSher: Kinda inconsistent. It is still `Builtin` in other places.

src/Analysis/Engine/Impl/Values/BuiltinClassInfo.cs
```diff
@@ -127,19 +125,12 @@ public BuiltinClassInfo(IPythonType classObj, PythonAnalyzer projectState)
...
             get {
-                // TODO: sometimes might have a specialized __init__.
-                // This just covers typical .NET types
-                var ctors = _type.GetConstructors();
-
-                if (ctors != null) {
-                    return ctors.Overloads.Select(ctor => new BuiltinFunctionOverloadResult(ProjectState, _type.Name, ctor, 1, () => Documentation));
-                }
-                return new OverloadResult[0];
+                var ctors = Type.GetConstructor();
```
AlexanderSher: Let's leave the comment. We may need it during switching PTVS to LS

src/Analysis/Engine/Impl/PythonAnalyzer.cs
```diff
@@ -751,43 +751,50 @@ private static IMemberResult[] MemberDictToMemberResult(Dictionary<string, List<
...
-                    if (md.IsBound) {
-                        return new BuiltinFunctionInfo(md.Function, this);
-                    } else {
-                        return new BuiltinMethodInfo(md, this);
-                    }
-                }) ?? _noneInst;
-            } else if (attr is IPythonBoundFunction pbf) {
-                return GetCached(attr, () => new BoundBuiltinMethodInfo(pbf, this)) ?? _noneInst;
-            } else if (attr is IBuiltinProperty bp) {
+            if (attr is IPythonFunction pf) {
```
AlexanderSher: Looks like a good candidate for `switch` statement, but up to you.

src/Analysis/Engine/Impl/OverloadResult.cs
```diff
@@ -333,43 +335,57 @@ internal override OverloadResult WithoutLeadingParameters(int skipCount = 1)
...
 
-                        // always add kw dict last.  When defined in C# and combined w/ params 
-                        // it has to come earlier than it's legally allowed in Python so we 
-                        // move it to the end for intellisense purposes here.
-                        if (kwDict != null) {
-                            result.Add(kwDict);
-                        }
-                        _parameters = result.ToArray();
+            if (_overload != null) {
+                var target = _overload;
```
AlexanderSher: Looks like not used

src/Analysis/Engine/Impl/Values/Protocols.cs
MikhailArkhipov: This is to restore null since code that determines instance type of the protocol depends on it. We have two tests failing in master b/c of this.

src/Analysis/Engine/Impl/PythonAnalyzer.cs
MikhailArkhipov: All this will go away when DDG is replaced by AST analysis.

src/Analysis/Engine/Impl/KnownTypes.cs
```diff
@@ -147,27 +135,17 @@ class FallbackBuiltinModule : IBuiltinPythonModule, IPythonModule {
...
-        public IEnumerable<string> GetMemberNames(IModuleContext moduleContext) => Enumerable.Empty<string>();
         public void Imported(IModuleContext context) { }
     }
 
-    class FallbackBuiltinPythonType : IPythonType {
-        public FallbackBuiltinPythonType(IBuiltinPythonModule module, BuiltinTypeId typeId, string name = null) {
-            DeclaringModule = module;
-            Name = name ?? typeId.GetModuleName((DeclaringModule as FallbackBuiltinModule)?.LanguageVersion ?? PythonLanguageVersion.None);
+    class FallbackBuiltinPythonType : AstPythonType {
+        public FallbackBuiltinPythonType(IBuiltinPythonModule declaringModule, BuiltinTypeId typeId, string name = null) :
```
AlexanderSher: `FallbackBuiltinPythonType` constructor is called only in two places, in both `declaringModule` is `FallbackBuiltinModule` and its first constructor parameter, and `name` is always `null`

src/Analysis/Engine/Impl/Values/BuiltinClassInfo.cs
MikhailArkhipov: Are we keeping IronPython in PTVS? It has been dead for many years.

AlexanderSher: I'm not sure about it: https://github.com/IronLanguages/ironpython2/commits/master

MikhailArkhipov: BTW, comment is misleading. This code is not .NET specific at all. It is primarily used in generating method override snippet in completion when deriving classes...

src/Analysis/Engine/Impl/Interpreter/Ast/AstPythonMultipleMembers.cs
```diff
@@ -134,56 +101,102 @@ class AstPythonMultipleMembers : IPythonMultipleMembers, ILocatedMember {
...
+
+                var unresolved = _members.OfType<ILazyMember>().ToArray();
+                if (unresolved.Length > 0) {
+                    // Publish non-lazy members immediately. This will prevent recursion
+                    // into EnsureMembers while we are resolving lazy values.
+                    var resolved = _members.Where(m => !(m is ILazyMember)).ToList();
+                    _resolvedMembers = resolved.ToArray();
+
+                    foreach (var lm in unresolved) {
+                        var m = lm.Get();
```
AlexanderSher: It is dangerous to call this under lock. `ILazyMember.Get()` triggers `Module.Imported` which can start module analysis.

src/Analysis/Engine/Impl/Interpreter/Ast/AstPythonClass.cs
```diff
@@ -0,0 +1,183 @@
...
+        }
+
+        internal AstPythonClass(string name) : base(name, BuiltinTypeId.Type) { }
+
+        #region IPythonType
+        public override PythonMemberType MemberType => PythonMemberType.Class;
+
+        public override IMember GetMember(IModuleContext context, string name) {
+            IMember member;
+            lock (_lock) {
```
AlexanderSher: Is this a new code or was it moved from somewhere?

src/LanguageServer/Impl/Implementation/CompletionAnalysis.cs
```diff
@@ -884,11 +881,17 @@ private IEnumerable<CompletionItem> GetModules()
...
+
+            sb.AppendLine(result.Name + "(" + string.Join(", ", parameters.Select(MakeOverrideDefParamater)) + "):");
             sb.Append(indentation);
 
-            if (result.Parameters.Length > 0) {
-                var parameterString = string.Join(", ", result.Parameters.Skip(1).Select(MakeOverrideCallParameter));
+            if (parameters.Length > 0) {
+                var parameterString = string.Join(", ", 
+                    result.Parameters
+                        .Where(p => p.Name != "self")
```
jakebailey: I'm not sure this is correct. There's nothing special about the name `self` other than that it's the first parameter. It's not a reserved name like `this` would be in something like C# or C++. I think this change would also break `@classmethod` overrides, potentially.

MikhailArkhipov: This is hacky way to improve another hack. Problem is that code used some acrobatics with `internal _fromFunction` to control if properly returning parameters with overloads should skip first argument. I am not into reworking this since it is all DDG specific. The only place where is seems to matter is here where snippets for overrides are generated in completion. If there is an issue with class methods here, we should add a test and fix the case here.

jakebailey: I'm just trying to say that this:

```python
class A(object):
    def foo(x, y):
        x.y = y

    @classmethod
    def foo(p, q):
		p.q = q
```

Should be just as well supported as:

```python
class A(object):
    def foo(self, y):
        self.y = y

    @classmethod
    def foo(cls, q):
		cls.q = q
```

When generating overrides in class derived from `A`. The old code would just skip the first thing, which should be correct regardless of it being a classmethod or not.

I guess one question I have is that it seems to me like this function is only ever called when the completion item is being generated in a class, in which case an override should only show up in a place where the first argument is the thing to skip. Is that not actually the case here? Maybe I'm just not understanding what you're saying the `fromFunction` thing affects.

src/Analysis/Engine/Impl/Values/BuiltinFunctionInfo.cs
MikhailArkhipov: OK, all to `IsBuiltin` then

src/Analysis/Engine/Impl/OverloadResult.cs
MikhailArkhipov: Yep

src/Analysis/Engine/Impl/Interpreter/Ast/AstPythonMultipleMembers.cs
MikhailArkhipov: Yeah but its been like this forever... Suggestions?

src/Analysis/Engine/Impl/Interpreter/Ast/AstPythonClass.cs
MikhailArkhipov: Moved. It was all in python type which I split into type and class so SetBases don't apply to basic types.

src/LanguageServer/Impl/Implementation/CompletionAnalysis.cs
MikhailArkhipov: Corrected

src/Analysis/Engine/Impl/Interpreter/Ast/AstBuiltinsPythonModule.cs
```diff
@@ -96,13 +95,12 @@ public AstBuiltinsPythonModule(PythonLanguageVersion version)
         }
 
         protected override void PostWalk(PythonWalker walker) {
-            AstPythonBuiltinType boolType = null;
-            AstPythonBuiltinType noneType = null;
+            IPythonType boolType = null;
```
jakebailey: This doesn't necessarily have to be IPythonType. `AstPythonType` implements that interface (as the old `AstPythonBuiltinType` did).

MikhailArkhipov: If they are types, they are `IPythonType`. Unless you prefer specific class?

jakebailey: I guess it may not matter. In the previous code, the type was `AstPythonBuiltinType`, which meant in the code below there was nothing other than `biType` to assign to it, which prevents a future code change from making a mistake and assigning the wrong thing to it.

Unless I'm mistaken, now we don't have `AstPythonBuiltinType`, so we don't know if what is coming out of `_members` is a builtin type or not. I'm not sure what the implications of that change are, though.

MikhailArkhipov: Correct, there is no special 'built-in' class anymore. `IPythonType` provides `IsBuiltin` if type comes from the bultins module. Which is
```cs
bool IsBuiltin => DeclaringModule == null || DeclaringModule is IBuiltinPythonModule
```

MikhailArkhipov: Yep

src/Analysis/Engine/Impl/Values/SequenceBuiltinInstanceInfo.cs
```diff
@@ -26,11 +26,11 @@ namespace Microsoft.PythonTools.Analysis.Values {
     class SequenceBuiltinInstanceInfo : BaseIterableValue {
         private readonly bool _supportsMod;
 
-        public SequenceBuiltinInstanceInfo(BuiltinClassInfo klass, bool sequenceOfSelf, bool supportsMod)
-            : base(klass) {
+        public SequenceBuiltinInstanceInfo(BuiltinClassInfo ClassInfo, bool sequenceOfSelf, bool supportsMod)
```
AlexanderSher: probably `classInfo`.

MikhailArkhipov: Yes

src/Analysis/Engine/Impl/Values/BuiltinInstanceInfo.cs
```diff
@@ -141,7 +139,7 @@ public BuiltinInstanceInfo(BuiltinClassInfo klass)
         public override IEnumerable<OverloadResult> Overloads {
             get {
                 IAnalysisSet callRes;
-                if (_klass.GetAllMembers(ProjectState._defaultContext).TryGetValue("__call__", out callRes)) {
+                if (ClassInfo.GetAllMembers(ProjectState._defaultContext).TryGetValue("__call__", out callRes)) {
```
AlexanderSher: In `TypeId`, `ClassInfo` is checked for `null`, and here it isn't. Which one is correct?

MikhailArkhipov: I checked ctor calls and I don't see how it can be null. And what would be the legal scenario for the null case...

PR #441, opened 2018-11-30
[WIP]: Split analysis into several assemblies

### LINKED ISSUE #257: Move common infrastructure to a separate assembly
We have bunch of common code in the engine, it is worth creating something similar to `Microsoft.Common.Core` in RTVS. Perhaps `Microsoft.DSTool.Core`. And copy tests for common stuff like `CountdownDisposable` etc to this repo.

MikhailArkhipov:
Fixes #257 

`Microsoft.Python.Core` - common objects usable in many places
`Microsoft.Python.Parsing` - Python parser and AST
`Microsoft.Python.Analysis.Engine` - current DDG analysis
`Microsoft.Python.Analysis.Ast` - AST analysis

Reasoning: TDD development of the AST based analysis. Having separate assembly for AST analysis allows development of the DDG replacement without having to remove DDG first. DDG can continue to be used until we have sufficient test bed for the AST analysis to prove it is ready for the prime time.

Based on https://github.com/Microsoft/python-language-server/pull/439.

Still TODO: separate fluent classes for AST analysis since current one use definitions specific to DDG. Therefore tests for both AST and DDG analysis may have to stay in the same assembly for a while.


AlexanderSher:
Ok, I'll have a huge merge for this...

MikhailArkhipov:
No, if this go in, it will happen after earlier PR which is intended for after your imports.

PR #447, opened 2018-12-03
Add None with type NoneType during builtin PostWalk

### LINKED ISSUE #442: Fix None inside of [] in type hints
Something like this warns about use-before-def on `None`:

```python
from typing import Generator

def fun() -> Generator[int, None, None]:
    yield from range(10)
```

`None` needs to be added early like `True` and `False`, as per comments in #429.

Making a separate issue from #391 to better distinguish the two separate bugs.

jakebailey:
Fixes #442.
Replaces #429, which only masks the underlying True/False bug in #391.

I also removed the hack in #336 which checked for `ConstantExpression`.

### DIFF
--- a/src/Analysis/Engine/Impl/Interpreter/Ast/AstBuiltinsPythonModule.cs
+++ b/src/Analysis/Engine/Impl/Interpreter/Ast/AstBuiltinsPythonModule.cs
@@ -90,9 +90,9 @@ namespace Microsoft.PythonTools.Interpreter.Ast {
 #endif
 
             var walker = new AstAnalysisWalker(
-                interpreter, ast, this, filePath, null, _members, 
-                includeLocations, 
-                warnAboutUndefinedValues: true, 
+                interpreter, ast, this, filePath, null, _members,
+                includeLocations,
+                warnAboutUndefinedValues: true,
                 suppressBuiltinLookup: true
             );
             walker.CreateBuiltinTypes = true;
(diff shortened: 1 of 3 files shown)


src/Analysis/Engine/Impl/Interpreter/Ast/AstBuiltinsPythonModule.cs
```diff
@@ -119,6 +120,10 @@ public AstBuiltinsPythonModule(PythonLanguageVersion version)
                     if (typeId == BuiltinTypeId.Bool) {
                         boolType = m as IPythonType;
                     }
+
+                    if (typeId == BuiltinTypeId.NoneType) {
+                        noneType = m as IPythonType;
```
AlexanderSher: Can it be something else than `IPythonType`?

jakebailey: No, because the `AstPythonConstant` constructor accepts that specifically.

AlexanderSher: Then let's do a direct cast: `noneType = (IPythonType)m`;

jakebailey: Done, and some more improvements in the same vein.

PR #460, opened 2018-12-05
Send telemetry on unhandled exceptions inside the AnalysisQueue

### LINKED ISSUE #445: Send telemetry for Analysis_UnhandledException
This function captures unhandled exceptions from the ConsumerLoop. It already calls `LogMessage`, which is LSP, so we should also send a telemetry event like for RPC exceptions.

jakebailey:
Fixes #445.

I've also reworked the telemetry for exception-type events to a common place. If the exception is an AggregateException, then the first inner exception is the "main" exception reported. The outer exception is also included, but in different fields. This makes it easier to group events by where the first exception thrown actually occurs rather than special casing it in some query later.

AggregateExceptions can of course contain more than one inner exception, but because the telemetry properties must be strings, we can't really represent them without encoding something to JSON first and then re-parsing it later. I'd rather not bother until we have a good reason to send a big blob in the telemetry events.

One thing I want feedback on: Is `analyzer.unhandledException` alright for the event name?

### DIFF
--- a/src/LanguageServer/Impl/LanguageServer.cs
+++ b/src/LanguageServer/Impl/LanguageServer.cs
@@ -465,8 +465,6 @@ namespace Microsoft.Python.LanguageServer.Implementation {
             }
 
             var te = Telemetry.CreateEventWithException("analyzer.unhandledException", ex);
-            te.Properties["isTerminating"] = e.IsTerminating.ToString();
-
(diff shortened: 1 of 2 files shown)


jakebailey:
I've changed the event name to `analysis_queue.unhandled_exception` to better match the look/feel of the VSC exception's events.

Issue #407, opened 2018-11-19
Transfer type determination parts from DDG walker to AST walker


MikhailArkhipov:
Transfer valuable analysis/assignment/type determination parts from DDG walker to the AST walker

MikhailArkhipov:
https://github.com/Microsoft/python-language-server/pull/465

Issue #406, opened 2018-11-19
Generics & typing module (support for module specializations)


MikhailArkhipov:
- Instead of using protocols, provide proper implementation of IPythonModule that replaces standard module by specialized implementation.
- Ditto for any module member

Basically rework typeinfo wrapper

MikhailArkhipov:
https://github.com/Microsoft/python-language-server/pull/465

PR #465, opened 2018-12-07
AST analyzer and assemblies split

### LINKED ISSUE #408: Rename objects and namespaces reflecting role changes
- Remove `Ast` prefix since it is no longer special case
- Fix namespaces reflecting new folder structure
- Split files into appropriate subfolder per their roles

### LINKED ISSUE #426: Reimplement module analysis via AST

### LINKED ISSUE #473: Fluent assertions for the AST analysis
Make FA for `IDocumentAnalysis` and friends

MikhailArkhipov:
Fixes #426
Fixes #408
Fixes #473

Base of of #406

Microsoft.Python.Core - common objects usable in many places
Microsoft.Python.Parsing - Python parser and AST
Microsoft.Python.Analysis.Engine - current DDG analysis
Microsoft.Python.Analysis.Core - dependency resolution and other code common to DDG and AST.
Microsoft.Python.Analysis - AST analysis

- No logic changes in DDG analysis apart from minor changes in references due to parser and core utilities relocation.

AST analysis:
- New test assembly with 92 tests
- Updated fluent assertions in the new test assembly
- Asynchronous document loading and analysis
- Asynchronous imports
- Asynchronous AST walk 
- Removed lazy types
- Removed 'multiple types' constructs
- Stub is imported first
- Functions are not analyzed if stub equivalent is present. Instead, overload instance from the stub is used.
- Partial transfer of DDG functionality (#407).


src/Analysis/Ast/Test/FunctionTests.cs
```diff
@@ -256,5 +256,51 @@ class Cxxxx(object):
...
+        public async Task Decorators() {
+            const string code = @"
+class cls(object):
+    @property
+    def a(self): pass
+    
+    @staticmethod
+    def b(): pass
+
+    @abstractproperty
```
jakebailey: Just to warn you ahead of time, `abstractproperty` and `abstractclassmethod` aren't builtins, they come from the `abc` library. A user would do:

```python
import abc

def Foo():
	@abs.abstractproperty
	def foo(): ...
```

or

```python
from abc import abstractproperty

def Foo():
	@abstractproperty
	def foo(): ...
```

Or even `from abs import abstractpropery as absprop`. Unless you're doing this just to have it done for a future change, I'd be really careful what you're showing is in the builtins.

In fact, those two abstract decorators are just wrappers around the builtin: https://github.com/python/cpython/blob/master/Lib/abc.py

MikhailArkhipov: Yes. However, there is runtime case and there is editor case. In the latter we may can mark method as abstract and then check if 'abc' is imported and report an error if it is not (like `@abstract... requires 'abc' module` and maybe even offer code action to fix it. OR we can try and get types from `abc` (provided it is loaded) and if we fail, squiggle as `unknown attribute`. Probably possible to tell that the attribute is `abstract*` and then produce the descriptive error (rather than generic `unknown attribute`). I am fine either way.

jakebailey: Why are the decorators for abstract methods/properties being handled specially at all? I don't see how the latter method is any different than any other handling of decorators, which I'd think we get "for free" because we handle decorators.

The only reason I can think of for storing if something is abstract is to tell the user that they're accessing something they can't access, or have forgotten to implement something, but that sounds like a linting problem that we'd want to handle in the future, not as something required for the analysis as a whole. Is there another case I'm forgetting?

If we're going to end up with a way to do something like "replace the stdlib typing module with this C# code" or "augment X external lib with this extra C# code", then I'd think that implementing abc in a similar way would work out.

MikhailArkhipov: _The only reason I can think of for storing if something is abstract is to tell the user that they're accessing something they can't access_

Yes, that is the reason. I have been adding `TODO` to report diagnostics in certain places so we can go and add it at some point. Most probably there will be some immediate diagnostics such as assignment of incompatible types or missing generic arguments, and then another, deeper pass that may look into other things.

Point is in collecting info when it is available.

jakebailey: I can understand that, but I sort of feel like this will come back to bite us when someone inevitably chooses to refer to those decorators absolutely and we've encoded an assumption. But, as long as it's only going to be used for diagnostics and nothing else, it shouldn't break things in the meantime...

jakebailey:
Running the tests before approving this:

- ListCallCtor fails: Expected Type.TypeId to be Int, but found Unknown.
- Many of the parsing tests fail
  - MutateStdLib gives things like: Assert.AreEqual failed. Expected:<5>. Actual:<4>. 
  - Most roundtrip tests fail, like: Assert.AreEqual failed. Expected:<*>. Actual:< >. Characters differ at 21, got  , expected *

jakebailey:
Alright, the tests seem to pass now, other than the `FullStdLibAnaconda3` scrape test timing out. Building/running the old language server works as well.

PR #467, opened 2018-12-07
Fix for extra paths inside workspace

### LINKED ISSUE #281: Support "go to definition" for namespace packages
https://github.com/Microsoft/vscode-python/issues/231

### LINKED ISSUE #466: Fix "go to definition" and resolving imports
I have `"python.analysis.downloadChannel": "daily"` activated.

The update two days ago broke the 'go to definition' feature for my setup.
I'm using vscode and it's regarding a large plone project using the omelette as described below.

To avoid language server load infinitely (which would be another issue, maybe related to [this](https://github.com/Microsoft/python-language-server/issues/181)), I excluded the `parts/omelette` with `search.exclude`.
Below that in the settings, I added the `parts/omelette` with: `"python.autoComplete.extraPaths": ["${workspaceFolder}/parts/omelette", ],` which seemed to work for the last few days now.
By 'it seemed to work' I mean it could resolve the omelette imports and jump to definitions.
Now it sais 'no definition found for...' and the imports can't be resolved anymore.

For what I can see in the output, before the update, it used to analyse from `autoComplete.extraPath`, which it does not anymore since it updated the language server today.

(`parts/omelette` contains symlinks to cross-project sources in a local directory.)

AlexanderSher:
Fix #281: Support "go to definition" for namespace packages
Fix #466: Fix "go to definition" and resolving imports

The fix is to put user search paths in front of workspace directory so that modules inside extra paths can be used as roots for packages

### DIFF
--- a/src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
+++ b/src/Analysis/Engine/Impl/DependencyResolution/PathResolverSnapshot.cs
@@ -359,11 +359,11 @@ namespace Microsoft.PythonTools.Analysis.DependencyResolution {
                 .ToArray();
 
             var filteredInterpreterSearchPaths = interpreterSearchPaths.Select(FixPath)
-                .Except(filteredUserSearchPaths.Prepend(rootDirectory))
+                .Except(filteredUserSearchPaths.Append(rootDirectory))
                 .ToArray();
 
             userRootsCount = filteredUserSearchPaths.Length + 1;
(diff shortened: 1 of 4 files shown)


jakebailey:
If I have a folder structure like:

```
root
│   main.py
│
├───foo
│       __init__.py
│
└───some
    └───other
        └───foo
                __init__.py
```

Will adding `some/other` to the user search path then prevent the import of the first `foo`?

AlexanderSher:
Yes, PathResolver will pick the first module, which in this case would be `some/other/foo`.

AlexanderSher:
What we can do is add `foo` to `root` as `some.other.foo` and to `root/some/other` as `foo`, but preserve old resolution order.

AlexanderSher:
Unfortunately, this idea won't work because of the #468. If file is added to several roots, there will be more than one module name associated with it, which isn't supported by analyzer.

PR #498, opened 2018-12-21
Fix various common exceptions, log on a failed directory load

### LINKED ISSUE #440: Error when go to definition of PyQt5's module
I have code like following
```python
from PyQt5.QtChart import QChart, QChartView, QLineSeries, QValueAxis
from PyQt5.QtWidgets import QApplication, QTabWidget, QVBoxLayout, QWidget
```

When I go to the definition of `PyQt5.QtChart` or `PyQt5.QtWidgets`, then VSCode popup
the OUTPUT panel and print like following

> [Error - 7:18:34 PM] Request textDocument/definition failed.
  Message: Value cannot be null.
Parameter name: source
  Code: -32000 
   at System.Linq.Enumerable.Select[TSource,TResult](IEnumerable`1 source, Func`2 selector)
   at Microsoft.Python.LanguageServer.Implementation.Server.GetModuleReferences(IPythonProjectEntry entry, PythonAst tree, BufferVersion version, ReferencesParams params)
   at Microsoft.Python.LanguageServer.Implementation.Server.FindReferences(ReferencesParams params, CancellationToken cancellationToken)
   at Microsoft.Python.LanguageServer.Implementation.Server.GotoDefinition(TextDocumentPositionParams params, CancellationToken cancellationToken)
   at Microsoft.Python.LanguageServer.Implementation.LanguageServer.GotoDefinition(JToken token, CancellationToken cancellationToken)

### LINKED ISSUE #448: Server remains open for requests after unhandled "noncritical" exceptions
https://github.com/Microsoft/python-language-server/blob/4a00ee80cb2d7b3e736329e82b2ed441ae70b5ca/src/Analysis/Engine/Impl/Intellisense/AnalysisQueue.cs#L60-L71

When an unhandled exception occurs, and it's deemed "noncritical", the error is logged and then the AnalysisQueue gets disposed. However, the users of the queue have no idea this is happening and will continue to use it. Specifically, the language server will get requests which only lead to ObjectDisposedException.

I'm not sure what the correct behavior is here, but if an exception occurs and we are choosing to dispose of the queue, then I feel like either a new queue needs to be created or the server needs to shut down.

#446 related.

### LINKED ISSUE #463: Catch/ignore/handle exceptions from FileSystemGlobbing
There seem to be quite a few of exceptions coming out of `FileSystemGlobbing`, which is only used inside of `LoadDirectoryFiles`. Most frequent is an `UnauthorizedAccessException`:

```
 at System.IO.Enumeration.FileSystemEnumerator`1.CreateDirectoryHandle(String path, Boolean ignoreNotFound)
 at System.IO.Enumeration.FileSystemEnumerator`1..ctor(String directory, EnumerationOptions options)
 at System.IO.Enumeration.FileSystemEnumerable`1..ctor(String directory, FindTransform transform, EnumerationOptions options)
 at System.IO.Enumeration.FileSystemEnumerableFactory.FileSystemInfos(String directory, String expression, EnumerationOptions options)
 at System.IO.DirectoryInfo.InternalEnumerateInfos(String path, String searchPattern, SearchTarget searchTarget, EnumerationOptions options)
 at Microsoft.Extensions.FileSystemGlobbing.Abstractions.DirectoryInfoWrapper.EnumerateFileSystemInfos()+MoveNext()
 at System.Collections.Generic.List`1.AddEnumerable(IEnumerable`1 enumerable)
 at System.Collections.Generic.List`1.InsertRange(Int32 index, IEnumerable`1 collection)
 at Microsoft.Extensions.FileSystemGlobbing.Internal.MatcherContext.Match(DirectoryInfoBase directory, S
```

But we also see `DirectoryNotFoundException`, `IOException`, `PathTooLongException`, and `FileNotFoundException`.

Right now, these exceptions during `workspace/didChangeConfiguration` go unhandled, produce an error message for the user via RPC, and record the exception.

Looking at the call site:

```csharp
if (!_filesLoaded) {
    await LoadDirectoryFiles();
}
_filesLoaded = true;
```

It may be better idea to do something on the order of:

```csharp
if (!_filesLoaded) {
    try {
        await LoadDirectoryFiles();
        _filesLoaded = true;
    } catch (Exception ex) when (!ex.IsCriticalException()) {
        // Log something to the user
    }
}
```

That way, we don't record events that are really just errors in the user's configuration and also give feedback as to what's wrong.

jakebailey:
Fixes #448.
Fixes #463.
Fixes #440. (I didn't realize `?` would short circuit. Found all mistakes with `rg '\w+(\(\))?\?\.\w+\.Maybe'`)

Also fixes the nullrefs coming from Shutdown.

Opening this PR into `release`, but I can easily switch it to `master` (since they're even).

There is one TODO I'd like feedback on, for the fix to #448.

### DIFF
--- a/src/Analysis/Engine/Impl/Values/BuiltinNamespace.cs
+++ b/src/Analysis/Engine/Impl/Values/BuiltinNamespace.cs
@@ -101,7 +101,7 @@ namespace Microsoft.PythonTools.Analysis.Values {
 
         public virtual ILocatedMember GetLocatedMember() => null;
 
-        public override IEnumerable<ILocationInfo> Locations => GetLocatedMember()?.Locations.MaybeEnumerate();
+        public override IEnumerable<ILocationInfo> Locations => (GetLocatedMember()?.Locations).MaybeEnumerate();
 
         public override bool Equals(object obj) {
             if (obj is BuiltinNamespace<TMemberContainer> bn && GetType() == bn.GetType()) {
--- a/src/Analysis/Engine/Impl/Values/BuiltinNamespace.cs
+++ b/src/Analysis/Engine/Impl/Values/BuiltinNamespace.cs
@@ -16,6 +16,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.PythonTools.Analysis.Infrastructure;
 using Microsoft.PythonTools.Interpreter;
 using Microsoft.PythonTools.Parsing.Ast;
@@ -101,7 +102,7 @@ namespace Microsoft.PythonTools.Analysis.Values {
 
         public virtual ILocatedMember GetLocatedMember() => null;
 
-        public override IEnumerable<ILocationInfo> Locations => (GetLocatedMember()?.Locations).MaybeEnumerate();
+        public override IEnumerable<ILocationInfo> Locations => GetLocatedMember()?.Locations ?? Enumerable.Empty<ILocationInfo>();
 
         public override bool Equals(object obj) {
             if (obj is BuiltinNamespace<TMemberContainer> bn && GetType() == bn.GetType()) {
--- a/src/LanguageServer/Impl/LanguageServer.cs
+++ b/src/LanguageServer/Impl/LanguageServer.cs
@@ -17,6 +17,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
@@ -168,7 +169,17 @@ namespace Microsoft.Python.LanguageServer.Implementation {
                 await _server.DidChangeConfiguration(new DidChangeConfigurationParams { settings = settings }, cancellationToken);
 
                 if (!_filesLoaded) {
-                    await LoadDirectoryFiles();
+                    try {
+                        await LoadDirectoryFiles();
+                    } catch (Exception ex) when (
+                        ex is IOException // FileNotFoundException, DirectoryNotFoundException, PathTooLongException, etc
+                        || ex is UnauthorizedAccessException
+                    ) {
+                        // These exceptions are not caused by the LS, but by invalid/inaccessible user-specified paths.
+                        _server.LogMessage(MessageType.Error, $"Failed to load files in {_initParams.rootPath} - {ex.GetType()}: {ex.Message}");
+                        // TODO: Should this exit completely, or stay open and give the user a chance to fix their config?
+                        return;
+                    }
                 }
                 _filesLoaded = true;
             }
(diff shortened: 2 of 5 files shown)


src/Analysis/Engine/Impl/Values/BuiltinNamespace.cs
```diff
@@ -101,7 +101,7 @@ class BuiltinNamespace<TMemberContainer> : AnalysisValue where TMemberContainer
 
         public virtual ILocatedMember GetLocatedMember() => null;
 
-        public override IEnumerable<ILocationInfo> Locations => GetLocatedMember()?.Locations.MaybeEnumerate();
+        public override IEnumerable<ILocationInfo> Locations => (GetLocatedMember()?.Locations).MaybeEnumerate();
```
MikhailArkhipov: You probably can avoid `MaybeEnumerate` completely with `GetLocatedMember()?.Locations ?? Enumerable.Empty<>()`

jakebailey: Yes, absolutely, I was just making the minimal change. I'll probably do that.

jakebailey:
Any comment on my TODO at the bottom of the PR?

src/LanguageServer/Impl/LanguageServer.cs
```diff
@@ -168,7 +169,17 @@ private void OnUnregisterCapability(object sender, UnregisterCapabilityEventArgs
...
-                    await LoadDirectoryFiles();
+                    try {
+                        await LoadDirectoryFiles();
+                    } catch (Exception ex) when (
+                        ex is IOException // FileNotFoundException, DirectoryNotFoundException, PathTooLongException, etc
+                        || ex is UnauthorizedAccessException
+                    ) {
+                        // These exceptions are not caused by the LS, but by invalid/inaccessible user-specified paths.
+                        _server.LogMessage(MessageType.Error, $"Failed to load files in {_initParams.rootPath} - {ex.GetType().Name}: {ex.Message}");
+                        // TODO: Should this exit completely, or stay open and give the user a chance to fix their config?
```
MikhailArkhipov: I think staying open is OK. Some conflict with, say, virus scanner, should not cause crashes.

jakebailey: I think the answer is probably "no", but should that `LogMessage` instead be a `ShowMessage` to produce a more visible indication that something's wrong? Before, it'd crash, but now it'll (somewhat) silently sit there having not loaded any files.

MikhailArkhipov: We can upgrade the error to `ShowMessage` to give user stronger hint about the problem.

jakebailey: Upgraded to `ShowMessage` and removed the TODO.

Issue #495, opened 2018-12-17
Microsoft.Python.LanguageServer excessive resource usage


d3r3kk:
@KaiLicht commented on [Sun Dec 16 2018](https://github.com/Microsoft/vscode-python/issues/3724)

## Environment data

- VS Code version: 1.30.0
- Extension version (available under the Extensions sidebar): 2018.12.1
- OS and version: macOS 10.14.1
- Python version (& distribution if applicable, e.g. Anaconda): Anaconda, Python 3.7.0
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): conda 
- Relevant/affected Python packages and their versions: --

## Expected behaviour

Microsoft.Python.LanguageServer is using a reasonable amount of ressources.

## Actual behaviour

Microsoft.Python.LanguageServer is using 100%+ CPU Power which is of course draining my battery very fast.

## Steps to reproduce:
1. Install and git pull the library [fast.ai](https://github.com/fastai/fastai)
2. Open the folder in VSCode

## Logs
Output for `Python` in the `Output` panel (`View`→`Output`, change the drop-down the upper-right of the `Output` panel to `Python`)

```
There's no output generated.
```   

Output from `Console` under the `Developer Tools` panel (toggle Developer Tools on under `Help`)

```
[ms-python.python]: Command `python.datascience.exportoutputasnotebook` appears multiple times in the `commands` section.
TMSyntax.ts:46 Overwriting grammar scope name to file mapping for scope source.python.
Old grammar file: file:///Applications/Visual%20Studio%20Code.app/Contents/Resources/app/extensions/python/syntaxes/MagicPython.tmLanguage.json.
New grammar file: file:///Users/kai/.vscode/extensions/magicstack.magicpython-1.1.0/grammars/MagicPython.tmLanguage
e.register @ TMSyntax.ts:46
t._handleGrammarExtensionPointUser @ TMSyntax.ts:329
(anonymous) @ TMSyntax.ts:182
e._handle @ extensionsRegistry.ts:114
e.acceptUsers @ extensionsRegistry.ts:98
t._handleExtensionPoint @ extensionService.ts:490
t._onHasExtensions @ extensionService.ts:344
(anonymous) @ extensionService.ts:332
(anonymous) @ errors.ts:237
s.__generator.a.label @ errors.ts:237
s @ errors.ts:237
Promise resolved (async)
c @ errors.ts:237
s @ errors.ts:237
Promise resolved (async)
[...]

PR #517, opened 2018-12-31
Mitigate some of #495 via MRO member selection and analysis set optimization

jakebailey:
This isn't a complete fix, but does seem to improve #495 in some cases. Adds back the early MRO return removed in #277, so now large class hierarchies won't over-propagate types (where some of the trouble with fastai happens do to the Transform class). I also optimized AnalysisHashSet a little to do length checks when possible.

There are still times when things get caught up comparing unions for a while, but it seems to be nondeterministic.

### DIFF
--- a/src/Analysis/Engine/Impl/AnalysisHashSet.cs
+++ b/src/Analysis/Engine/Impl/AnalysisHashSet.cs
@@ -275,7 +275,15 @@ namespace Microsoft.PythonTools.Analysis.AnalysisSetDetails {
                 return true;
             }
 
+            if (Count == 0 && other.Count == 0) {
+                return true;
+            }
+
             if (Comparer == other.Comparer) {
+                if (Count != other.Count) {
+                    return false;
+                }
+
                 // Quick check for any unmatched hashcodes.
                 // This can conclusively prove the sets are not equal, but cannot
                 // prove equality.
@@ -290,17 +298,16 @@ namespace Microsoft.PythonTools.Analysis.AnalysisSetDetails {
                 }
             }
 
-            var otherHc = new HashSet<AnalysisValue>(other, _comparer);
-            foreach (var key in this) {
-                if (!otherHc.Remove(key)) {
-                    return false;
-                }
-            }
-            if (otherHc.Any()) {
+            if (Count == 0 || other.Count == 0) {
                 return false;
             }
 
-            return true;
+            if (Count == 1 && other.Count == 1) {
+                return _comparer.Equals(this.First(), other.First());
+            }
+
+            var hs = new HashSet<AnalysisValue>(other, _comparer);
+            return hs.SetEquals(this);
         }
 
         /// <summary>
--- a/src/Analysis/Engine/Impl/AnalysisHashSet.cs
+++ b/src/Analysis/Engine/Impl/AnalysisHashSet.cs
@@ -275,12 +275,16 @@ namespace Microsoft.PythonTools.Analysis.AnalysisSetDetails {
                 return true;
             }
 
-            if (Count == 0 && other.Count == 0) {
+            // Only access Count once to prevent wasted time in bucket locks.
+            var lCount = Count;
+            var rCount = other.Count;
+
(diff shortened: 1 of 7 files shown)


src/Analysis/Engine/Impl/AnalysisHashSet.cs
```diff
@@ -275,7 +275,19 @@ public AnalysisHashSet(IEnumerable<AnalysisValue> enumerable, IEqualityComparer<
...
+            // Only access Count once to prevent wasted time in bucket locks.
+            var lCount = Count;
+            var rCount = other.Count;
+
+            if (lCount == 0 && rCount == 0) {
+                return true;
+            }
+
             if (Comparer == other.Comparer) {
+                if (lCount != rCount) {
```
AlexanderSher: This optimization may not work. Code below gets hash codes and collapses them into hash sets, which means that two buckets arrays with different amount of items will be considered equal. I can't say if that is correct or not.

jakebailey: If the comparers are the same, should there ever be a situation where two sets have different contents that are the same? To me that seems related to the issue of keys sometimes changing after being added (that `Debug.Fail`), except that this specific check is costly to do.

AlexanderSher: > Or does a union not actually do any unioning and instead concatenates?

I've seen sets with empty buckets under debugger, but I have no idea how to reproduce it and I don't know if that is a buggy result or an expected one. This code is quite muddy. @MikhailArkhipov , what is your opinion?

> Namely, I started with the count checks, then saw a lot of CPU time used on accessing Count (since it's put behind a lock), then moved the access above.

Just for the reference. [Thin locks](https://blogs.msdn.microsoft.com/seteplia/2017/09/06/managed-object-internals-part-2-object-header-layout-and-the-cost-of-locking/) are really fast, but I don't know if we have thin or fat lock here. `GetHashCode` is overridden, but it isn't the only thing that can promote thin lock to fat lock.

> The MRO lookup change curbs things in some cases.

Yes, MRO fix is clear.

jakebailey: The locking I'm referring to is here: https://github.com/Microsoft/python-language-server/blob/cb1c3dfbad1c7a521857a5dae8eea7c94225a096/src/Analysis/Engine/Impl/AnalysisHashSet.cs#L95-L106

Now that I look more closely, most other functions grab `Bucket` at the start. The original function here did (by grabbing `lBuckets` and maybe grabbing `rBuckets`). Perhaps I should have moved those upward and operated on them instead of introducing two extra possible locks.

I'm happy to remove some of the more suspect length checks if you think this may be an issue (but some of them should still be valid).

jakebailey: Oh, I see, the thin vs fat locking is past `lock` and more to do with object layout, so me showing where the locking is doesn't matter. Oops.

MikhailArkhipov: Different numbers of items (types) in buckets and yet equality seem odd to me. I am not overly concerned with keeping correct number of items in buckets since collecting types from various type of calls is generally not what we want to be doing. Ex, `def f(a): return a` called with `int` now has int return. Called then with `str` it is suddenly show as returning `int, str`. But that's not true, it is not returning combo or a union. And then user adds code calling with bool and it now it returns triplet... Which has nothing to do with the nature of the code.

MikhailArkhipov: New analysis has special types for call arguments and generics so the said function is considered returning type of its argument so no need to collect possible outcomes. `f(1).` will show completion for `int` members and `f('a').` will yield `str` members without need to hold unions.

MikhailArkhipov: In general LGTM. I'd build insiders and let people try it out.

Issue #486, opened 2018-12-12
Query symbol index for document symbols if analysis on document is not complete


jakebailey:
For #433.

jakebailey:
This is obsolete; the indexer will always be queried for info regardless of the status of the analysis. I'll open an issue for allowing the analysis to augment the indexed data (better detection of imports, for example).

PR #521, opened 2019-01-05
Symbol indexer

### LINKED ISSUE #483: Walker to index symbols in AST
For #433.

### LINKED ISSUE #484: Run symbol indexer AST walker in parallel with normal analysis
For #433.

### LINKED ISSUE #485: Query symbol index for workspace symbols
For #433.

jakebailey:
For #433.
Closes #483.
Closes #484.
Closes #485.

General overview:
- Indexing happens on each parse, so with the current LS, every file in the workspace gets indexed. This will need to be reworked.
- All symbol queries go to the index, I have not yet approached #486.
- I do some extra heuristics to highlight some of the more interesting symbols, namely:
  - `__init__` in a `class` will be marked as a constructor.
  - Functions annotated with a property decorator are given the kind `Property`.
  - Functions that fit the pattern `__.*__` are given the kind `Operator`.
  - Other functions in a class are given the kind `Method`.
  - Variables that look like `UPPERCASE_UNDERSCORE_123` at the top level or directly inside a class are given the kind `Constant`.
- Any parse updates the index. I don't do any version management other than to look up the AST on the parse complete event.
- VS-specific "functionKind" is preserved.

Things to do in future PRs:
- Fallback to analysis when it's present. I left all of the functions previously used there.
- Using the import system to answer the question "is this thing being imported a module or not", probably augmented/modified outside the indexer.
- Parse version management.
- Asynchronous behavior (switching the dictionary to be a map from `Uri` to `TaskCompletionSource<HierarchicalSymbol>`).


src/Analysis/Engine/Impl/Indexing/ISymbolIndex.cs
```diff
@@ -0,0 +1,28 @@
...
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.PythonTools.Analysis.Indexing {
+    internal interface ISymbolIndex {
+        Task<IEnumerable<HierarchicalSymbol>> HierarchicalDocumentSymbolsAsync(Uri uri, CancellationToken token = default);
```
AlexanderSher: `Task<IEnumerable<T>>` has weird semantic meaning, something like `Lazy<Lazy<T>>`. It is better to have `Task<IReadOnlyList<T>>`.

src/Analysis/Engine/Impl/Indexing/IndexerWalker.cs
```diff
@@ -0,0 +1,245 @@
...
+
+namespace Microsoft.PythonTools.Analysis.Indexing {
+    internal class IndexerWalker : PythonWalker {
+        private readonly PythonAst _ast;
+
+        private readonly Stack<HierarchicalSymbol> _scope = new Stack<HierarchicalSymbol>();
+        private HierarchicalSymbol Parent => _scope.Count > 0 ? _scope.Peek() : null;
+
+        private bool _skipNext;
+        private bool SkipNext {
```
AlexanderSher: Property is a method that looks like field, so one of the expectation about the property in the single-threaded context is that sequential call produces the same result, i.e.:
```cs
walker.SkipNext = true;
var a = walker.SkipNext;
var b = walker.SkipNext;
Assert.Equal(a, b);
```

src/Analysis/Engine/Impl/Indexing/ISymbolIndex.cs
jakebailey: Yeah, I know. I currently can't do a task which gives back a read only list because I'm using (perhaps abusing) `yield return` to lazily traverse the index, then a `Take` at the use of the enumerable to limit how far it goes based on the limits provided by the LS. `IAsyncEnumerable` hasn't left the proposal phase yet, so I'll have to come up with something else.

AlexanderSher: Do you actually use `Task`?

jakebailey: Likely I'll just end up passing limits down to the index when doing lookups, it just feels a lot less clean and inflexible. In any case, everything here in the first commit is just prototype code that I wanted to push somewhere so I wasn't sitting on it anymore. 😃

src/Analysis/Engine/Impl/Indexing/IndexerWalker.cs
jakebailey: That's a good point, I'll switch it to a function (or just access `_skipNext`). I think I might have had too much fun making a property with side effects.

jakebailey: By the way, the only reason I have `SkipNext` is to skip over function/class NameExpressions, since I pick them up in their definitions instead. I do it like this because I don't seem to have a way to control how things are walked (i.e. it's the function definition node which controls how it is walked, something I can't override). Is there some obvious way to modify how it's walking, or is this a valid thing to do to get around the issue?

src/Analysis/Engine/Impl/Indexing/ISymbolIndex.cs
jakebailey: Missed your question, sorry.

I don't actually use Task, but Mikhail had suggested sticking async on things. The future would be to have another implementation of this interface that instead goes to something like a sqlite database on disk or similar, so I wasn't sure if it was a good idea to make this completely synchronous.

The other thing that I was considering was cancellation; If a user is searching for a symbol, I believe the editor will cancel the request and send a new one when the input changes, so I think I'd need to be passing the cancellation token down so I can allow it to throw when it's cancelled to stop the query. Unfortunately, that has the opposite effect of my `IEnumerable` being tied to that token, so if a user of this interface were to give the query result to something else, that something else may end up being cancelled unexpectedly where it doesn't make sense.

Perhaps I drop the async on this and make it fully synchronous instead, and make a new interface if we need something that's asynchronous (and have the service provide both?).

MikhailArkhipov: I think indexer should be working similar to what async parser and analysis do. I.e. 
- get notification on text changes (probably push rather than listening on events to be less dependent on the source),  - Cancel current indexing on the changed document 
- Start internal indexing task asynchronously.
- Fire event 'indexing complete' on every document and on overall end of indexing.

It can probably run indexing on several files in parallel.

src/Analysis/Engine/Impl/Indexing/IndexerWalker.cs
MikhailArkhipov: You can walk the AST manually. See https://github.com/MikhailArkhipov/python-language-server/blob/analysis10/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs#L68

src/Analysis/Engine/Impl/Indexing/IndexerWalker.cs
```diff
@@ -0,0 +1,245 @@
...
+            get {
+                var v = _skipNext;
+                _skipNext = false;
+                return v;
+            }
+            set { _skipNext = value; }
+        }
+
+        public List<HierarchicalSymbol> Symbols = new List<HierarchicalSymbol>();
```
MikhailArkhipov: Public things should be readonly interfaces

src/Analysis/Engine/Impl/Indexing/IndexerWalker.cs
MikhailArkhipov: You can also override just `Expression` and handle cases in a switch.

src/Analysis/Engine/Impl/Indexing/ISymbolIndex.cs
AlexanderSher: Then I probably don't understand what this method does. Is it a method that awaits for the most recent snapshot or it runs indexing over single document?

src/Analysis/Engine/Impl/Indexing/SymbolExtensions.cs
```diff
@@ -0,0 +1,39 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.PythonTools.Analysis.Indexing {
+    internal static class SymbolExtensions {
+        public static IEnumerable<FlatSymbol> Flatten(this IEnumerable<HierarchicalSymbol> docSyms, Uri uri, string parent = null) {
+            foreach (var sym in docSyms ?? Enumerable.Empty<HierarchicalSymbol>()) {
+                yield return new FlatSymbol {
+                    Name = sym.Name,
```
MikhailArkhipov: I'd rather make an immutable shapshots and update active one when indexing is over. Easier to handle in threaded environments.

src/Analysis/Engine/Impl/Indexing/ISymbolIndex.cs
jakebailey: This method is just the query over the index's state. It does no indexing itself. This is somewhat similar to `GetModuleVariables` which the indexer will partially replace.

src/Analysis/Engine/Impl/Indexing/SymbolIndex.cs
```diff
@@ -0,0 +1,80 @@
...
+                }
+
+                foreach (var child in sym.Children.MaybeEnumerate()) {
+                    ll.AddLast((child, sym.Name));
+                }
+            }
+        }
+
+        public void UpdateParseTree(Uri uri, PythonAst ast) {
+            var walker = new IndexerWalker(ast);
```
MikhailArkhipov: This technically should be canceling existing/starting a new indexing task.

src/LanguageServer/Impl/Implementation/Server.WorkspaceSymbols.cs
```diff
@@ -16,56 +16,95 @@
...
-            return members
-                .GroupBy(mr => mr.Name)
-                .Select(g => g.First())
-                .Select(ToSymbolInformation)
-                .ToArray();
+            // var uri = @params.textDocument.uri;
+            // var symbols = await SymbolIndex.HierarchicalDocumentSymbolsAsync(uri, cancellationToken);
+            // var flat = symbols.Flatten(uri);
+            
+            throw new Exception("this function is not called");
```
MikhailArkhipov: `NotImplementedException`?

jakebailey: This is all WIP, this is a temporary thing I did to "implement" the function. I have no idea if this is called from VS or not.

AlexanderSher:
Design question: is it possible that indexing of one file can trigger indexing of another file?

src/LanguageServer/Impl/Implementation/Server.WorkspaceSymbols.cs
```diff
@@ -16,56 +16,95 @@
...
+            //     .Select(ToSymbolInformation)
+            //     .ToArray();
         }
 
         public override async Task<DocumentSymbol[]> HierarchicalDocumentSymbol(DocumentSymbolParams @params, CancellationToken cancellationToken) {
-            await WaitForCompleteAnalysisAsync(cancellationToken);
-            var opts = GetMemberOptions.ExcludeBuiltins | GetMemberOptions.DeclaredOnly;
-            var entry = ProjectFiles.GetEntry(@params.textDocument.uri);
+            var uri = @params.textDocument.uri;
+            var symbols = await SymbolIndex.HierarchicalDocumentSymbolsAsync(uri, cancellationToken);
```
MikhailArkhipov: This could then await on a `TaskCompletionSource` that would be set when indexing for this particular URI completes.

jakebailey:
> Design question: is it possible that indexing of one file can trigger indexing of another file?

No, I don't think so. Symbols are not analysis dependent, at least not in the way I'm trying to implement it. It's a pure function from file content to symbols.

MikhailArkhipov:
@AlexanderSher - no, it shouldn't as indexing only consumes changes.

AlexanderSher:
>  no, it shouldn't as indexing only consumes changes.

Then I don't understand how indexing can be async, unless we plan to run on different parts of the document in parallel.
Public wrapper methods that allows some code to wait until indexing is complete will be async of course, but I'm not sure that those methods should be part of indexer, cause they will be parser-dependent.

jakebailey:
The end game here is that the indexer does its own parsing instead of relying on some event from the language server to give it the trees. When the new analyzer comes online, there will no longer be a parse event for every file in the project, so the indexer will need to do its own thing in lieu of being given everything. Technically speaking, that doesn't _need_ to be async either, since the walker is fast and the process would generally be "read file, parse, update index", which is just a synchronous action.

There are two sides; the API for querying (ISymbolIndex), and then the other side which provides ASTs, the parsing, etc, which will need to be changed.

MikhailArkhipov:
It should be async as in indexing in its own task/thread upon receiving document change. When done, set result on the task. Like in https://github.com/MikhailArkhipov/python-language-server/blob/analysis10/src/Analysis/Ast/Impl/Modules/PythonModule.cs#L171 and https://github.com/MikhailArkhipov/python-language-server/blob/analysis10/src/Analysis/Ast/Impl/Modules/PythonModule.cs#L244

MikhailArkhipov:
@AlexanderSher _"Meaning we will be doing the same parsing work twice?"_

Possibly, but not always. Indexer processes all files in the workspace. New analyzer is only loading open file and dependencies. So there can be `AST Provider` service where indexer can ask for an AST given URI and if it is not available, create its own. When file does become opened, it can switch then to use live AST and flip back to private when file is closed. Technically analyzer does not need to preserve the AST.

AlexanderSher:
> Like in https://github.com/MikhailArkhipov/python-language-server/blob/analysis10/src/Analysis/Ast/Impl/Modules/PythonModule.cs#L171 and https://github.com/MikhailArkhipov/python-language-server/blob/analysis10/src/Analysis/Ast/Impl/Modules/PythonModule.cs#L244

Ok, so `Parse(CancellationToken cancellationToken)` runs `parser.ParseFile` synchronously, so async isn't for the parsing, but for the callers to wait for the newest AST. And here it needs to know more about those callers. For example:
- If `Update(IEnumerable<DocumentChangeSet> changes)` is called two times, it is possible that two `parser.ParseFile` calls are executed concurrently and second one completes before the first one. As a result, `GetAstAsync` may await on the wrong task and will never get the right one - `_parsingTask` will be set to `null`.
- If `Update` is called multiple times, we can get several parsers running concurrently, and I'm not sure if we can solve this with throttling.
- If we have changes in several files (e.g., user has undone a rename refactoring), we shouldn't start analysis immediately after the first parse is completed.

And so on. That's why I thought that we need some kind of a service that takes care of all parsing requests and provides awaiters for the code that needs to wait for the most recent parse or index.

MikhailArkhipov:
Next update cancels prev parse. May be helpful to prinkle parser with cancellation. In RL parsing takes like 10-20ms tops. Throttle is done in standard way in editors, i.e. parse on idle + ~200 ms (not yet implemented). Changes 99% come in a single active file, unless someone stats flipping across files and changing them. So # of parallel parsers depends more on # of imports that can be going in parallel. I haven't seen more ~2-3 so far.

jakebailey:
Dropping the WIP, since I think it's looking alright. Let me know what you think.

All of the old IMember code is still there in the LS symbols file; I didn't delete any of it even though it's no longer referenced.

src/Analysis/Engine/Impl/Indexing/SymbolExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.PythonTools.Analysis.Indexing {
+    internal static class SymbolExtensions {
+        public static IEnumerable<FlatSymbol> Flatten(this IEnumerable<HierarchicalSymbol> docSyms, Uri uri, string parent = null, int? depthLimit = null) {
+            foreach (var sym in docSyms ?? Enumerable.Empty<HierarchicalSymbol>()) {
```
AlexanderSher: Do we have code that calls `Flatten` with `docSyms == null`?

src/Analysis/Engine/Impl/Indexing/SymbolIndex.cs
```diff
@@ -0,0 +1,66 @@
...
+using Microsoft.PythonTools.Analysis.Infrastructure;
+using Microsoft.PythonTools.Parsing.Ast;
+
+namespace Microsoft.PythonTools.Analysis.Indexing {
+    internal class SymbolIndex : ISymbolIndex {
+        private readonly ConcurrentDictionary<Uri, IReadOnlyList<HierarchicalSymbol>> _index = new ConcurrentDictionary<Uri, IReadOnlyList<HierarchicalSymbol>>();
+
+        public SymbolIndex() { }
+
+        public IEnumerable<HierarchicalSymbol> HierarchicalDocumentSymbols(Uri uri) {
```
AlexanderSher: Looks like you can easily return `IReadOnlyList<HierarchicalSymbol>` here.

src/Analysis/Engine/Impl/Indexing/SymbolIndex.cs
```diff
@@ -0,0 +1,66 @@
...
+            foreach (var kvp in _index) {
+                foreach (var found in WorkspaceSymbolsQuery(query, kvp.Key, kvp.Value)) {
+                    yield return found;
+                }
+            }
+        }
+
+        private IEnumerable<FlatSymbol> WorkspaceSymbolsQuery(string query, Uri uri, IEnumerable<HierarchicalSymbol> symbols) {
+            // Some semblance of a BFS.
+            var ll = new LinkedList<(HierarchicalSymbol, string)>(symbols.Select(s => (s, (string)null)));
```
AlexanderSher: Can you use `Queue<T>` here?

src/Analysis/Engine/Impl/Indexing/SymbolIndex.cs
jakebailey: That's true, though my intention was that a future PR allows this data to be persisted in a local database, where not needing to collect the entire result beforehand. Though, I guess it's a bit pointless when the `HierarchicalSymbol` itself contains an `IList<HierarchicalSymbol>`.

src/Analysis/Engine/Impl/Indexing/SymbolIndex.cs
jakebailey: Yes, though I avoided doing so to prevent a the Queue's backing array from getting too large. If you don't think that's a problem, I can probably switch to using `TraverseBreadthFirst`.

src/Analysis/Engine/Impl/Indexing/SymbolExtensions.cs
jakebailey: No, I have not written any code which does this. `SymbolIndex` never returns a null enumerable.

src/Analysis/Engine/Impl/Indexing/SymbolIndex.cs
jakebailey: Ah, I remember why I didn't use `TraverseBreadthFirst`; The type I want to traverse is `IEnumerable<HierarchicalSymbol>`, which means that `TraverseBreadthFirst` would need to be given `Func<IEnumerable<HierarchicalSymbol>, IEnumerable<IEnumerable<HierarchicalSymbol>>>`, and even then it wouldn't really be the best BFS where I want to traverse multiple nodes' children at once and flatten them into the same queue.

src/Analysis/Engine/Impl/Indexing/SymbolExtensions.cs
AlexanderSher: Then I'd prefer to throw `ArgumentNullException` rather than treat `null` it as `Enumerable.Empty`

src/Analysis/Engine/Impl/Indexing/SymbolIndex.cs
AlexanderSher: Array can be up to 2^31 items. That should be enough for us. If you care about LOH, I'd prefer couple arrays in LOH than tons of `LinkedListNode` instances in G2.

jakebailey: I guess I should rephrase; I don't mean that it will get too large in general, just that it'd grow very quickly as the tree is much, much wider than it is deep and incur a lot of large allocations and copies. I'm happy to change it, just wanted to rephrase my thinking.

src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs
```diff
@@ -0,0 +1,267 @@
...
+            }
+
+            return false;
+        }
+
+        public override bool Walk(FromImportStatement node) {
+            foreach (var imp in node.Names.Zip(node.AsNames, (name, asName) => (name, asName))) {
+                if (imp.asName == null) {
+                    var span = imp.name.GetSpan(_ast);
+                    _stack.AddSymbol(new HierarchicalSymbol(imp.name.Name, SymbolKind.Module, span));
```
AlexanderSher: Btw, why don't you want to save reference to `node` in `HierarchicalSymbol`?

src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs
```diff
@@ -0,0 +1,267 @@
...
+                    var span = imp.asName.GetSpan(_ast);
+                    _stack.AddSymbol(new HierarchicalSymbol(imp.asName.Name, SymbolKind.Module, span));
+                }
+            }
+
+            return false;
+        }
+
+        public override bool Walk(FromImportStatement node) {
+            foreach (var imp in node.Names.Zip(node.AsNames, (name, asName) => (name, asName))) {
```
AlexanderSher: `foreach (var (name, asName) in node.Zip...)`

src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs
jakebailey: What do you mean? Hierarchical symbols only refer to their children, and I'm only treating "declared names" as symbols.

MikhailArkhipov: I think current analysis tends not to keep ASTs around for memory consumption reasons.

AlexanderSher: AST nodes are currently referenced in analysis.

src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs
AlexanderSher: Actually:
```
foreach (var nameNode node.Names.Zip(node.AsNames, (name, asName) => asName ?? name))) {
    var span = nameNode.GetSpan(_ast);
    _stack.AddSymbol(new HierarchicalSymbol(nameNode.Name, SymbolKind.Module, span));
}

File: src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs (hunk 11 above)
Comment (jakebailey) [reply src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs]: Regardless of memory usage, I'm not sure why I'd want to do this. Plus, the symbol classes are technically part of the "public API", so I'm not sure it should need to know anything about the AST.

File: src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs (hunk 12 above)
Comment (jakebailey) [reply src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs]: Ooo, yes, way better.

Comment (jakebailey) [reply src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs]: Agh, unfortunately doesn't help the other `Zip`.

Comment (jakebailey) [reply src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs]: Actually, I can just generate the `span` early.

Comment (jakebailey) [reply src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs]: Maybe not. There's not a super clean way to do the one above unless I write:

```csharp
foreach (var (name, span) in node.Names.Zip(node.AsNames,
	(name, asName) => (asName?.Name ?? name.MakeString(), ((asName as Node) ?? (name as Node)).GetSpan(_ast)))) {
	_stack.AddSymbol(new HierarchicalSymbol(name, SymbolKind.Module, span));
}
```

src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs
AlexanderSher: Ok, what do you expect to be the typical usage of `HierarchicalSymbol`?

src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs
AlexanderSher:
```cs
foreach (var (nameNode, nameString) in node.Names.Zip(node.AsNames, (name, asName) => asName != null ? (asName, asName.Name) : (name, name.MakeString()) {
    var span = nameNode.GetSpan(_ast);
    _stack.AddSymbol(new HierarchicalSymbol(nameString, SymbolKind.Module, span));
}
```

jakebailey: That doesn't quite work, since the compiler can't figure out how to combine `asName` and `name`, but I can do `asName as Node` or `name as Node` and have it figure it out.

src/Analysis/Engine/Impl/Indexing/SymbolIndex.cs
jakebailey: Switched to a `Queue`.

src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs
jakebailey: i.e.:
```csharp
foreach (var (nameNode, nameString) in node.Names.Zip(node.AsNames, (name, asName) => asName != null ? (asName, asName.Name) : (name as Node, name.MakeString()) {
	var span = nameNode.GetSpan(_ast);
	_stack.AddSymbol(new HierarchicalSymbol(nameString, SymbolKind.Module, span));
}
```
Up to you if you think that looks better than the plain if statement.

AlexanderSher: To me it looks better

src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs
jakebailey: The main use (currently) is to return back as `DocumentSymbol`s. There may be some use past that for finding references, but this walker is pretty simple compared to the real analysis' information (which should be preferred).

src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs
jakebailey: `(Node)name` or `node as Name`? This cast can never fail, so `as` is technically overkill.

src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        private void WalkAndDeclareAll(IEnumerable<Node> nodes) {
+            foreach (var node in nodes.MaybeEnumerate()) {
+                WalkAndDeclare(node);
+            }
+        }
+
+        private class SymbolStack {
+            private readonly Stack<(SymbolKind?, List<HierarchicalSymbol>)> _symbols;
```
AlexanderSher: `Stack<(SymbolKind? kind, List<HierarchicalSymbol> symbols)> _symbols`

jakebailey: Done, and the other case where this would help.

src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs
jakebailey: I casted, but let me know if you want `as` instead for looks.

AlexanderSher: `(Node)name` is better

jakebailey: Great, it's done already then.

src/Analysis/Engine/Impl/Indexing/SymbolIndex.cs
jakebailey: Resolving, this will come in a future PR.

src/Analysis/Engine/Impl/Indexing/IndexerWalker.cs
jakebailey: I switched it to a `IReadOnlyList`.

src/Analysis/Engine/Impl/Indexing/IndexerWalker.cs
jakebailey: My rework walks manually when it's needed, instead of the `SkipNext` thing, so resolving.

src/LanguageServer/Impl/Implementation/Server.WorkspaceSymbols.cs
jakebailey: Resolving, as this will come in a future PR.

src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs
AlexanderSher: Ok, let's leave it as it is for now.

src/LanguageServer/Impl/Implementation/Server.WorkspaceSymbols.cs
jakebailey: I ported the VS-specific functionKind, so resolving this.

jakebailey:
We figured out that I forgot to specialize for-loops and generators, so I'm doing that now.

jakebailey:
Added for loops and comprehensions. For example:

image

This to me is much better than other outlines where `s` would get dumped in the wrong scope.

src/Analysis/Engine/Impl/Indexing/SymbolIndexWalker.cs
```diff
@@ -149,6 +149,66 @@ internal class SymbolIndexWalker : PythonWalker {
...
+            _stack.EnterDeclared();
+            node.Else?.Walk(this);
+            _stack.ExitDeclaredAndMerge();
+
+            return false;
+        }
+
+        public override bool Walk(ComprehensionFor node) {
+            AddVarSymbolRecursive(node.Left);
+            return false;
```
jakebailey: This should be `base.Walk(node)` to allow nesting to work. I'll add that and a test for it.

jakebailey: Fixed.

jakebailey:
Any last opinions on this?

MikhailArkhipov:
Ship it.

PR #546, opened 2019-01-22
Language server using new analysis

### LINKED ISSUE #7: Support PEP comment type hints
https://github.com/Microsoft/PTVS/issues/3917

https://www.python.org/dev/peps/pep-0484/
https://www.python.org/dev/peps/pep-0526/

```
PEP_526_style: str = "hello world"
captain: str  # Note: no initial value!
PEP_484_style = SOMETHING # type: str

 
PEP_484_style.
PEP_526_style.upper()
captain.upper()
```

Type declaration is not recognized as string.

image

Jedi

image

Related: https://github.com/Microsoft/PTVS/issues/3674

### LINKED ISSUE #120: `cls` in classmethods of derived classes always refers to the parent class (rather than the derived one)
Easier to explain with an example:

```python
class Foo:
    @classmethod
    def get_an_instance(cls):
        return cls()

class Bar(Foo):
    pass

my_bar = Bar.get_an_instance()
```

In VSCode w/ Python plugin 2018.8.0, `my_bar` is typed as a `Foo`, while I'd expect it to be typed as a `Bar`.

### LINKED ISSUE #471: Language server for the new analysis
Implement LS layer over IDocumentAnalysis

### LINKED ISSUE #492: 'Go To Definition' takes many minutes to resolve in a moderately-sized project
I tried to switch from Jedi to Python Language Server. DonJayamanne, at https://github.com/Microsoft/vscode-python/issues/2790, hinted that Jedi would become unsupported, and suggested creating bugs about why Python Language Server is unusable.

I switched off Jedi, restarted VSCode, opened the project I was working on (which sloccount estimates at about 4000 lines of code), and pressed F12 on a pretty typical local import, which looked like this:

```
from .sibling_module import function_name
```

A busy indicator (without an indication of progress) started moving under the tab bar. 15 minutes later, it's still going.

I would like to be able to provide more information, such as debug output from the language server, so I can tell what kind of rabbit hole it's falling down. But I can't locate any more information at the moment.

I am using VS Code 1.30 on Ubuntu 18.04, running Python 3.6.5 in a virtualenv.

### LINKED ISSUE #531: No auto completion for asyncio.get_event_loop()
LS Version - 0.1.75
Python Code:
```python
import asyncio
loop = asyncio.get_event_loop()
loop.<no completion>
```

### LINKED ISSUE #543: LS: Completion source
Make completion analysis code to use new analysis

### LINKED ISSUE #544: LS: basic hover / tooltips
Make hover/tooltip construction in LS use new analysis

### LINKED ISSUE #545: LS: basic function signatures
Make LS signature help code use new analysis

### LINKED ISSUE #551: Port existing hover and completion tests from DDG

### LINKED ISSUE #563: file.write is not in the completion list
A case to test with the new analysis code:
image

Jedi:
image

### LINKED ISSUE #564: No completions for requests.Response type
A case to test with the new analysis code:
image

requests.get shows it returns a requests.Response type:
image

Jedi:
image

### LINKED ISSUE #567: LS: Goto definition using analysis

### LINKED ISSUE #574: Handle relative imports in stubs
Since typeshed root is not in the path resolver snapshot, `os` stub is unable to resolve `path` submodule since it uses

```python
from . import path as path
```
This causes
```python
import os
os.path.
```
not to show any completions. Works in 3.x. 

Also, for stubs it may be worth resolving stub module first and then real module if the stub is missing. I.e. if A is a stub, import B should resolve to a stub for B and only if stub for B is not found, resolve to real B. This should improve performance of the analysis.

Test: `OsPathMembers`, Python 2.7

MikhailArkhipov:
Closes #543
Closes #544
Closes #545
Closes #471
Closes #551
Closes #531
Closes #574 
Closes #567
Closes #564
Closes #563
Closes #492
Closes #120
Closes #7 

Basically port of `CompletionAnalysis` to use new analysis API.


src/Parsing/Impl/Tokenizer.cs
```diff
@@ -2302,9 +2287,10 @@ struct State : IEquatable<State> {
                     NextWhiteSpace = null;
                 }
                 GroupingRecovery = null;
+                IndentFormat = new string[maxIndent];
             }
 
-            public State(TokenizerOptions options) {
+            public State(TokenizerOptions options, int maxIndent) {
                 Indent = new int[MaxIndent]; // TODO
```
jakebailey: Is `maxIndent` needed if `MaxIndent` exists? I thought `MaxIndent` was used previously to initialize `IndentFormat`, just in the wrong place.

jakebailey:
I'm not sure I understand your implementation of generics. Here's an example which I would expect to work:

```python
from typing import TypeVar, Generic

_T = TypeVar("_T")

class Box(Generic[_T]):
    def __init__(self, v: _T):
        self.v = v
    
    def get(self) -> _T:
        return self.v


boxed = Box(1234)
```

But:

image

image

See also: https://docs.python.org/3/library/typing.html#user-defined-generic-types

```python
from typing import TypeVar, Generic
from logging import Logger, getLogger

T = TypeVar('T')

class LoggedVar(Generic[T]):
    def __init__(self, value: T, name: str, logger: Logger) -> None:
        self.name = name
        self.logger = logger
        self.value = value

    def set(self, new: T) -> None:
        self.log('Set ' + repr(self.value))
        self.value = new

    def get(self) -> T:
        self.log('Get ' + repr(self.value))
        return self.value

    def log(self, message: str) -> None:
        self.logger.info('%s: %s', self.name, message)

v = LoggedVar(1234, "name", getLogger("oh_no"))
v.get().
```

MikhailArkhipov:
Just need to handle `(type_instance)` syntax, aka `CallExpression`. Coming shortly.

src/Analysis/Ast/Test/PepHintTests.cs
```diff
@@ -0,0 +1,80 @@
...
+a = ...  # type: str
+x = ...  # type: int
+primes = []  # type: List[int]
+stats = {}  # type: Dict[str, int]
+
+class Response: # truncated
+    encoding = ...  # type: str
+    cookies = ...  # type: int
+    elapsed = ...  # type: datetime.timedelta
+";
```
jakebailey: Do comment hints work on functions?

- https://www.python.org/dev/peps/pep-0484/#suggested-syntax-for-python-2-7-and-straddling-code
- https://mypy.readthedocs.io/en/latest/python2.html

MikhailArkhipov: Not yet. I made it work in cases that we need to handle stubs better. Generally we probably will handle what we need to provide better completion, but not necessarily all cases to the spec, esp for 2.7. 

Function annotation syntax is a bit more complicated but doable. Do we need it as we needed hints for `requests`?

jakebailey: Not for anything in typeshed, no.

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
```diff
@@ -0,0 +1,193 @@
...
+            : base(root, services) { }
+
+        internal async Task InitializeAsync(CancellationToken cancellationToken = default) {
+            // Add names from search paths
+            await ReloadAsync(cancellationToken);
+            cancellationToken.ThrowIfCancellationRequested();
+
+            // Initialize built-in
+            var moduleName = BuiltinTypeId.Unknown.GetModuleName(_interpreter.LanguageVersion);
+            var modulePath = ModuleCache.GetCacheFilePath(_interpreter.Configuration.InterpreterPath ?? "python.exe");
```
jakebailey: Is `python.exe` actually what the default should be? I can see that it's a carryover from `AstBuiltinsPythonModule.cs`, it just seems strange.

src/Analysis/Core/Impl/Interpreter/InterpreterConfiguration.cs
```diff
@@ -184,7 +184,7 @@ public static InterpreterConfiguration FromDictionary(IReadOnlyDictionary<string
         /// <summary>
         /// Module cache folder.
         /// </summary>
-        public string ModuleCachePath { get; set; }
+        public string DatabasePath { get; set; }
```
jakebailey: I assume this isn't a breaking change? (I think so, given https://github.com/Microsoft/vscode-python/blob/9187479ea42963d28640d8abaaa7345df77c0c7b/src/client/activation/languageServer/analysisOptions.ts#L79)

src/Parsing/Impl/Tokenizer.cs
jakebailey: Pinging on this comment again. I don't think these constructors need to accept `maxIndent`, given `MaxIndent` is a constant that's just being passed down (when it was already accessible, just used incorrectly). It'll work as-is, it's just a little weird.

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
MikhailArkhipov: We should drop default. Either real one, or out :-)

src/Analysis/Core/Impl/Interpreter/InterpreterConfiguration.cs
MikhailArkhipov: Client passes it in property dictionary in
```ts
properties['DatabasePath']
```
which we read via 
```cs
DatabasePath = Read(properties, nameof(DatabasePath));
```
 so it should be OK.

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
MikhailArkhipov: Remove in upcoming PR

AJVA2580:
This doesn't appear to work with python>=3.7.4 type hints

MikhailArkhipov:
@AJVA2580 - please open specific issue with repro steps. This is a very catch-all item for multiple PRs.
https://github.com/microsoft/python-language-server/blob/master/TROUBLESHOOTING.md#filing-an-issue. Thanks.

PR #553, opened 2019-01-24
[WIP] Symbol Index v2: Independent parsing

brianbok:


src/LanguageServer/Impl/Implementation/WorkspaceManager.cs
```diff
@@ -0,0 +1,74 @@
...
+        public async void ParseRootDirAsync() {
+            foreach (var path in await AsyncDirectoryFilePaths()) {
+                _indexingFileParser.ParseForIndex(path);
+            }
+        }
+
+        public void StartParseRootDir() {
+            Task.Run(() => {
+                ParseRootDirAsync();
+            }).Start();
```
brianbok: This is only as a test

src/LanguageServer/Impl/Implementation/RunningDocumentIndexer.cs
```diff
@@ -0,0 +1,40 @@
...
+        private WorkspaceManager _workspaceManager;
+
+        public RunningDocumentIndexer(SymbolIndex index, PythonLanguageVersion pythonLanguageVersion, WorkspaceManager workspaceManager) {
+            _index = index;
+            _pythonLanguageVersion = pythonLanguageVersion;
+            _workspaceManager = workspaceManager;
+        }
+
+        public void OpenDocument(Uri uri, IDocument doc) {
+            _index.UpdateParseTree(uri, new PythonAst(DocAsts(doc)));
```
brianbok: The new code would allow me to write something like
doc.GetAnyAst() for free

src/LanguageServer/Impl/Implementation/IndexingFileParser.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using Microsoft.PythonTools.Analysis.Indexing;
+using Microsoft.PythonTools.Parsing;
+
+namespace Microsoft.Python.LanguageServer.Implementation {
+    class IndexingFileParser {
+        private SymbolIndex _index;
```
MikhailArkhipov: `private readonly`

src/LanguageServer/Impl/Implementation/IndexingFileParser.cs
```diff
@@ -0,0 +1,27 @@
...
+        private PythonLanguageVersion _pythonLanguageVersion;
+
+        public IndexingFileParser(SymbolIndex index, PythonLanguageVersion pythonLanguageVersion) {
+            _index = index;
+            _pythonLanguageVersion = pythonLanguageVersion;
+        }
+
+        public void ParseForIndex(string path) {
+            var uri = new Uri(path);
+            using (var stream = new StreamReader(uri.AbsolutePath)) {
```
MikhailArkhipov: `StringReader`

src/LanguageServer/Impl/Implementation/RunningDocumentIndexer.cs
MikhailArkhipov: You prob want to look into my branch, there are no more document parts etc. One AST per file, single part.

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -344,6 +343,8 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
         #region IPythonLanguageServer
         public ILogger Logger => this;
 
+        internal WorkspaceManager WorkspaceManager { get; private set; }
```
MikhailArkhipov: Generally you should try and code to interfaces for mocking/testability

src/LanguageServer/Impl/Implementation/IndexingFileParser.cs
jakebailey: He'd have to change the interface, since that's a path to a file, not the file contents.

src/LanguageServer/Impl/Implementation/WorkspaceManager.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Python.LanguageServer.Implementation {
+    class WorkspaceManager {
+        private string _rootDir;
+        private string[] _incluideFiles;
```
jakebailey: Typo.

src/LanguageServer/Impl/Implementation/WorkspaceManager.cs
```diff
@@ -0,0 +1,74 @@
...
+            _incluideFiles = includeFiles;
+            _excludeFiles = excludeFiles;
+            _indexingFileParser = indexingFileParser;
+            if (string.IsNullOrEmpty(_rootDir)) {
+                throw new ArgumentException($"{nameof(rootDir)} null or empty", nameof(rootDir));
+            }
+            InitializeMatcher();
+        }
+
+        private Task<IEnumerable<string>> AsyncDirectoryFilePaths() {
```
MikhailArkhipov: All async calls must have cancellation support. Ie add `CancellationToken cancellationToken = default` and when task runs, do `cancellationToken.ThrowIfCancellationRequested` in some places that may take time, like long loops or enumerations.

src/LanguageServer/Impl/Implementation/WorkspaceManager.cs
```diff
@@ -0,0 +1,74 @@
...
+using Microsoft.PythonTools.Parsing;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Python.LanguageServer.Implementation {
+    class WorkspaceManager {
```
MikhailArkhipov: `internal sealed class`. `WorkspaceManager` sounds a bit generic. Since it is specific to the indexer, you may want to add separate `Indexing` folder with namespace like `Microsoft.Python.LanguageServer.Indexing` and perhaps name `WorkspaceIndexingManager`.

src/LanguageServer/Impl/Implementation/WorkspaceManager.cs
```diff
@@ -0,0 +1,74 @@
...
+            });
+        }
+
+        private void InitializeMatcher() {
+            _matcher = new Matcher();
+            _matcher.AddIncludePatterns(_incluideFiles.IsNullOrEmpty() ? new[] { "**/*" } : _incluideFiles);
+            _matcher.AddExcludePatterns(_excludeFiles ?? Enumerable.Empty<string>());
+        }
+
+        private IEnumerable<string> DirectoryFilePaths() {
```
MikhailArkhipov: Cancellable as above

src/LanguageServer/Impl/Implementation/WorkspaceManager.cs
```diff
@@ -0,0 +1,74 @@
...
+
+            var dib = new DirectoryInfoWrapper(new DirectoryInfo(_rootDir));
```
MikhailArkhipov: This should be using `IFileSystem` rather than direct calls

src/LanguageServer/Impl/Implementation/WorkspaceManager.cs
```diff
@@ -0,0 +1,74 @@
...
+            var matchResult = _matcher.Execute(dib);
+            foreach (var file in matchResult.Files) {
+                var path = Path.Combine(_rootDir, PathUtils.NormalizePath(file.Path));
+                if (ModulePath.IsPythonSourceFile(path)) {
+                    yield return path;
```
MikhailArkhipov: yield in async return sounds scare. Just do the list.

src/LanguageServer/Impl/Implementation/WorkspaceManager.cs
MikhailArkhipov: You may want `IReadOnlyList<string>` instead. Tasks returning delay-yield enumerables are scary.

src/LanguageServer/Impl/Implementation/WorkspaceManager.cs
```diff
@@ -0,0 +1,74 @@
...
+                }
+            }
+        }
+
+        public async void ParseRootDirAsync() {
+            foreach (var path in await AsyncDirectoryFilePaths()) {
```
MikhailArkhipov: Cancellation

src/LanguageServer/Impl/Implementation/WorkspaceManager.cs
MikhailArkhipov: `Run` starts the task. Oh, and void starting task is not a good idea. What happens if we want to cancel it.

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -374,7 +375,18 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
...
+                _rootDir = @params.rootUri.ToAbsolutePath();
+            } else if (!string.IsNullOrEmpty(@params.rootPath)) {
+                _rootDir = PathUtils.NormalizePath(@params.rootPath);
+            }
+
             Analyzer = await AnalysisQueue.ExecuteInQueueAsync(ct => CreateAnalyzer(@params.initializationOptions.interpreter, token), AnalysisPriority.High);
+            Version version = Version.Parse(@params.initializationOptions.interpreter.properties.TryGetValue("Version", out var o) ? o as string : null);
+            IndexingFileParser IndexingFileParser = new IndexingFileParser(SymbolIndex, version.ToLanguageVersion());
+            WorkspaceManager = new WorkspaceManager(_rootDir, @params.initializationOptions.includeFiles, @params.initializationOptions.excludeFiles, IndexingFileParser);
+            WorkspaceManager.StartParseRootDir();
```
MikhailArkhipov: WorkspaceManager should be disposable/stoppable. Typically `Dispose` would cancel the work. You can then add the object to the `disposableBag`

src/LanguageServer/Impl/Implementation/WorkspaceManager.cs
jakebailey: Yeah, until `IAsyncEnumerable` hits, I wouldn't mix them.

src/LanguageServer/Impl/Implementation/EditorFiles.cs
```diff
@@ -143,7 +143,7 @@ internal sealed class EditorFile {
                 }
             } finally {
                 if (enqueueForParsing) {
-                    _server.UpdateSymbolIndexFromDoc(uri, doc);
+                    _server.RunningDocumentIndexer.UpdateDocument(uri, doc);
```
MikhailArkhipov: I don't think name related to the RDT since indexer works on all files, not just running ones :-). So just `WorkspaceIndexer` will do.

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -99,6 +99,7 @@ private sealed class ReloadModulesQueueItem : IAnalyzable {
             _editorFiles = new EditorFiles(this);
 
             SymbolIndex = new SymbolIndex();
+            RunningDocumentIndexer = new RunningDocumentIndexer(SymbolIndex, PythonLanguageVersion.V37);
```
MikhailArkhipov: Service container has `IPythonInterpreter` that can tell you the language version.

src/LanguageServer/Impl/Indexing/IndexingFileParser.cs
```diff
@@ -17,8 +17,8 @@ class IndexingFileParser {
 
         public void ParseForIndex(string path) {
```
MikhailArkhipov: Since class is already named `IndexingFileParser` just `Parse` is sufficient.  `IndexingFileParser.ParseForIndex` sounds a bit mouthful :-)

src/LanguageServer/Impl/Indexing/WorkspaceIndexingManager.cs
```diff
@@ -10,17 +10,17 @@
 using System.Text;
 using System.Threading.Tasks;
 
-namespace Microsoft.Python.LanguageServer.Implementation {
-    class WorkspaceManager {
+namespace Microsoft.Python.LanguageServer.Indexing {
+    class WorkspaceIndexingManager {
```
MikhailArkhipov: Probably `WorkspaceIndexManager`

PR #566, opened 2019-01-30
Fix couple of bugs

AlexanderSher:
- Fix nuspec folder
- Fix PTVS issue #5026: Imported packages remain unresolved once a new environment is finished being created


jakebailey:
Can these module reload fixes be backported to `release`?

PR #584, opened 2019-02-06
Output syntax errors from the parser

MikhailArkhipov:


src/Analysis/Ast/Impl/Diagnostics/IDiagnosticsService.cs
```diff
@@ -18,8 +18,25 @@
...
+
+        /// <summary>
+        /// Replaces diagnostics for the document by the new set.
+        /// </summary>
+        void Replace(Uri documentUri, IEnumerable<DiagnosticsEntry> entries);
+
+        /// <summary>
+        /// Removes document from the diagnostics report. Typically when document closes.
+        /// </summary>
+        void Remove(Uri documentUri);
```
AlexanderSher: I thought `Remove` is equal to `Replace(documentUri, Enumerable.Empty<DiagnosticsEntry>())`

MikhailArkhipov: No, Replace with empty means document is active and will be reported as having no errors. Remove means it is closed. VS Code requires full diag snapshot every time for all documents. I guess it is possible to remove documents pretending it does not exist when it has no errors.

MikhailArkhipov: Actually, I'd leave it as it is. The problem is that upon closing document we need to force diagnostic update so entries for the document are removed. In normal case we simply publish normally on idle. I added tests for publishing calls.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.cs
```diff
@@ -33,9 +33,10 @@ namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
     /// Helper class that provides methods for looking up variables
     /// and types in a chain of scopes during analysis.
     /// </summary>
-    internal sealed partial class ExpressionEval : IExpressionEvaluator {
```
jakebailey: I think dropping `IExpressionEvaluator` means `ExpressionEval.Hints.cs` is the file which declares that this class implements the interface, which seems a bit strange (considering I think the interface is really implemented in this file).

src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -376,7 +381,13 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
                     throw new OperationCanceledException();
                 }
                 _ast = ast;
-                _parseErrors = sink.Diagnostics;
+                _parseErrors = sink?.Diagnostics ?? Array.Empty<DiagnosticsEntry>();
+
+                // Do not report issues with libraries or stubs
+                if (sink != null) {
+                    _diagnosticsService?.Replace(Uri, _parseErrors);
```
jakebailey: I'm assuming this is valid because at this point the module is going to be re-analyzed, so we wouldn't be throwing away any diagnostics we actually need to keep?

src/Analysis/Ast/Test/Microsoft.Python.Analysis.Tests.csproj
```diff
@@ -41,6 +42,11 @@
     <ProjectReference Include="..\..\Core\Impl\Microsoft.Python.Analysis.Core.csproj" />
     <ProjectReference Include="..\Impl\Microsoft.Python.Analysis.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <Reference Include="NSubstitute">
+      <HintPath>..\..\..\..\..\Users\mikhaila\.nuget\packages\nsubstitute\4.0.0\lib\netstandard2.0\NSubstitute.dll</HintPath>
```
jakebailey: Not sure if it matters, but the hint path looks like it's specific to your system. The other `PackageReference`s don't have hints. I'm not a NuGet expert, though.

MikhailArkhipov:
Yes should fix this. Looks like nuget manager bug it didn’t used to do this

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.cs
MikhailArkhipov: It can be in any file, but yes, it is better to have it in the main one.

src/Analysis/Ast/Impl/Modules/PythonModule.cs
MikhailArkhipov: Correct. I am adding reporting of imports in the next PR and that will be combining parse + analysis diagnostics.

src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -352,18 +353,22 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
...
             lock (AnalysisLock) {
                 version = _buffer.Version;
-                parser = Parser.CreateParser(new StringReader(_buffer.Text), Interpreter.LanguageVersion, new ParserOptions {
-                    StubFile = FilePath != null && Path.GetExtension(FilePath).Equals(".pyi", FileSystem.StringComparison),
-                    ErrorSink = sink
-                });
+                var options = new ParserOptions {
+                    StubFile = FilePath != null && Path.GetExtension(FilePath).Equals(".pyi", FileSystem.StringComparison)
+                };
+                if (ModuleType == ModuleType.User) {
```
jakebailey: The only downside of this is that if there's a syntax error or other diagnostic in a library file and the user does a go-to-def, nothing will show up (since the module type isn't user, so errors/etc are not stores). I'm not sure how critical that is, though.

MikhailArkhipov: This only affects error collection, the library goto def should still work, provided library is in source.

jakebailey: I mean that the newly opened non-user file will not show any diagnostics because they were never collected, not that go-to-def won't work.

MikhailArkhipov: Yes, that is intentional. Diagnostics would be noise for the user since this is not something that is actionable for them.

PR #593, opened 2019-02-10
Report unresolved imports

### LINKED ISSUE #512: Diagnostic list flickers annoyingly when opening a document
In VSCode, if you display all diagnostics for your project (i.e. set the "python.analysis.openFilesOnly" to "false"), you can view the list in the "Problems" panel. If you open a Python file that contains a reported diagnostic the analysis diagnostics for that file briefly disappear and then reappear. This is not only aesthetically unpleasing, it's also disorienting if you just clicked on one of these items to open the file and view an error in the editor.

The problem appears to be that the PLS is reporting diagnostics in two phases: parsing and analysis. When a file is closed, the server sees version -1 of the file, but when it's opened, the server sees version 1. This causes the server to re-parse and re-analyze the file from scratch. (As an aside, this seems wasteful given that the file hasn't changed. Maybe it could be skipped altogether?) When the parse is complete, the server reports the incomplete diagnostics, which don't contain the analysis drags. Then shortly later it reports the full diagnostics. In the meantime, VSCode has updated the diagnostic list in the "Problems" panel.

I don't fully understand the original reason behind updating the parse and analysis errors independently, so I'm not sure of the correct fix. My attempt at a fix is to comment out the following three lines from EditorFile.UpdateParseDiagnostics:
```
// if (diags != null) {
//     PublishDiagnostics(diags);
// }
```
This skips the reporting of the partial diagnostics and waits for completion of the analysis. With this change, the annoying flicker, and I haven't noticed any bad side effects.

### LINKED ISSUE #583: Report unresolved imports

MikhailArkhipov:
Closes #583
Closes #512

- Adds reporting of unresolved imports
- Incorporates suggestions from #512
- Restores message severity mapping
- Addresses comments missed in https://github.com/Microsoft/python-language-server/pull/546 review.
- Tests

I guess I will need to re-fork my master since all commits from earlier work still drag behind.


jakebailey:
I don't know the exact line in the PR to comment on but for `as` imports, but I think that the original name is the more useful one to be reporting. For example, `matplotlib`/`matplotlib.pyplot` is the thing that's missing here, but the message reports `plt`:

image

jakebailey:
For example, in the old LS:

image

MikhailArkhipov:
Yep, changing

src/Analysis/Ast/Test/ImportTests.cs
```diff
@@ -151,5 +154,59 @@ import sys
...
+
+        [TestMethod, Priority(0)]
+        public async Task UnresolvedImport() {
+            var analysis = await GetAnalysisAsync(@"import nonexistent");
+            analysis.Should().HaveVariable("nonexistent")
+                .Which.Value.GetPythonType<IPythonModule>().ModuleType.Should().Be(ModuleType.Unresolved);
+            analysis.Diagnostics.Should().HaveCount(1);
+            var d = analysis.Diagnostics.First();
+            d.ErrorCode.Should().Be(ErrorCodes.UnresolvedImport);
+            d.SourceSpan.Should().Be(1, 8, 1, 19);
```
AlexanderSher: Can we use `Range` instead of `SourceSpan`?

src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -460,6 +460,11 @@ public override void Add(string message, SourceSpan span, int errorCode, Severit
                     OnAnalysisComplete();
                     ContentState = State.Analyzed;
 
+                    // Do not report issues with libraries or stubs
```
AlexanderSher: What if user has opened library file explicitly?

src/LanguageServer/Impl/Diagnostics/DiagnosticsService.cs
```diff
@@ -91,7 +110,7 @@ internal sealed class DiagnosticsService : IDiagnosticsService, IDisposable {
 
         private void PublishDiagnostics() {
             lock (_lock) {
-                foreach (var kvp in _diagnostics) {
+                foreach (var kvp in Diagnostics) {
```
AlexanderSher: This loop calls third party method under lock. We can move it outside.

src/Analysis/Ast/Impl/Analyzer/Handlers/ImportHandler.cs
```diff
@@ -48,20 +52,21 @@ internal sealed partial class ImportHandler: StatementHandler {
...
                     case ModuleImport moduleImport when moduleImport.FullName == Module.Name:
                         Eval.DeclareVariable(memberName, Module, VariableSource.Declaration, location);
                         break;
                     case ModuleImport moduleImport:
-                        module = await HandleImportAsync(node, moduleImport, cancellationToken);
+                        module = await HandleImportAsync(moduleImport, location, cancellationToken);
                         break;
                     case PossibleModuleImport possibleModuleImport:
-                        module = await HandlePossibleImportAsync(node, possibleModuleImport, cancellationToken);
+                        module = await HandlePossibleImportAsync(possibleModuleImport, moduleName, location, cancellationToken);
```
AlexanderSher: We should use `PossibleModuleImport.PossibleModuleFullName` instead of `moduleName`. Do we use `DottedName.MakeString()` anywhere else?

src/Analysis/Ast/Impl/Analyzer/Handlers/FromImportHandler.cs
```diff
@@ -33,15 +33,14 @@ internal sealed partial class ImportHandler {
...
-            IImportSearchResult imports = null;
             if (rootNames.Count == 1) {
                 var rootName = rootNames[0].Name;
                 if (rootName.EqualsOrdinal("__future__")) {
                     return false;
                 }
             }
 
-            imports = ModuleResolution.CurrentPathResolver.FindImports(Module.FilePath, node);
+            var imports = ModuleResolution.CurrentPathResolver.FindImports(Module.FilePath, node);
```
AlexanderSher: Below this line it should be `mi.FullName == Module.Name`. And you can include this line into switch.

src/Analysis/Ast/Impl/Modules/PythonModule.cs
jakebailey: Also see: https://github.com/Microsoft/python-language-server/pull/584#discussion_r255287562

src/Analysis/Ast/Test/ImportTests.cs
MikhailArkhipov: Range is not in analysis, it is in LSP. Can theoretically move it down.

src/Analysis/Ast/Impl/Modules/PythonModule.cs
MikhailArkhipov: We don’t report issues in libraries, this is not something actionable to the user. Something like resharper does not produce issues in disassembled code...

src/Analysis/Ast/Test/ImportTests.cs
AlexanderSher: I thought `Diagnostics` is part of LSP as well. We also have `IndexSpan`.

MikhailArkhipov: `Diagnostics` yes. `DiagnosticsEntry`, no. Switching away from SourceSpan would require changing tens of tests. Analysis is using `LocationInfo` which produces `SourceSpan`.

src/LanguageServer/Impl/Diagnostics/DiagnosticsService.cs
MikhailArkhipov: Sure. We only need `_changed` there.

src/Analysis/Ast/Impl/Analyzer/Handlers/ImportHandler.cs
MikhailArkhipov: Not for analysis. It is used here and in `HoverSource` just to get the string to show. I update here as you suggested.

src/Analysis/Ast/Impl/Analyzer/Handlers/FromImportHandler.cs
MikhailArkhipov: OK

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsSeverityMap.cs
```diff
@@ -0,0 +1,45 @@
...
+
+namespace Microsoft.Python.Analysis.Diagnostics {
+    public sealed class DiagnosticsSeverityMap {
+        private readonly Dictionary<string, Severity> _map = new Dictionary<string, Severity>();
+
+        public DiagnosticsSeverityMap() { }
+
+        public DiagnosticsSeverityMap(string[] errors, string[] warnings, string[] information, string[] disabled) {
+            _map.Clear();
+            // disabled > error > warning > information
```
AlexanderSher: Is it compatible with VSC?

src/LanguageServer/Impl/Diagnostics/DiagnosticsService.cs
```diff
@@ -74,6 +83,16 @@ internal sealed class DiagnosticsService : IDiagnosticsService, IDisposable {
         }
 
         public int PublishingDelay { get; set; } = 1000;
+
+        public DiagnosticsSeverityMap DiagnosticsSeverityMap {
+            get => _severityMap;
+            set {
+                lock (_lock) {
+                    _severityMap = value;
+                    PublishDiagnostics();
```
AlexanderSher: One more place under lock.

src/LanguageServer/Impl/Diagnostics/DiagnosticsService.cs
```diff
@@ -90,16 +109,19 @@ internal sealed class DiagnosticsService : IDiagnosticsService, IDisposable {
...
+            KeyValuePair<Uri, IReadOnlyList<DiagnosticsEntry>>[] diagnostics;
             lock (_lock) {
-                foreach (var kvp in _diagnostics) {
-                    var parameters = new PublishDiagnosticsParams {
-                        uri = kvp.Key,
-                        diagnostics = kvp.Value.Select(ToDiagnostic).ToArray()
-                    };
-                    _clientApp.NotifyWithParameterObjectAsync("textDocument/publishDiagnostics", parameters).DoNotWait();
-                }
+                diagnostics = Diagnostics.ToArray();
```
AlexanderSher: We aren't calling `_diagnostics.Clear();` anywhere. Is that correct?

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsSeverityMap.cs
MikhailArkhipov: Yes. Same code as before

src/LanguageServer/Impl/Diagnostics/DiagnosticsService.cs
MikhailArkhipov: As a matter of fact, I made PublishDiagnostics inside locks intentionally. The problem here if publishing goes off the lock, then we can go out of sync between `_changed` and the state of the collection. For example,

```cs
        public void Remove(Uri documentUri) {
            lock (_lock) {
                // Before removing the document, make sure we clear its diagnostics.
                _diagnostics[documentUri] = new List<DiagnosticsEntry>();
                PublishDiagnostics();
                _diagnostics.Remove(documentUri);
            }
        }
```

`PublishDiagnostics` must be called between `_diagnostics[documentUri] = new List<DiagnosticsEntry>()` which makes collection of the error empty and VSC clears the task list and `_diagnostics.Remove(documentUri)` which removes the document. Calling `PublishDiagnostics` after removal will lead to stale errors in the error list.

src/LanguageServer/Impl/Diagnostics/DiagnosticsService.cs
MikhailArkhipov: `ClearAllDiagnostics`, on Dispose. Otherwise it stays.

PR #596, opened 2019-02-12
Docstring converter

### LINKED ISSUE #548: Rework docstring to markdown converter
For #424. See also #422, #17, #444.

RestTextConverter worked by doing some amount of ReST parsing, but isn't always correct in all cases.

It may be worthwhile to use an actual ReST parser (perhaps an existing grammar).

jakebailey:
Closes #548.

Still missing Google/Numpy style param/return/etc list parsing. I'm also not too happy with the code duplication in the markdown documentation source.

### DIFF
--- a/src/LanguageServer/Impl/Documentation/DocstringConverter.cs
+++ b/src/LanguageServer/Impl/Documentation/DocstringConverter.cs
@@ -139,24 +139,10 @@ namespace Microsoft.Python.LanguageServer.Documentation {
                 return false;
             }
 
+            // TODO: Push into Google/Numpy style list parser.
+
             AppendTextLine(CurrentLine);
             return true;
-        }
-
-        private string PreprocessTextLine(string line) {
-            // http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#literal-blocks
-            if (Regex.IsMatch(line, @"^\s*::$")) {
-                return string.Empty;
-            }
-            line = Regex.Replace(line, @"\s+::$", "");
-            line = Regex.Replace(line, @"(\S)\s*::$", "$1:");
-
-            // http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#interpreted-text
-            line = Regex.Replace(line, @":[\w_\-+:.]+:`", "`");
-            line = Regex.Replace(line, @"`:[\w_\-+:.]+:", "`");
-
-            line = line.Replace("``", "`");
-            return line;
         }
 
         private void AppendTextLine(string line) {
@@ -164,7 +150,7 @@ namespace Microsoft.Python.LanguageServer.Documentation {
 
             // Hack: attempt to put directives lines into their own paragraphs.
             // This should be removed once proper list-like parsing is written.
-            if (!_insideInlineCode && Regex.IsMatch(line, @"^:(param|arg|type|return|rtype|raise|except|var|ivar|cvar|copyright|license)")) {
+            if (!_insideInlineCode && Regex.IsMatch(line, @"^\s*:(param|arg|type|return|rtype|raise|except|var|ivar|cvar|copyright|license)")) {
                 AppendLine();
             }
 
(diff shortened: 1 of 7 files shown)


jakebailey:
Oops, accidentally hit "submit PR" early. Still a little bit to be done, but it works very well as it is.

src/LanguageServer/Impl/Documentation/DocstringConverter.cs
```diff
@@ -0,0 +1,460 @@
...
+            }
+
+            var parts = line.Split('`');
+
+            for (var i = 0; i < parts.Length; i++) {
+                var part = parts[i];
+
+                if (i > 0) {
+                    _insideInlineCode = !_insideInlineCode;
+                    Append("`");
```
MikhailArkhipov: Can be Append(‘’) ie char

src/LanguageServer/Impl/Documentation/DocstringConverter.cs
```diff
@@ -0,0 +1,460 @@
...
+        /// first strip any common leading indention (like inspect.cleandoc),
+        /// then remove duplicate empty/whitespace lines.
+        /// </summary>
+        /// <param name="docstring">The docstring to convert, likely from the AST.</param>
+        /// <returns>The converted docstring, with Environment.NewLine line endings.</returns>
+        public static string ToPlaintext(string docstring) {
+            var lines = SplitDocstring(docstring);
+            var output = new List<string>();
+
+            foreach (var line in lines) {
```
MikhailArkhipov: Can be full Linq if you want, condition in continue is Where.

src/LanguageServer/Impl/Documentation/DocstringConverter.cs
```diff
@@ -0,0 +1,460 @@
...
+            }
+
+            PushAndSetState(ParseDirective);
+            _blockIndent = NextBlockIndent;
+            _appendDirectiveBlock = false;
+            return true;
+        }
+
+        private bool ParseDirective() {
+            // http://docutils.sourceforge.net/docs/ref/rst/restructuredtext.html#directives
```
MikhailArkhipov: Do we need to add credits to license file?

jakebailey: I'm not sure; I only used it as a reference (like we'd use the Python spec or similar). There's no actual parser described at that reference. These notes are just to point to the right places since the naming of certain terms in ReST is a bit hard to remember.

src/LanguageServer/Impl/Documentation/DocstringConverter.cs
jakebailey: Sure, I can add an overload for that below (though this converter is not exactly the most efficient given the large number of replacement, regex matches, and string splits/joins...).

src/LanguageServer/Impl/Documentation/DocstringConverter.cs
jakebailey: Not exactly, since that condition looks at the output to prevent empty lines from being duplicated. Best I could do is:

```csharp
var lines = SplitDocstring(docstring);
var output = lines.Where(
	(l, i) => !string.IsNullOrWhiteSpace(l) || !string.IsNullOrWhiteSpace(lines.ElementAtOrDefault(i - 1))
);
return string.Join(Environment.NewLine, output).TrimEnd();
```

Which I can do if you find that more readable.

jakebailey:
@MikhailArkhipov Do you want this merged even without Numpy/Google style docstring support and the copying between plaintext/markdown doc sources?

MikhailArkhipov:
Up to you. I think more features can be added in stages, doesn’t have to handle everything off the bat. If it handles typical cases, it is already way better than nothing :-)

jakebailey:
Fair enough. I'll spend a little time seeing if there's a way to make them a little more presentable (like how I handle `:param:` for n ow), then merge it in and make issues for those other parts.

PR #600, opened 2019-02-13
Resolve issue when specific types shared data after creation off generic template

### LINKED ISSUE #590: Parameterized types of generic classes sometimes shared between class instances
In the below, both `intval` and `strval` come to be `int`, even though hovering over `boxedint` and `boxedstr` show `Box[int]` and `Box[str]` respectively.

```python
from typing import TypeVar, Generic

_T = TypeVar("_T")

class Box(Generic[_T]):
    def __init__(self, v: _T):
        self.v = v
    
    def get(self) -> _T:
        return self.v


boxedint = Box(1234)
intval = boxedint.get()

boxedstr = Box("foo")
strval = boxedstr.get()
```

MikhailArkhipov:
Fixes #590
Includes https://github.com/Microsoft/python-language-server/pull/593

The problem was that when specific type is created it shares method definitions with the original template. Now function determines return type dynamically. Specific type stores original type arguments. This is necessary to determine function return type when function is a method of the specific type but annotated as returning generic type. When specific type is created, it received value-type members (variables) of the original generic class with updated types. Methods are not cloned. Instead, specific class type stores the original type arguments and function returning generic uses them to determine its actual return type. When 'called' the function received actual class it is bound to (self) and can extract the stored type arguments.


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
```diff
@@ -47,57 +48,61 @@ internal sealed partial class ExpressionEval {
             // rather than specific type instantiation as in List[str].
 
             IPythonType[] specificTypes;
+            var returnInstance = false;
             switch (expr) {
+                // Indexing returns type as from A[int]
                 case IndexExpression indexExpr: {
-                    // Generic[T1, T2, ...] or A[type]()
```
jakebailey: Was this formatting change done for a specific reason? Seems weird to have an extra indent for just this one case, but maybe there's some C# semantic you're using that I'm missing.

MikhailArkhipov: Yeah it is C# formatter bug with comments inside nested block. I'll fix that.

jakebailey:
Hmm, it seems like the `Box` example is now getting more members than it should:

image

src/Analysis/Ast/Impl/Types/Definitions/IPythonFunctionOverload.cs
```diff
@@ -50,17 +50,17 @@ public interface IPythonFunctionOverload {
...
         /// Return value documentation.
         /// </summary>
-        string ReturnDocumentation { get; }
-
-        /// <summary>
-        /// Function definition is decorated with @overload.
-        /// </summary>
-        bool IsOverload { get; }
+        /// <param name="self">If function is in generic class it may need the specific type to determine the return type.</param>
+        string GetReturnDocumentation(IPythonType self);
```
jakebailey: Should this be `IPythonType self = null` to get rid of all of the calls that look like `GetReturnDocumentation(null)`, and to better indicate that `self` is allowed to be null? I can see other functions you've modified that do `self = null`, but wasn't sure if you were skipping this specific one to make it explicit that there is no `self`.

MikhailArkhipov: Sure why not. I made it without `= null` so I could find and fix relevant cases.

jakebailey: That's what I had guessed; if keeping `(null)` prevents future errors, like forgetting to specify `self`, then I wouldn't remove it.

PR #602, opened 2019-02-13
De-duplicate completions 

### LINKED ISSUE #591: Some import suggestions are duplicated
image

Perhaps due to duplicate module names in the search paths.

MikhailArkhipov:
Fixes #591


jakebailey:
I'm still seeing duplicates:

image

image

Seems like now it's being limited to two copies instead of N copies.

MikhailArkhipov:
Yes, there is a) duplicated chunk of code and b) when adding test I found that `sys` and other builtins are missing from completion b/c of 

```cs
        private static IEnumerable<string> GetModuleNames(IEnumerable<Node> roots) => roots
            .SelectMany(r => r.TraverseBreadthFirst(n => n.IsModule ? Enumerable.Empty<Node>() : n.Children))
            .Where(n => n.IsModule)
            .Select(n => n.FullModuleName);
```

where `n.IsModule` check for ModulePath but builtins don't have it.

PR #606, opened 2019-02-13
Port import tests from old LS

### LINKED ISSUE #540: Port imports test to new analysis
Current implementation uses completions. Need something more down to Earth.

MikhailArkhipov:
Closes #540.
Also fixes case where there were no completions after `from typing import`


src/LanguageServer/Test/ImportsTests.cs
```diff
@@ -0,0 +1,301 @@
...
+            rdt.OpenDocument(new Uri(init2Path), string.Empty);
+            rdt.OpenDocument(new Uri(init3Path), string.Empty);
+            rdt.OpenDocument(new Uri(init4Path), string.Empty);
+
+            var doc = rdt.OpenDocument(new Uri(appPath), appCode, appPath);
+            var analysis = await doc.GetAnalysisAsync();
+
+            var cs = new CompletionSource(new PlainTextDocumentationSource(), ServerSettings.completion);
+            var comps = await cs.GetCompletionsAsync(analysis, new SourceLocation(7, 10));
+            comps.Should().HaveLabels("foo");
```
jakebailey: I know in #540 you said you didn't want to use completion to test these; are you fine with this for now then?

src/Analysis/Ast/Impl/Modules/Resolution/ModuleResolutionBase.cs
```diff
@@ -77,8 +77,16 @@ internal abstract class ModuleResolutionBase {
...
-            => _modules.TryGetValue(name, out var module) ? module : null;
+        public IPythonModule GetImportedModule(string name) {
+            var module = _interpreter.ModuleResolution.GetSpecializedModule(name);
+            if (module != null) {
+                return module;
+            }
+            return _modules.TryGetValue(name, out module) ? module : null;
+        }
+
+        public void SetUserSearchPaths(in IEnumerable<string> searchPaths)
```
jakebailey: With this added to the interface, MainModuleResolution can now probably call this instead of going to the path resolver directly.

src/LanguageServer/Test/ImportsTests.cs
MikhailArkhipov: Yes, it is simpler this way for now

PR #617, opened 2019-02-14
Fix issues with 'import not found' messages and missing bultins

### LINKED ISSUE #601: Sometimes no analysis for pip installed libraries
Some libraries installed via pip such as `numpy` end up with an message like:

```
[Info  - 12:17:31 PM] Import not found: numpy
```

And no analysis info.

Sometimes the analysis gets hung up on analyzing some pip related libraries and appears to not proceed, but that may be another issue.

MikhailArkhipov:
Fixes #601
- Change message for 'import not found' to clearly indicate stub
- Fix issue with missing builtin modules in import completion
- Fix hang on recursion into the same user module


MikhailArkhipov:
I will refork branch then re-submit

PR #621, opened 2019-02-15
Add dependency resolution to the analyzer

AlexanderSher:


src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -14,145 +14,366 @@
...
+        private struct ModuleKey : IEquatable<ModuleKey> {
+            public string Name { get; }
+            public string FilePath { get; }
+            public bool IsTypeshed { get; }
+
+            public ModuleKey(IPythonModule module) {
+                Name = module.Name;
+                FilePath = module.ModuleType == ModuleType.CompiledBuiltin ? null : module.FilePath;
+                IsTypeshed = module.ModuleType == ModuleType.Stub;
             }
```
MikhailArkhipov: What about user stubs ie stubs next to the user code (pyi) and not in the typeshed paths. They would have `module.ModuleType == ModuleType.Stub`

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -14,145 +14,366 @@
...
-                node.Analyzable.NotifyAnalysisPending();
+        [DebuggerDisplay("{Name} : {FilePath}")]
+        private struct ModuleKey : IEquatable<ModuleKey> {
+            public string Name { get; }
+            public string FilePath { get; }
+            public bool IsTypeshed { get; }
+
+            public ModuleKey(IPythonModule module) {
+                Name = module.Name;
+                FilePath = module.ModuleType == ModuleType.CompiledBuiltin ? null : module.FilePath;
```
MikhailArkhipov: I think just `compiled` won't have file path either. I.e. compiled libraries that are not part of the language library set.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
AlexanderSher: Good point. Do you think we should add another flag for `Typeshed`?

MikhailArkhipov: Gets complicated... Is it used to tell which module resolution to use? Maybe just attach the module resolution reference.

MikhailArkhipov:
If you wish, merge https://github.com/Microsoft/python-language-server/pull/617 here, it is related.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -14,145 +14,366 @@
...
+                }
+
+                using (var cts = CancellationTokenSource.CreateLinkedTokenSource(timeoutCts.Token, cancellationToken)) {
+                    var timeoutToken = timeoutCts.Token;
+                    while (!timeoutToken.IsCancellationRequested) {
+                        try {
+                            var analysis = await entry.GetAnalysisAsync(cts.Token);
+                            lock (_syncObj) {
+                                if (entry.Version == analysis.Version) {
+                                    return analysis;
```
MikhailArkhipov: So it goes in and out of lock. Do we care if `entry` changes in the dictionary between we fetched it and when we reached `entry.Version == analysis.Version` ?

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -14,145 +14,366 @@
...
+            return entry.PreviousAnalysis;
         }
 
-        public void Dispose() => _globalCts.Cancel();
+        private async Task AnalyzeDocumentAsync(IPythonModule module, PythonAst ast, int version, CancellationToken cancellationToken) {
+            var key = new ModuleKey(module);
+            AnalysisEntry entry;
+            lock (_syncObj) {
+                if (_analysisEntries.TryGetValue(key, out entry)) {
+                    entry.Invalidate(_version);
```
MikhailArkhipov: `_version` or `version`?

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -14,145 +14,366 @@
...
-            // the same at the time the analysis completes.
-            var analysisVersion = node.Analyzable.ExpectedAnalysisVersion;
+            public ModuleDependencyFinder(IFileSystem fileSystem) {
+                _fileSystem = fileSystem;
+            }
+
+            public Task<ImmutableArray<ModuleKey>> FindDependenciesAsync(AnalysisEntry value, CancellationToken cancellationToken) {
+                var dependencies = new HashSet<ModuleKey>();
+                var module = value.Module;
+                var isTypeshed = module.ModuleType == ModuleType.Stub; // TODO: This is not the correct way to determine Typeshed.
```
MikhailArkhipov: I think `ModuleKey` has `IsTypeshed`. Or maybe `ModuleKey` could just hold the desired resolution service instance.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
AlexanderSher: AFAIK, we have files like `.pyd` or `.dylib` for `Compiled`

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
AlexanderSher: Hm, I can't find where entry is changed outside of the lock.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
AlexanderSher: Renamed to `bufferVersion`.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
MikhailArkhipov: I mean code fetches entry inside lock, then goes out. At this time theoretically entry can get replaced in the dictionary with something else. Or not. Just checking.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
AlexanderSher: Sure it can, but it doesn't solve the question how to determine which resolution service instance is the right one.

MikhailArkhipov: I guess when module is first created as a stub we can attach the stub resolution service to it so when we need to find dependencies we'll just use the service attached to the module.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
MikhailArkhipov: OK

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
AlexanderSher: Good idea, will do.

src/Parsing/Impl/Ast/FunctionDefinition.cs
```diff
@@ -185,10 +185,20 @@ public FunctionDefinition(NameExpression name, Parameter[] parameters)
...
+            if (_body != null) yield return _body;
+            if (ReturnAnnotation != null) yield return ReturnAnnotation;
+        }
+
         public override void Walk(PythonWalker walker) {
             if (walker.Walk(this)) {
                 NameExpression?.Walk(walker);
-                foreach (var p in _parameters.MaybeEnumerate()) {
+                foreach (var p in Parameters) {
                     p.Walk(walker);
```
MikhailArkhipov: `p?.Walk`?

src/Parsing/Impl/Ast/FunctionDefinition.cs
```diff
@@ -204,7 +214,7 @@ public FunctionDefinition(NameExpression name, Parameter[] parameters)
                     await NameExpression.WalkAsync(walker, cancellationToken);
                 }
 
-                foreach (var p in _parameters.MaybeEnumerate()) {
+                foreach (var p in Parameters) {
                     await p.WalkAsync(walker, cancellationToken);
```
MikhailArkhipov: `p?.Walk`?

src/Parsing/Impl/Ast/GlobalStatement.cs
```diff
@@ -18,29 +18,30 @@
...
+                    n.Walk(walker);
                 }
             }
             walker.PostWalk(this);
         }
 
         public override async Task WalkAsync(PythonWalkerAsync walker, CancellationToken cancellationToken = default) {
             if (await walker.WalkAsync(this, cancellationToken)) {
-                foreach (var n in _names.MaybeEnumerate().ExcludeDefault()) {
+                foreach (var n in Names) {
```
MikhailArkhipov: I've seen names, asNames, etc being nulls a lot

src/Parsing/Impl/Ast/ImportStatement.cs
```diff
@@ -16,18 +16,17 @@
...
 namespace Microsoft.Python.Parsing.Ast {
     public class ImportStatement : Statement {
-        private readonly ModuleName[] _names;
-        private readonly NameExpression[] _asNames;
-
-        public ImportStatement(ModuleName[] names, NameExpression[] asNames, bool forceAbsolute) {
-            _names = names;
-            _asNames = asNames;
+        public ImportStatement(ImmutableArray<ModuleName> names, ImmutableArray<NameExpression> asNames, bool forceAbsolute) {
+            Names = names;
```
MikhailArkhipov: So they are guaranteed not be nulls?

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
```diff
@@ -0,0 +1,103 @@
...
+                }
+
+                if (Version > version) {
+                    return;
+                }
+
+                Version = version;
+            }
+
+            Version = version;
```
MikhailArkhipov: Setting version twice?

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
```diff
@@ -0,0 +1,103 @@
...
+        public void TrySetException(Exception ex, int version, object syncObj) {
+            lock (syncObj) {
+                if (Version > version) {
+                    return;
+                }
+
+                Version = version;
+            }
+
+            Version = version;
```
MikhailArkhipov: As above, twice?

src/Parsing/Impl/Ast/FunctionDefinition.cs
AlexanderSher: `Parameter` can be `null`?

src/Parsing/Impl/Ast/GlobalStatement.cs
AlexanderSher: `asNames` can contain `null` for the case like `from a.b import c as cc, d, e as ee`. 
In this particular case, `GlobalStatement` is created only in `ParseGlobalStmt`, where `names` is created by `ReadNameList`, which fills array with `NameExpression` instances, produced by `MakeName`. So it can't have nulls.

src/Parsing/Impl/Ast/ImportStatement.cs
AlexanderSher: That's the idea. `ImmutableArray<T>` is a struct, so it can't be null. It can contain `null` elements. (see comment above). The only thing here is that `ImmutableArray<T>` must be created using `ImmutableArray<T>.Empty` or `ImmutableArray<T>.Create`. `default(ImmutableArray<T>)` will create uninitialized instance.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
AlexanderSher: Copy/paste issue :)

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
AlexanderSher: Yep, one time is enough.

MikhailArkhipov:
OK, approving pending the merge conflicts.

PR #648, opened 2019-02-22
Don't  squiggle new lines after incomplete member expression

### LINKED ISSUE #345: Syntax errors for unfinished expressions appear on the next line
image

I think the syntax error should be on the line where the `.` access isn't finished yet, like what you get if you remove anything following the bad expression:

image

(It's also very strange that only a subset of the comment gets squiggled instead of the whole thing, but...)

MikhailArkhipov:
Fixes #345 

If next token is a new line after dot, squiggle the dot instead.

### DIFF
--- a/src/Parsing/Impl/Parser.cs
+++ b/src/Parsing/Impl/Parser.cs
@@ -335,7 +335,12 @@ namespace Microsoft.Python.Parsing {
                 return n;
             }
 
-            ReportSyntaxError(prevTokenStart, prevTokenStart + prevTokeLength, "syntax error");
+            if (_lookahead.Token.Kind == TokenKind.NewLine) {
+                // Incomplete member expression, report dot, don't point to new line.
+                ReportSyntaxError(prevTokenStart, prevTokenStart + prevTokeLength, "syntax error");
+            } else {
+                ReportSyntaxError("syntax error");
+            }
             return Name.Empty;
         }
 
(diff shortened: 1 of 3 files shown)


src/Parsing/Impl/Parser.cs
```diff
@@ -335,7 +335,12 @@ public class Parser {
                 return n;
             }
 
-            ReportSyntaxError("syntax error");
+            if (_lookahead.Token.Kind == TokenKind.NewLine) {
+                // Incomplete member expression, report dot, don't point to new line.
+                ReportSyntaxError(prevTokenStart, prevTokenStart + prevTokeLength, "syntax error");
```
jakebailey: Previously, the behavior without a comment is to squiggle at the character right after the dot; is it possible to do that again without the squiggle showing after a comment? I think with this, the squiggle shows right on top of the dot as though it's the issue rather than missing something following.

src/Parsing/Impl/Parser.cs
```diff
@@ -2264,7 +2274,7 @@ struct Name {
             var itemWhiteSpace = MakeWhiteSpaceList();
 
             var items = ImmutableArray<WithItem>.Empty
-                .Add(ParseWithItem(itemWhiteSpace));;
+                .Add(ParseWithItem(itemWhiteSpace)); ;
```
jakebailey: There's a weird extra `;` at the end of this line. Not sure if there are anymore, but the reformat shows it in the diff so I noticed it.

PR #653, opened 2019-02-25
Remove path cache file

### LINKED ISSUE #652: Remove database search path cache file
The current implementation of this cache means that different installations/virtualenvs of Python with different locations will sometimes be treated as one during the analysis. I don't think that it provides any meaningful benefit (especially given we have to run the scraper anyway), so we should remove it and just run `get_search_paths.py` every time.

jakebailey:
Closes #652.

### DIFF
--- a/src/Analysis/Core/Impl/Interpreter/PythonLibraryPath.cs
+++ b/src/Analysis/Core/Impl/Interpreter/PythonLibraryPath.cs
@@ -99,22 +99,12 @@ namespace Microsoft.Python.Analysis.Core.Interpreter {
         /// efficiently as possible. This may involve executing the
         /// interpreter, and may cache the paths for retrieval later.
         /// </summary>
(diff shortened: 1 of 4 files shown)


src/Analysis/Core/Impl/Interpreter/PythonLibraryPath.cs
```diff
@@ -99,22 +99,12 @@ public override string ToString()
...
-                    }
-                }
 
                 try {
-                    paths = await GetUncachedDatabaseSearchPathsAsync(config.InterpreterPath);
-                    if (!string.IsNullOrEmpty(cachePath)) {
-                        WriteDatabaseSearchPaths(cachePath, paths);
-                    }
-                    return paths;
+                    return await GetUncachedDatabaseSearchPathsAsync(config.InterpreterPath);
```
AlexanderSher: We probably need another name for this method and also update comments below.

jakebailey: Renamed things and cleaned up the comments.

PR #655, opened 2019-02-25
Several bug fixes

AlexanderSher:
- Prevent several analysis of the same file to run concurrently
- Create initial analysis for every module type (not just the User)
- Add ability to add analysis dependencies found by another round of analysis


src/Core/Impl/Threading/AsyncAutoResetEvent.cs
```diff
@@ -0,0 +1,66 @@
...
+                    }
+                }
+
+                if (!_isSignaled && (waiterToRelease == default || waiterToRelease.Task.IsCompleted)) {
+                    _isSignaled = true;
+                }
+            }
+
+            if (waiterToRelease != null && !waiterToRelease.Task.IsCompleted) {
+                waiterToRelease.TrySetResult(true);
```
MikhailArkhipov: So it sets result when not completed?

AlexanderSher: Yes. Actually, there is a possibility that it may become canceled after lock but before `TrySetResult` is called. Need to fix it.

Issue #642, opened 2019-02-20
Missing numpy and tensorflow completion, members not showing up


jakebailey:
numpy and tensorflow.nn are missing completion.

```python
import numpy as np
np.
```

Nothing shows up after `np.` except some oddball stuff.

Doing this:

```python
import tensorflow
tensorflow.nn
```

`nn` has no completion, and `tensorflow.nn` itself doesn't say that `nn` exists at all.

MikhailArkhipov:
This is about submodules in folders you are working on

MikhailArkhipov:
Related: #490 

```python
from setuptools. # misses `archive*`
```
![image](https://user-images.githubusercontent.com/12820357/53657221-71e9d480-3c0a-11e9-8a0a-084cb384b2b9.png)

jakebailey:
So in looking at #690, I realized that I hadn't checked `__all__` for numpy, and behold:

https://github.com/numpy/numpy/blob/15b092f5541e80d7c3d0108957406c6f8686aba0/numpy/__init__.py#L169

```python
    __all__.extend(['__version__', 'show_config'])
    __all__.extend(core.__all__)
    __all__.extend(_mat.__all__)
    __all__.extend(lib.__all__)
    __all__.extend(['linalg', 'fft', 'random', 'ctypeslib', 'ma'])
```

Which explains why we aren't getting numpy completion anymore; we have the `__all__` filtering.

jakebailey:
Commenting out all of the code in `PythonModule.GetMemberNames` which filters:

![image](https://user-images.githubusercontent.com/5341706/53843552-aab1e280-3f57-11e9-86b5-0e903c6bee0c.png)

It's really show to show up, but it's there. Maybe we need to drop the filters for now until we can better handle `__all__`. The old LS worked because it didn't even do anything.

MikhailArkhipov:
If we drop all filters then all the `typing` and others imported will show up in the completion. But it may be smaller price to pay.

jakebailey:
The above is just:

```csharp
            // Try __all__ since it contains exported members
            var all = Analysis.GlobalScope.Variables["__all__"];
            // if (all?.Value is IPythonCollection collection) {
            //     return collection.Contents
            //         .OfType<IPythonConstant>()
            //         .Select(c => c.GetString())
            //         .Where(s => !string.IsNullOrEmpty(s));
            // }

            // __all__ is not declared. Try filtering by origin:
            // drop imported modules and generics.
[...]

PR #665, opened 2019-02-26
Fix the issue with DependencyResolver removing vertices from changed list without them being processed

AlexanderSher:


MikhailArkhipov:
https://github.com/Microsoft/python-language-server/issues/642 ?

Issue #581, opened 2019-02-06
LS: Output lint-level diagnostics


MikhailArkhipov:
Check `//TODO` related to various conditions and report diagnostics. Unresolved imports and undefined variables are implemented.

https://github.com/Microsoft/python-language-server/pull/593
https://github.com/Microsoft/python-language-server/pull/693

#395 et al

MikhailArkhipov:
Will open specific ones as needed

Issue #535, opened 2019-01-17
Additional typing specialization


MikhailArkhipov:
P2 level - Callable, Generator, IO, etc.

jakebailey:
We may want to consider throwing in some of these as plain objects in the meantime, just to have them listed in the completion (#1607).

jakebailey:
FWIW the new LS we released today has full typing support: https://devblogs.microsoft.com/python/announcing-pylance-fast-feature-rich-language-support-for-python-in-visual-studio-code/

HuskyNator:
This does not work for `Union[List[type], type, None]`, it just assumes it's `list[type]`?

jakebailey:
This project is no longer actively maintained; consider trying Pylance, which has full typing support.

judej:
Thank you for the report. This project is no longer active. If you are using the Python language server in VSCode, the language server you are using is Pylance and please the http://github.com/microsoft/pylance-release to report issues. 

If you are using Visual Studio, then, please use http://github.com/microsoft/PTVS to report issues.

thank you

PR #666, opened 2019-02-27
Fix handling of generics and class constructors that define variables

### LINKED ISSUE #337: Attribute __dict__ not available
The attribute `__dict__` is not listed in classes.

```python
class Something:
    pass
Something.__dict__
```

See above, when entering `Something.` the attr `__dict__` is not listed in the auto completion list.

This works in Jedi

### LINKED ISSUE #338: Method `mro` not listed in auto completion list
```python
class A:
	pass
A.mro()
```

The item `mro` is not listed in auto completion list.
Note: `__mro__` is listed.

This works in Jedi.

### LINKED ISSUE #462: No intellisense for azure package when creating a Python Function App
Create an Azure Function App in Python and here's the sample code that gets generated (`requirements.txt` further below):
```python
import logging
import azure.functions as func
import sys

def main(req: func.HttpRequest) -> func.HttpResponse:
    logging.info("Python HTTP trigger function processed a request.")
    name = req.params.get("name")
```

There's no intellisense for `azure` package, hence no intellisense for `req`.

I've tried pip installing the following into a separate python environment still nothing:
```
azure-functions==1.0.0a5
azure-functions-worker==1.0.0a6
grpcio==1.14.2
grpcio-tools==1.14.2
protobuf==3.6.1
six==1.11.0
```

Then checking for intellisense for `azure`, the result is nothing.
No errors, nothing, waited for 10 minutes, still nothing.

## Jedi works and its available immediately.

MikhailArkhipov:
Fixes #462
Fixes #338 
Fixes #337 

- The issue was with incorrect interpretation of arguments to generic class constructor. Instead of type the actual specific instance is being passed so we need to extract type arguments from the instance (such as dict) and create specific type based in the passed instance content and type.
- Added support for generic->specific argument match for generic functions in non-generic classes.
- Support standalone generic-returning functions (+enabled test).
- Fixed processing of constructors that define class variables + test.
- Split typing and generic tests into separate classes.


MikhailArkhipov:
Eeew PR number 666...

src/LanguageServer/Impl/Program.cs
```diff
@@ -13,7 +13,7 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
-// #define WAIT_FOR_DEBUGGER
+#define WAIT_FOR_DEBUGGER
```
jakebailey: Don't forget to remove this before merging.

jakebailey: Agh, beat me by 10 seconds.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
```diff
@@ -26,97 +27,72 @@ namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
...
-                        // The expression is still generic. For example, generic return
-                        // annotation of a class method, such as 'def func(self) -> A[_E]: ...'.
-                        // Leave it alone, we don't want resolve generic with generic.
-                        return null;
-                    }
 
-                    if (genericTypeArgs.Length > 0 && genericTypeArgs.Length != indices.Count) {
-                        // TODO: report that some type arguments are not declared with TypeVar.
-                    }
+            using (OpenScope(target.GetPythonType()?.DeclaringModule, GetScope(target), out _)) {
```
AlexanderSher: `OpenScope` changes `Scope CurrentScope { get; private set; }`. Should it be defined as `AsyncLocal<Scope>`?

src/Analysis/Ast/Impl/Types/PythonFunctionOverload.cs
```diff
@@ -166,23 +167,38 @@ internal void SetReturnValueProvider(ReturnValueProvider provider)
...
-                        // Match arguments 
-                        var typeArgs = selfClassType.GenericParameters.Keys
+                case PythonClassType cls when cls.IsGeneric():
+                    // -> A[_T1, _T2, ...]
+                    // Match arguments
+                    IReadOnlyList<IPythonType> typeArgs = null;
+                    var classGenericParameters = selfClassType.GenericParameters.Keys.ToArray();
+                    if (classGenericParameters.Length > 0) {
+                        // Declaring class is specific and provides definitions of generic parameters
+                        typeArgs = classGenericParameters
```
AlexanderSher: What if there is a mix of class generic parameters and method generic parameters?

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
MikhailArkhipov: During analysis it is not a problem since there is no concurrency anymore. But there is indeed a bit of a problem with it indeed when several requests for tooltips may come interleaved. I've seen AF couple of times when trying to close scope that was not open. AF is disabled for now since nothing bad happens really. I could make it `AsyncLocal` if that would help, but I am not entirely sure.

src/LanguageServer/Impl/Sources/DefinitionSource.cs
```diff
@@ -54,6 +55,29 @@ internal sealed class DefinitionSource {
...
 
                 var value = eval.GetValueFromExpression(expr);
+                if (value.IsUnknown()) {
+                    // If this is 'import A as B' A is not declared as a variable, so try modules.
+                    string moduleName = null;
+                    if (!string.IsNullOrEmpty(name)) {
+                        switch (statement) {
+                            case ImportStatement imp when imp.Names.Any(x => x?.MakeString() == name):
+                            case FromImportStatement fimp when fimp.Names.Any(x => x?.Name == name):
+                                moduleName = name;
```
AlexanderSher: We should use `PathResolverSnapshot` here, otherwise it won't work for any complex cases.

src/LanguageServer/Impl/Sources/DefinitionSource.cs
```diff
@@ -38,13 +39,13 @@ internal sealed class DefinitionSource {
             }
 
             var eval = analysis.ExpressionEvaluator;
-            using (eval.OpenScope(analysis.Document, exprScope)) {
+            var name = (expr as NameExpression)?.Name;
```
AlexanderSher: Can we find anything at all if `expr` is not a `NameExpression`?

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
AlexanderSher: Yeah, if there are no concurrent requests, `AsyncLocal` won't help.

src/Analysis/Ast/Impl/Types/PythonFunctionOverload.cs
MikhailArkhipov: Its OK. Specific class stores dictionary of the original generic parameter names (such as K, V in Dict) and their assigned specific types (int, str) if Dict[K, V] was created via 

```python
from typing import Dict

d1 = {1:'a', 2:'b'}
d2 = Dict(d1))
```
So here we are trying to match names of the generic function parameters to those of the class so we can figure out which specific type to use. Generally speaking method generic names must either be different or they match the class. Per Python `T` is basically a variable with the assigned type declared in the class scope. So function that is generic of T will be using class T.

src/LanguageServer/Impl/Sources/DefinitionSource.cs
MikhailArkhipov: Yep, it doesn't work in more complex cases like `from a.b.c import ...` - `a.b.c` is not handled completely correctly (right click on `b` -> goto definition does not work`. Probably should file a separate issue to improve this.

src/LanguageServer/Impl/Sources/DefinitionSource.cs
MikhailArkhipov: Member expression also works.

src/Analysis/Ast/Impl/Types/PythonFunctionOverload.cs
AlexanderSher: I've been thinking of a case when function has a different generic parameter. Something like this:
```python
from typing import TypeVar, Generic, Callable

T = TypeVar('T')
V = TypeVar('V')

class C(Generic[T]):
    def __init__(self, value: T) -> None:
        self.value = value

    def f(self, input: V, callback: Callable[[T, V], str]) -> str: 
        return callback(self.value, input)

def g(s1: int, s2: str) -> str:
    return str(s1) + s2

print(C(1).f("b", g))
```

src/LanguageServer/Impl/Sources/DefinitionSource.cs
AlexanderSher: Ok, let's have a separate task for it. There were also couple of todos in the files. Do you want tasks for them as well?

MikhailArkhipov: Depends on TODO. There is a bunch of placeholders to report future diagnostics, there is already one task for them - #581. Others may or may not common, so case by case.

src/Analysis/Ast/Impl/Types/PythonFunctionOverload.cs
MikhailArkhipov: In this particular case it doesn't matter since annotation says `str` and hence

image

Now, if we are to eval this (we don't support `Callable` just yet though, that's #535)... it is still a string since eval with call `callback` and either use annotation or eval with arguments and yield a string. 

There are probably some convoluted examples, but I'd wait for users to give them to us :-)

AlexanderSher: Sorry, I meant `def f(self, input: V, callback: Callable[[T, V], V]) -> V` in the example above. But since we have separate task for it, that's not for this PR.

src/Analysis/Ast/Impl/Values/Scope.cs
```diff
@@ -42,7 +42,7 @@ internal class Scope : IScope {
         public IScope OuterScope { get; }
         public bool VisibleToChildren { get; }
 
-        public IReadOnlyList<IScope> Children => _childScopes ?? Array.Empty<IScope>() as IReadOnlyList<IScope>;
+        public IReadOnlyList<IScope> Children => _childScopes?.ToArray() ?? Array.Empty<IScope>();
```
AlexanderSher: `_childScopes` looks like a good candidate for `ImmutableArray`, cause the child scope is never removed. And it can be safely returned from the `Children` property as is. But it's up to you.

PR #667, opened 2019-02-27
Reload modules on pip install

### LINKED ISSUE #611: Restore reloading of modules on pip install

MikhailArkhipov:
Fixes #611


src/LanguageServer/Impl/LanguageServer.cs
```diff
@@ -377,19 +380,29 @@ public Task ExtensionCommand(JToken token, CancellationToken cancellationToken)
...
                 var interpreter = _services.GetService<IPythonInterpreter>();
+                var logger = _services.GetService<ILogger>();
                 _pathsWatcher = new PathsWatcher(
                     _initParams.initializationOptions.searchPaths,
-                    () => interpreter.ModuleResolution.ReloadAsync(cancellationToken).DoNotWait(),
-                    _services.GetService<ILogger>()
-                 );
+                    () => {
+                        logger.Log(TraceEventType.Information, Resources.ReloadingModules);
+                        interpreter.ModuleResolution.ReloadAsync(cancellationToken).ContinueWith(t => {
```
AlexanderSher: Should we clean the `ModuleResolutionBase._modules` when we do `ReloadAsync`?

src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -155,7 +155,6 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
                 return collection.Contents
                     .OfType<IPythonConstant>()
                     .Select(c => c.GetString())
-                    .ExcludeDefault()
```
AlexanderSher: Why?

MikhailArkhipov: Next condition is !string.isNullOfEmpty

src/LanguageServer/Impl/LanguageServer.cs
MikhailArkhipov: `ReloadAsync` technically should be doing it, let me check

PR #673, opened 2019-02-27
Сapture Python output reliably

### LINKED ISSUE #663: ProcessHelper occasionally misses data and GetSearchPathsAsync fails
This is because next event can come while previous is still being processed and hence sometimes Dispose() is called before all events were handled.

https://github.com/Microsoft/python-language-server/pull/653 made this much more visible.

MikhailArkhipov:
Fixes #663 

The issue was that events from the process come in multiple threads and next event can come while previous is still being handled (and get handled out of order occasionally) as well as process can exit before all events are actually sent.

- Ported process classes from RTVS.


AlexanderSher:
This whole thing looks like a huge overkill to me. In RTVS, we had this method to execute commands in docker on all platforms: https://github.com/Microsoft/RTVS/blob/master/src/Containers/Impl/Docker/LocalDockerService.cs#L174
I don't see much difference between this one and executing python script.

src/Core/Impl/OS/ProcessServices.cs
```diff
@@ -0,0 +1,54 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Python.Core.OS {
+    public sealed class ProcessServices : IProcessServices {
```
AlexanderSher: Do we need to add this service into tests?

MikhailArkhipov: Yes I added it

src/Analysis/Core/Impl/Interpreter/PythonLibraryPath.cs
```diff
@@ -122,59 +122,53 @@ public override string ToString()
...
-                        } catch (OperationCanceledException) {
-                            proc.Kill();
-                            exitCode = -1;
-                        }
-                        if (exitCode != 0) {
-                            throw new InvalidOperationException(string.Join(Environment.NewLine, errorLines));
-                        }
+                var lines = await ps.ExecuteAndCaptureOutputAsync(startInfo, cancellationToken);
+                return lines.Select(s => {
+                    if (s.StartsWithOrdinal(tempWorkingDir, ignoreCase: true)) {
```
AlexanderSher: `s.PathStartsWith`

src/Analysis/Ast/Impl/Modules/CompiledPythonModule.cs
```diff
@@ -73,34 +75,29 @@ public override string Documentation
...
+                RedirectStandardOutput = true,
+                RedirectStandardError = true
+            };
+            var ps = Services.GetService<IProcessServices>();
+
+            Log?.Log(TraceEventType.Verbose, "Scrape", startInfo.FileName, startInfo.Arguments);
+            Task<IReadOnlyList<string>> task = null;
+
+            try {
+                task = ps.ExecuteAndCaptureOutputAsync(startInfo);
```
AlexanderSher: IMO, it is better to pass timeout as parameter to `ExecuteAndCaptureOutputAsync`:
```cs
try {
    var token = new CancellationTokenSource(60000).Token;
    var lines = ps.ExecuteAndCaptureOutputAsync(startInfo, token).GetAwaiter().GetResult();
    return string.Join(Environment.NewLine, lines);
} catch (Exception) {
    return string.Empty;
}
```

MikhailArkhipov: I can create cancellation token with timeout instead

MikhailArkhipov: Done

PR #676, opened 2019-02-28
Improve handling of Union index and call

MikhailArkhipov:
Make sure we delegate index/call to types until we get something meaningful.


src/Analysis/Ast/Impl/Types/PythonUnionType.cs
```diff
@@ -65,10 +65,26 @@ internal sealed class PythonUnionType : IPythonUnionType {
...
 
-        public IMember Call(IPythonInstance instance, string memberName, IArgumentSet args)
-            => DeclaringModule.Interpreter.UnknownType;
-        public IMember Index(IPythonInstance instance, object index)
-            => DeclaringModule.Interpreter.UnknownType;
+        public IMember Call(IPythonInstance instance, string memberName, IArgumentSet args) {
+            lock (_lock) {
+                // Check if any types support calls
+                var result = _types
+                    .Select(t => t.Call(instance, memberName, args))
```
AlexanderSher: A bit scary to call `Call` under the lock. Can we move it out?

MikhailArkhipov: Sure although it is a simulated call in the same analysis.

PR #677, opened 2019-02-28
Fixes occasional 'collection modified' exception

### LINKED ISSUE #674: Occasional 'collection modified' exception when opening scope in another module
When running library analysis tests under debugger I occasionally get `collection modified` exception in `OpenScope` that traverses scope tree breadth first. Happens when code wants to evaluate function with arguments and function is located in another module.

MikhailArkhipov:
Fixes #674

Root cause is that stub function has different `FunctionDefinition` from the original one so search for the scope fails because stub AST != primary AST and then `OpenScope` attempts to create the scope hence modifying the collection it is enumerating.

The fix is to avoid evaluating stubs with arguments. It does not make sense to eval them with specific arguments anyway since their return type is known from the annotation. Evaluating with arguments is done as a last resort when the analysis could not figure out static/consistent return type as in

```python
def foo(x):
    return x
```


src/Analysis/Ast/Impl/Extensions/NodeExtensions.cs
```diff
@@ -54,5 +54,12 @@ public static class NodeExtensions {
             }
             return LocationInfo.Empty;
         }
+
+        public static bool IsInAst(this ScopeStatement node, PythonAst ast) {
```
AlexanderSher: Should this return `true` if `node` is `ast` itself?

MikhailArkhipov: Interesting. Probably yes.

MikhailArkhipov: In fact it is does `ast.Parent` is null so loop is bypassed and `ast == node` will do the right thing.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Scopes.cs
```diff
@@ -77,7 +77,7 @@ public void DeclareVariable(string name, IMember value, VariableSource source, N
 
             scope = scopes.FirstOrDefault(s => s.Variables.Contains(name));
             var value = scope?.Variables[name].Value;
-            if (value == null) {
+            if (value == null && (options & LookupOptions.Builtins) == LookupOptions.Builtins) {
```
AlexanderSher: You can use `options.HasFlag`

PR #680, opened 2019-02-28
Treat default value None as nothing

### LINKED ISSUE #604: Treat default parameter values of None as being unspecified/Optional
Take:

```python
def f(x=None):
    return x

f("4")
```

We currently see `None`, then say that `x` is of type `NoneType`, when the user probably actually meant `Optional[_T]` or similar. It'd be better to treat the special case of `None` as though there were no default specified, for example like:

```python
def g(x):
    return x

g("4")
```

image

MikhailArkhipov:
Fixes #604

Treat default value None as nothing since it is not a parameter type but just says the parameter may be missing.

### DIFF
--- a/src/Analysis/Ast/Impl/Extensions/PythonTypeExtensions.cs
+++ b/src/Analysis/Ast/Impl/Extensions/PythonTypeExtensions.cs
@@ -28,18 +28,14 @@ namespace Microsoft.Python.Analysis {
         public static bool IsGeneric(this IPythonType value)
             => value is IGenericTypeDefinition || value is IGenericType || (value is IPythonClassType c && c.IsGeneric());
 
-        public static void TransferDocumentationAndLocation(this IPythonType src, IPythonType dst) {
-            if (src != null && dst is PythonType pt) {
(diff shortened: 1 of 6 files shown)


src/Analysis/Ast/Impl/Extensions/PythonTypeExtensions.cs
```diff
@@ -28,18 +28,14 @@ public static bool IsGenericParameter(this IPythonType value)
         public static bool IsGeneric(this IPythonType value)
             => value is IGenericTypeDefinition || value is IGenericType || (value is IPythonClassType c && c.IsGeneric());
 
-        public static void TransferDocumentationAndLocation(this IPythonType src, IPythonType dst) {
-            if (src != null && dst is PythonType pt) {
-                pt.TrySetTypeId(dst.TypeId);
+        public static void TransferDocumentationAndLocation(this IPythonType s, IPythonType d) {
+            if (s != null && s != d && s is PythonType src && d is PythonType dst) {
```
AlexanderSher: I think we don't need `s != null` check here. Pattern matching will fail for `s == null`

PR #681, opened 2019-02-28
Fix EndIndex in parser

huguesv:
Makes the EndIndex of a suite statement use the EndIndex of the last statement, rather than use the current token location, which may have advanced further, past some new line chars.

Example: In some scenarios, the EndIndex of a function ended up the same as the StartIndex of the next function.

That was a regression introduced in https://github.com/Microsoft/PTVS/commit/33bf79ebb758f4592f683b44587bd3e096cabae2#diff-90ae8164360db2dc89b75eeace0edd36L3868

This fixes outlining https://github.com/Microsoft/PTVS/issues/4703 and test discovery https://github.com/Microsoft/PTVS/issues/4908

### DIFF
--- a/src/Analysis/Engine/Test/ParserTests.cs
+++ b/src/Analysis/Engine/Test/ParserTests.cs
@@ -3025,6 +3025,34 @@ pass
             );
         }
 
+        [TestMethod, Priority(0)]
+        public void FunctionLocations() {
+            var ast = ParseFileNoErrors("FunctionLocations.py", PythonLanguageVersion.V36);
+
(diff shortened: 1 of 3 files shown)


src/Analysis/Engine/Test/ParserTests.cs
```diff
@@ -3025,6 +3025,34 @@ public class ParserTests {
             );
         }
 
+        [TestMethod, Priority(0)]
+        public void FunctionLocations() {
+            var ast = ParseFileNoErrors("FunctionLocations.py", PythonLanguageVersion.V36);
```
MikhailArkhipov: Please use general 3+ since not everyone has 3.6, most are on 3.7

PR #684, opened 2019-03-01
Wire up restart of analysis on changes in configuration

MikhailArkhipov:
Also upgrade dependencies.


src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -154,17 +154,20 @@ public sealed partial class Server : IDisposable {
...
             _disposableBag.ThrowIfDisposed();
-
-            var reanalyze = true;
-            if (@params.settings != null) {
-                if (@params.settings is ServerSettings settings) {
-                    reanalyze = HandleConfigurationChanges(settings);
-                } else {
-                    _log?.Log(TraceEventType.Error, "change configuration notification sent unsupported settings");
+            switch (@params.settings) {
+                case null:
```
AlexanderSher: No logging for `null`?

MikhailArkhipov: Null means it is missing so it is not `unsupported`

Issue #647, opened 2019-02-21
Move strings in parser to resources


MikhailArkhipov:
Currently all hardcoded ENU strings

jamesralstin:
Hello @MikhailArkhipov, I am new to Open Source Software Development (currently taking a class on it), I am trying to get involved and make my first contribution, do you think this would be a good first issue for me to, given I am not that familiar with the project? Thank you for any and all advice!

jamesralstin:
Also, I have forked the project and cloned it locally, I think i found the strings referenced in this issue.

I noticed a number of hard coded strings in C# Library Microsoft.Python.Parsing.Parser.cs, is the hope to move these strings to the Resources.resx file in the same project?

MikhailArkhipov:
Sure, take it. Yes, hardcoded string relevant to user should be in Resources. Not all string, just messages and anything that should be translated. Node names, for example, do not need to move, they are for internal use. But syntax error messages - yes.

MikhailArkhipov:
While you are at it, you can also look at #228 at least in the parser. I.e. change
```cs
throw new ArgumentNullException("text");
```
to
```
throw new ArgumentNullException(nameof(text));
```

jamesralstin:
Thanks for the quick response @MikhailArkhipov, I will get right on this!

jamesralstin:
@MikhailArkhipov I have made these changes, pushed them to my branch, and ran the 'Microsoft.Pythong.Parsing.Tests' unit tests, everything passes except for the 'Microsoft.Pythong.Parsing.Tests.ParserTests.Errors' test. I can't for the life of me figure out why the error is returning an empty string when testing this:

The error message is

"Message: Wrong msg for error 38: expected "new ErrorResult("invalid syntax, class decorators require 2.6 or later.", new SourceSpan(95, 1, 95, 6))", got "new ErrorResult("", new SourceSpan(95, 1, 95, 6))""

The error is coming from the "public void Errors()" method in ParserTests.cs, it seems like it has something to do with line 520:

`new ErrorResult("invalid syntax, class decorators require 2.6 or later.", new SourceSpan(93, 1, 93, 6)),`

In the Parser.cs file, I replaced line 1808 with:

`res = ErrorStmt(Resources.ClassDecoratorsRequireTwodotSixErrorMsg, decorators, cls);`

It was:

'res = ErrorStmt("invalid syntax, class decorators require 2.6 or later.", decorators, cls);'

[...]

PR #685, opened 2019-03-01
Parsing f-strings

### LINKED ISSUE #6: Support completion in { } in f-strings
https://github.com/Microsoft/PTVS/issues/4635
https://github.com/Microsoft/vscode-python/issues/2379

https://user-images.githubusercontent.com/12820357/44060308-1f9d37ce-9f09-11e8-9f0a-62d4a4cf26b8.png

### LINKED ISSUE #221: Add new token to distinguish fstrings
For now, this can just be a wrapper around `ConstantValueToken` that does nothing interesting. Requires that the tokenizer pass `isFormatted` farther than it currently does. Add new tests which check for the new token, and verify that the existing tokenizer/parser tests continue to work the same (for now).

Epic #6.

### LINKED ISSUE #222: Implement fstring literal tokenizer to break string into text/interpolated regions
For example, turning `f'My name is {name:r}: {{hi}}'` into:

```python
[
    FStringText("f'My name is "),
    FStringInterpolation("name", "r"),
    FStringText(": {{hi}}",
]
```

Or similar. This precedes actually tokenizing/parsing the interpolations themselves.

Epic #6.

### LINKED ISSUE #223: Tokenize/parse fstring interpolations
Using the scanned items from #222, tokenize each interpolation and produce an AST.

Example of tokenizing/parsing a string: https://github.com/Microsoft/python-language-server/blob/master/src/Analysis/Engine/Impl/ExpressionFinder.cs

Parsing needs to start at the expression level (https://docs.python.org/3/reference/lexical_analysis.html#f-strings), so likely also needs to call a function similar to `ParseExprStmt` which disallows things like assignment.

Epic #6.

### LINKED ISSUE #224: Add fstring AST nodes
There should be an fstring node which holds the scanned result from #222 and properly allows for walking. There may need to be an extra node for the text portion of the fstring (as any interpolations will be nodes of a type we don't care about).

Epic #6.

brianbok:
f-strings were treated as normal strings before. Now we parse their content, including f-string expressions inside.
We provide completion and hovering for sub-expressions.
Closes #221.
Closes #222.
Closes #223.
Closes #224.
Closes #6.


src/Parsing/Impl/Ast/FStringExpression.cs
```diff
@@ -0,0 +1,36 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Python.Parsing.Ast {
+    public class FStringExpression : Expression {
+        public List<Expression> Children { get; }
+
+        public FStringExpression(List<Expression> children) {
```
MikhailArkhipov: Input should be `IReadOnlyList` normally. As is public property. Allowing public API to modify AST is dangerous.

src/Parsing/Impl/Ast/FStringExpression.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Python.Parsing.Ast {
+    public class FStringExpression : Expression {
+        public List<Expression> Children { get; }
```
MikhailArkhipov: Can we reuse `IEnumerable<Node> GetChildNodes()` ? Having two similar methods is confusing.

src/Parsing/Impl/Ast/FSubExpressionWithOptions.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Microsoft.Python.Parsing.Ast {
+    public class FSubExpressionWithOptions : Expression {
+        private Expression _expr;
```
MikhailArkhipov: `readonly`

src/Parsing/Impl/Ast/FSubExpressionWithOptions.cs
```diff
@@ -0,0 +1,40 @@
...
+        private Expression _conversionExpression;
+
+        public FSubExpressionWithOptions(Expression expr, Expression formatExpression, Expression conversionExpression) {
+            _expr = expr;
+            _formatExpression = formatExpression;
+            _conversionExpression = conversionExpression;
+        }
+
+        public override IEnumerable<Node> GetChildNodes() {
+            return new List<Node>() { _expr, _formatExpression, _conversionExpression };
```
MikhailArkhipov: `new []` is more frugal memory-wise.

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,147 @@
...
+namespace Microsoft.Python.Parsing {
+    public class FStringParser {
+        private readonly string _fString;
+        private readonly StringBuilder _buffer = new StringBuilder();
+        private readonly FStringBuilder _builder;
+        private readonly ErrorSink _errors;
+        private int _errorCode = 0;
+        private int _position = 0;
+
+        public FStringParser(FStringBuilder builder, string fString, ErrorSink errors) {
```
MikhailArkhipov: Is `FStringBuilder` ever used outside of `FStringParser`? If not, `FStringParser` can just `new` its own builder

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,147 @@
...
+        public void Parse() {
+            while (!EndOfFString()) {
+                if (IsDoubleBrace()) {
+                    NextChar();
+                    _buffer.Append(NextChar());
+                } else if (CurrentChar() == '{') {
+                    AddBufferedSubstring();
+                    ParseInnerExpression();
+                } else if (CurrentChar() == '}') {
+                    throw new Exception("closing '}' without opening");
```
MikhailArkhipov: Nope, parser never throws. Should report the error, recover and continue.

MikhailArkhipov: What about `}}` matching `{{` ?

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,147 @@
...
+                } else {
+                    _buffer.Append(NextChar());
+                }
+            }
+            AddBufferedSubstring();
+        }
+
+        private bool IsDoubleBrace() {
+            StringSpan doubleOpen = new StringSpan("{{", 0, 2);
```
MikhailArkhipov: Should be static/const to reduce amount of garbage in memory. No need to create objects for every comparo.

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,147 @@
...
+            StringSpan doubleClose = new StringSpan("}}", 0, 2);
+            return IsNext(doubleOpen) || IsNext(doubleClose);
+        }
+
+        private bool IsNext(StringSpan span) {
+            return _fString.Slice(_position, span.Length).Equals(span);
```
MikhailArkhipov: `=>` syntax

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        private void ParseInnerExpression() {
+            Check.InvalidOperation(_buffer.Length == 0, "Current buffer is not empty");
+
```
MikhailArkhipov: Parser should not be throwing, only reporting errors unless this is internal bug.

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,147 @@
...
+            Read('{');
+            AppendAllSubExpression();
+            Read('}');
```
MikhailArkhipov: Can `}` be missing like when expression is not closed?

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,147 @@
...
+            if (!subExprStr.IsNullOrEmpty()) {
+                _builder.AppendExpression(CreateExpression(_buffer.ToString()));
+            }
+            _buffer.Clear();
+        }
+
+        private Expression CreateExpression(string subExprStr) {
+            Check.ArgumentNotNullOrEmpty(nameof(subExprStr), subExprStr);
+
+            var parser = Parser.CreateParser(new StringReader(subExprStr), PythonLanguageVersion.V36);
```
MikhailArkhipov: Hardcoded language version

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,147 @@
...
+            var expr = parser.ParseFStrSubExpr(out var formatExpression, out var conversionExpression);
+            if (expr is null) {
+                ReportSyntaxError("Subexpression failed to parse");
```
MikhailArkhipov: `ParseFStrSubExpr` should be probably reporting something more detailed

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,147 @@
...
+            } else {
+                return expr;
+            }
+        }
+
+        private void AppendAllSubExpression() {
+            int openedBraces = 1;
+            while (!(CurrentChar() == '}' && openedBraces == 1)) {
+                if (EndOfFString()) {
+                    throw new Exception("Inner expression without closing '}'");
```
MikhailArkhipov: Report error and recover

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,147 @@
...
+                _builder.AppendExpression(CreateExpression(_buffer.ToString()));
+            }
+            _buffer.Clear();
+        }
+
+        private Expression CreateExpression(string subExprStr) {
+            Check.ArgumentNotNullOrEmpty(nameof(subExprStr), subExprStr);
+
+            var parser = Parser.CreateParser(new StringReader(subExprStr), PythonLanguageVersion.V36);
+            var expr = parser.ParseFStrSubExpr(out var formatExpression, out var conversionExpression);
```
MikhailArkhipov: It should probably be getting error sink to report issues

src/Parsing/Impl/FStringParser.cs
MikhailArkhipov: Speaking of, if we discover f-string in Python 2 we should report that the syntax requires Python 3 since parser does this for other new constructs

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,147 @@
...
+            if (_buffer.Length == 0) {
+                return;
+            }
+            var s = _buffer.ToString();
+            _builder.AppendString(s);
+            _buffer.Clear();
+        }
+
+        private char NextChar() {
+            var peek = CurrentChar();
```
MikhailArkhipov: Should probably check EOF and not advance every time

src/Parsing/Test/FStringTests.cs
```diff
@@ -0,0 +1,368 @@
...
+            ";
+            var parser = Parser.CreateParser(MakeReader(code), PythonLanguageVersion.V36);
+            var ast = parser.ParseFile();
+            ast.Walk(new MyPythonWalker(expr => {
+                switch (expr) {
+                    case FStringExpression outerFstring:
+                        if (outerFstring.GetChildNodes().First() is FStringExpression innerFString) {
+                            if (outerFstring.GetChildNodes().First() is ConstantExpression constExpr) {
+                                constExpr.Value.Should().BeEquivalentTo("f\"111\"");
+                            }
```
MikhailArkhipov: The test seems to only verify when `case FStringExpression outerFstring`. If condition is false, the test will pass which doesn't seem right.

MikhailArkhipov: The best for parser testing is to have some sort of AST persistence to disk so output of the test can be compared to a predefined baseline. Ex https://github.com/Microsoft/RTVS/blob/master/src/R/Core/Test/Parser/ParseBuiltInsTest.cs

src/Parsing/Test/FStringTests.cs
```diff
@@ -0,0 +1,368 @@
...
+
+        [TestMethod, Priority(0)]
+        public void FStringInternalSubString() {
+            var parser = Parser.CreateParser(MakeReader("f'bla'"), PythonLanguageVersion.V36);
+            var ast = parser.ParseFile();
+            ast.Walk(new MyPythonWalker(expr => {
+                switch (expr) {
+                    case ConstantExpression constExpr:
+                        if (!constExpr.Value.Equals("bla")) {
+                            throw new InternalTestFailureException("Internal const expr didn't match substring");
```
MikhailArkhipov: Rather than throwing this should be either `Assert` or `Should().*`

src/Parsing/Test/FStringTests.cs
```diff
@@ -0,0 +1,368 @@
...
+            _t(node);
+            return true;
+        }
+
+        public override bool Walk(NameExpression node) {
+            _t(node);
+            return true;
+        }
+    }
+}
```
MikhailArkhipov: I'd add some tests for longer constructs like function calls. I.e. `f"output:{func(a, b, c) + 1 + func(1, 2, 3) + C().x}" with some legal as well as not legal code fragments (i.e. try defining new class or function or import inside `{ }`

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,213 @@
...
+
+                if (ch == '#') {
+                    /* Error: can't include a comment character, inside parens
+                       or not. */
+                    ReportSyntaxError("f-string expression part cannot include '#'");
+                    errorReported = true;
+                } else if (ch == ')' || ch == '}') {
+                    _buffer.Append(ch);
+                    if (nestedParens.Pop() != ch) {
+                        ReportSyntaxError("");
```
MikhailArkhipov: empty string?

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,213 @@
...
+                    _buffer.Append(ch);
+                } else {
+                    _buffer.Append(ch);
+                }
+
+                NextChar();
+            }
+            if (EndOfFString()) {
+                if (nestedParens.Count > 0) {
+                    ReportSyntaxError("");
```
MikhailArkhipov: As above

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,213 @@
...
+                //return ErrorExpression();
+                return null;
+            }
+            var conversion = ' ';
+            Expression formatSpecifier = null;
+            if (CurrentChar() == '!') {
+                Read('!');
+                conversion = NextChar();
+                if (!(conversion == 's' || conversion == 'r' || conversion == 'a')) {
+                    ReportSyntaxError($"f-string: invalid conversion character: {conversion} expected 's', 'r', or 'a'");
```
MikhailArkhipov: Since we have https://github.com/Microsoft/python-language-server/issues/647 can you please relocate to resources strings at least in the new code.

jakebailey:
You're also going to need a fix similar to #752 to prevent completion from happening in the literal part of the fstring, i.e., not have this:

image

src/Parsing/Impl/FStringParser.cs
```diff
@@ -0,0 +1,448 @@
...
+            } else if ((ch == ')' || ch == '}') && _nestedParens.Count == 0) {
+                ReportSyntaxError(Resources.UnmatchedFStringErrorMsg.FormatInvariant(ch));
+            } else if (ch == '(' || ch == '{') {
+                _nestedParens.Push(ch);
+            }
+
+            _buffer.Append(NextChar());
+        }
+
+        private bool IsOpeningOf(char opening, char ch)
```
jakebailey: A plain switch would probably be better than the dictionary at the top.

```csharp
switch (opening) {
case '(' when ch == ')':
case '{' when ch == '}':
	return true;
default:
	return false
}
```

PR #687, opened 2019-03-01
Basic analysis progress reporting

### LINKED ISSUE #688: Add basic progress reporting to analysis

MikhailArkhipov:
Fixes #688

- Numbers can go up/down. We can nix numbers and just leave the spinner if that is weird.
- Special class that reduces frequency of reporting traffic and auto-disposes the message when there is no activity.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Definitions/IProgressReporter.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Definitions/IProgressReporter.cs
@@ -15,6 +15,6 @@
 
 namespace Microsoft.Python.Analysis.Analyzer {
     internal interface IProgressReporter {
-        void ReportRemaining(int count);
+        void ReportRemaining(int count = -1);
     }
 }
--- a/src/Analysis/Ast/Impl/Analyzer/Definitions/IProgressReporter.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Definitions/IProgressReporter.cs
@@ -1,0 +1,20 @@
+﻿// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+namespace Microsoft.Python.Analysis.Analyzer {
+    internal interface IProgressReporter {
+        void ReportRemaining(int count);
+    }
+}
--- a/src/Analysis/Ast/Impl/Analyzer/ProgressReporter.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/ProgressReporter.cs
@@ -1,0 +1,89 @@
+﻿// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Threading;
+using Microsoft.Python.Core;
+using Microsoft.Python.Core.Services;
+
+namespace Microsoft.Python.Analysis.Analyzer {
+    internal sealed class ProgressReporter : IProgressReporter, IDisposable {
+        private const int _initialDelay = 100;
+        private const int _reportingInterval = 300;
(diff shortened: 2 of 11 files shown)


src/Analysis/Ast/Impl/Analyzer/Definitions/IProgressReporter.cs
```diff
@@ -0,0 +1,20 @@
...
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+namespace Microsoft.Python.Analysis.Analyzer {
+    internal interface IProgressReporter {
+        void ReportRemaining(int count);
```
AlexanderSher: Maybe it's better to have a message here?

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -154,8 +158,8 @@ bool IsSignificant(IPythonModule m)
 
             using (var cts = CancellationTokenSource.CreateLinkedTokenSource(_disposeToken.CancellationToken, cancellationToken)) {
                 var analysisToken = cts.Token;
- 
-                var walker = await _dependencyResolver.AddChangesAsync(key, entry, cts.Token);
+
+                var walker = await _dependencyResolver.AddChangesAsync(key, entry, cts.Token, _progress);
```
AlexanderSher: `_dependencyResolver` is just a graph, it doesn't do any `progress` itself, so no reason to pass `IProgressService` to it. Instead, you can report the `walker.AffectedValues` at the beginning of the try block below.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -187,10 +191,13 @@ bool IsSignificant(IPythonModule m)
                     }
 
                     _log?.Log(TraceEventType.Verbose, $"Analysis version {walker.Version} of {walker.AffectedValues.Count} entries has started.");
+                    _progress.ReportRemaining(walker.Remaining);
```
AlexanderSher: You'd like to do this inside the loop in the `AnalyzeAffectedEntriesAsync`

src/Analysis/Ast/Impl/Analyzer/Definitions/IProgressReporter.cs
MikhailArkhipov: I tried to simplify and get code just report progress which is effectively count. Or 'X of Y' ready.

MikhailArkhipov:
Tweaked a bit. There is actually a bug in the extension, it missed `beginProgress` often b/c of sync issues. I will submit PR there separately.

MikhailArkhipov:
Extension fixed in https://github.com/Microsoft/vscode-python/pull/4620

src/Analysis/Ast/Impl/Analyzer/ProgressReporter.cs
```diff
@@ -0,0 +1,95 @@
...
+                    _lastReportedToClient = _lastReportedCount;
+                    _progress = _progress ?? _progressService.BeginProgress();
+                    if (_lastReportedCount > 0) {
+                        _progress?.Report(Resources.AnalysisProgress.FormatUI(_lastReportedCount)).DoNotWait();
+                    }
+                }
+            }
+        }
+
+        private void OnDisposeTimer(object o) {
```
jakebailey: For me, if the number doesn't change for a second or so, the message disappears, and then never returns again even if the analysis continues. I think it's because this function gets called and everything gets disposed early.

MikhailArkhipov: So you have Microsoft/vscode-python#4620? You'd need to get new code and F5 from the extension.

MikhailArkhipov: Also, the message does can get removed if update comes later than the dispose interval. This is normal at least for now.

jakebailey: It's pretty evident when you're working on a big repo like scikit-learn (`k_means_.py`, as usual).

MikhailArkhipov: There is no known way to learn that there is more to come. Not yet anyway.

jakebailey: I don't understand that statement; If the LS reports 500 things left to analyze, then pauses for 5 seconds, the text disappears. Then it sees that there are say 480 things left to analyze, but that message doesn't seem to be emitted. I'm not saying that it should know the future, but I wouldn't expect that the timeout would prevent all future indications to be thrown away or never rerun.

MikhailArkhipov: It does not dispose the report forever, it hides the progress. Next message will re-create the progress.

jakebailey: That's what I'm saying doesn't always happen. Screencap:



Note how it disappears sometimes, then disappears completely.

Though, modifying this file (which is imported in many other places) does cause it to show briefly. Maybe it's just this initial analysis that does it, I just thought it was strange for it to be gone altogether when there are so many things left to analyze and it's logging progress.

MikhailArkhipov: If it reports *complete*, the analysis is done and there is nothing to report.

jakebailey: I understand, but I can see real CPU usage (even without logging) after the message has disappeared. I don't believe the analysis is complete when the message goes away at 1185 items remaining in my screencap above.

MikhailArkhipov: Point is - there is no reliable progress reporting in analysis graph at the moment. Showing nothing is worse than at least giving some indication of work getting done. Not disposing of the message does not necessarily mean it will be showing something meaningful, most probably 0 or stale static data until the next update comes. There is an event when analysis complete fully, it may be possible to wire it up - but it wont be perfect still.

jakebailey: I can see from adding some extra logging that ReportRemaining stops getting called at some point even though there is some analysis being done. I guess this doesn't have to do with disposing at all (which I expected), but just generally how this first implementation writes the number remaining, which could be improved in the future.

MikhailArkhipov:
@AlexanderSher - is there anything better we could do at the moment?

PR #696, opened 2019-03-05
Fix analysis completion regression

MikhailArkhipov:
Caused by [leftover] change in https://github.com/Microsoft/python-language-server/pull/684


src/LanguageServer/Impl/Resources.resx
```diff
@@ -157,6 +157,6 @@
     <value>Unable to get analysis for the selected expression.</value>
   </data>
   <data name="RestartingAnalysis" xml:space="preserve">
```
jakebailey: Should the resource name be renamed if the text is changing?

Issue #636, opened 2019-02-20
Racy completion shows completions from the wrong scope


jakebailey:
In this specific example below, delete and replace the `.` at the end after `self` repeatedly (slowly, or quickly), and you'll see a different completion. One comes from the correct scope (shows `v`), but otherwise it gets evaluated in the global scope. It seems to be linked to when the red parse squiggle shows up.

![image](https://user-images.githubusercontent.com/5341706/53111578-280a3b80-34f2-11e9-8e6c-46fe9409e252.png)

The whole code below is required, removing some of this code seems to make the analysis fast enough to not show the issue, which makes me think it's a race.

```python
from typing import TypeVar, Generic

_T = TypeVar("_T")

class Box(Generic[_T]):
    def __init__(self, v: _T):
        self.v = v
    
    def get(self) -> _T:
        return self.v


boxedint = Box(1234)
intval = boxedint.get()

boxedstr = Box("foo")
strval = boxedstr.get()


Box(Box(Box(123))).get().get().get().bit_length()

class UntypedBox():
    def __init__(self, v):
        self.v = v

    def get(self):
        return self.

```

MikhailArkhipov:
Works for me, either #655 or #665 fixed it.

PR #698, opened 2019-03-05
Module-scoped imports

AlexanderSher:
- Make modules and subpackages available for __init__.py packages
- Scope imports to module
- Fix completion in tensorflow:
image

### DIFF
--- a/src/Analysis/Ast/Impl/Modules/PythonModule.cs
+++ b/src/Analysis/Ast/Impl/Modules/PythonModule.cs
@@ -74,8 +74,8 @@ namespace Microsoft.Python.Analysis.Modules {
             Services = services ?? throw new ArgumentNullException(nameof(services));
             ModuleType = moduleType;
 
-            Log = services?.GetService<ILogger>();
-            Interpreter = services?.GetService<IPythonInterpreter>();
+            Log = services.GetService<ILogger>();
+            Interpreter = services.GetService<IPythonInterpreter>();
             Analysis = new EmptyAnalysis(services, this);
 
             _diagnosticsService = services.GetService<IDiagnosticsService>();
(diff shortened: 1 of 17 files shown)


src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -330,9 +330,7 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
 
         public void Reset(string content) {
             lock (AnalysisLock) {
-                if (content != Content || content == null) {
-                    ContentState = State.None;
-                    Services.GetService<IPythonAnalyzer>().InvalidateAnalysis(this);
+                if (content != Content) {
```
MikhailArkhipov: I had an issue when analysis didn't start on `pip install` reload.

src/Analysis/Ast/Impl/Modules/PythonPackage.cs
```diff
@@ -25,28 +27,54 @@ namespace Microsoft.Python.Analysis.Modules {
     /// Represents package with child modules. Typically
     /// used in scenarios such as 'import a.b.c'.
     /// </summary>
-    internal sealed class PythonPackage : PythonModule, IPythonPackage {
-        private readonly ConcurrentDictionary<string, IPythonModule> _childModules = new ConcurrentDictionary<string, IPythonModule>();
+    internal sealed class PythonVariableModule : IPythonModule, IEquatable<IPythonModule> {
```
MikhailArkhipov: Didn't like the `package` name? :-)

src/Analysis/Ast/Impl/Modules/PythonModule.cs
AlexanderSher: Merge issue

src/Analysis/Ast/Impl/Modules/PythonPackage.cs
AlexanderSher: It isn't a package anymore. I actually don't know how to call it better. Need to update the comment btw.

jakebailey:
This PR didn't fix it for me:

image

The analysis never completes, and it completion does the same thing as in #636.

jakebailey:
Hmm, eventually it completed and `tf.nn` shows the right thing, but I can see in the logs quite a few repeat analysis items, like `tensorflow.python.estimator.keras(Library)` 5 times.

AlexanderSher:
Can you send me a full log?

PR #702, opened 2019-03-05
Fix #24: 'Go to Definition' for relative imports not resolving directly to the appropriate module

### LINKED ISSUE #24: 'Go to Definition' for relative imports not resolving directly to the appropriate module
Let's say I have `pkg` and `pkg.subpkg`, both packages. I also have `pkg.mod` and `pkg.subpkg.submod`, both modules. If I do `from .. import mod` from within `pkg.subpkg.submode` I get asked which package to resolve to -- `pkg/__init__.py` or `pkg/mod.py` -- even though `pkg.__init__` is empty. I would expect the lack of ambiguity to not ask me which package to resolve to.

Attached are the test files I was using.
[pkg.zip](https://github.com/Microsoft/python-language-server/files/2349710/pkg.zip)

AlexanderSher:


MikhailArkhipov:
Would be good to add test to `GoToDefinitionTests`

PR #703, opened 2019-03-05
Wire up events from RDT to module resolution

### LINKED ISSUE #700: Closing and reopening documents may stop editor functionality
Module is not removed from module resolution and the analyzer. Occasionally file may get recognized as library rather than user code.

MikhailArkhipov:
Fixes #700 

- Make sure RDT maintains proper lock count for imported/opened modules.
- Update diagnostics reporting when module is closed or removed.
- Fix search paths that were duplicated in user and interpreter paths.
- Fix determination if module is user or library.
- Tests.


src/Analysis/Core/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -233,6 +233,20 @@ public PathResolverSnapshot(PythonLanguageVersion pythonLanguageVersion)
             return new ImportNotFound(fullName);
         }
 
+        public bool IsLibraryFile(string filePath) {
```
AlexanderSher:
```
=> TryFindModule(filePath, out var edge, out _) && IsLibraryPath(edge.First.End);
```

src/LanguageServer/Test/RdtTests.cs
```diff
@@ -63,5 +70,91 @@ public void TestInitialize()
...
+            await Services.GetService<IPythonAnalyzer>().WaitForCompleteAnalysisAsync();
+
+            var docLc1 = rdt.First(d => d.Name.Contains("LockCount1"));
+            var docLc2 = rdt.First(d => d.Name.Contains("LockCount2"));
+            var docLc3 = rdt.First(d => d.Name.Contains("LockCount3"));
+
+            var ds = GetDiagnosticsService();
+            PublishDiagnostics();
+            ds.Diagnostics.Count.Should().Be(4);
+            ds.Diagnostics[uri].Count.Should().Be(0);
```
AlexanderSher: `ds.Diagnostics[uri].Should.BeEmpty()`

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
```diff
@@ -187,7 +194,8 @@ public IPythonModule GetSpecializedModule(string name)
             var interpreterPaths = await GetSearchPathsAsync(cancellationToken);
             addedRoots.UnionWith(PathResolver.SetInterpreterSearchPaths(interpreterPaths));
 
-            addedRoots.UnionWith(SetUserSearchPaths(_interpreter.Configuration.SearchPaths));
+            var userSearchPaths = _interpreter.Configuration.SearchPaths.Except(interpreterPaths, StringExtensions.PathsStringComparer);
+            addedRoots.UnionWith(SetUserSearchPaths(userSearchPaths));
```
jakebailey: I know this was here before you modified it, but does the use of a `HashSet` here mean that the path resolver is going to be nondeterministically choosing how import paths are resolved? From what I can tell, `ReloadModulePaths` iterates over `addedRoots`, calling `AddModulePath` on each, which then adds children to the path resolver internally in a specific order.

jakebailey: Hmm, maybe that's not how it actually works, given this does `PathResolver.SetRoot`, `PathResolver.SetInterpreterSearchPaths`, then `PathResolver.SetUserSearchPaths` in a row before `ReloadModulesPaths`, so I probably just misunderstood what `ReloadModulePaths` did.

jakebailey: @AlexanderSher 	added this set in #641 so I'm probably wrong about it all.

src/LanguageServer/Impl/Implementation/Server.Documents.cs
```diff
@@ -56,9 +56,13 @@ public sealed partial class Server {
         }
 
         public void DidChangeWatchedFiles(DidChangeWatchedFilesParams @params) {
+            _disposableBag.ThrowIfDisposed();
             foreach (var c in @params.changes.MaybeEnumerate()) {
-                _disposableBag.ThrowIfDisposed();
-                // TODO: handle?
+                switch (c.type) {
+                    case FileChangeType.Deleted:
```
jakebailey: Do you know if this case gets hit?

image

`b` and `food` still seem to exist after I make `b` with `def foo(): return 123`, import it in `a`, then delete `b`, but I could be testing something different than what this handles.

MikhailArkhipov: Frankly, I haven't seen it hit. I don't know when VSC sends it, this is why we have manual library watcher. But it doesn't hurt to handle.

MikhailArkhipov: We could start watching workspace manually too, but that's a different issue.

src/Analysis/Ast/Impl/Documents/RunningDocumentTable.cs
```diff
@@ -65,12 +66,14 @@ private class DocumentEntry {
             lock (_lock) {
                 entry = FindDocument(null, uri);
                 if (entry == null) {
+                    var resolver = _services.GetService<IPythonInterpreter>()?.ModuleResolution?.CurrentPathResolver;
```
AlexanderSher: What is the case when `_services.GetService<IPythonInterpreter>()` or `ModuleResolution` is null?

src/Analysis/Ast/Impl/Documents/RunningDocumentTable.cs
```diff
@@ -114,39 +117,63 @@ private class DocumentEntry {
             }
         }
 
+        public int LockDocument(Uri uri) {
+            lock (_lock) {
+                if (_documentsByUri.TryGetValue(uri, out var entry)) {
+                    entry.LockCount++;
+                    return entry.LockCount;
```
AlexanderSher: `return ++entry.LockCount`?

src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -244,6 +244,7 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
             _diagnosticsService?.Remove(Uri);
             _allProcessingCts.Cancel();
             _allProcessingCts.Dispose();
+            Services.GetService<IPythonAnalyzer>().RemoveAnalysis(this);
```
AlexanderSher: Can we use `DisposeToken` here instead of `_allProcessingCts`?

src/LanguageServer/Test/RdtTests.cs
```diff
@@ -63,5 +70,91 @@ public void TestInitialize()
...
+
+        [TestMethod, Priority(0)]
+        public async Task LockCount() {
+            var uri1 = TestData.GetDefaultModuleUri();
+            await CreateServicesAsync(PythonVersions.LatestAvailable3X, uri1.AbsolutePath);
+            var rdt = Services.GetService<IRunningDocumentTable>();
+
+            rdt.OpenDocument(uri1, "from LockCount1 import *");
+            var doc = rdt.GetDocument(uri1);
+            await doc.GetAstAsync(CancellationToken.None);
```
AlexanderSher: You shouldn't need this line anymore.

src/LanguageServer/Test/RdtTests.cs
```diff
@@ -63,5 +70,91 @@ public void TestInitialize()
...
+            var uri = TestData.GetDefaultModuleUri();
+            await CreateServicesAsync(PythonVersions.LatestAvailable3X, uri.AbsolutePath);
+            var rdt = Services.GetService<IRunningDocumentTable>();
+
+            rdt.OpenDocument(uri, "from LockCount1 import *");
+            var doc = rdt.GetDocument(uri);
+            await doc.GetAstAsync(CancellationToken.None);
+            await Services.GetService<IPythonAnalyzer>().WaitForCompleteAnalysisAsync();
+
+            var docLc1 = rdt.First(d => d.Name.Contains("LockCount1"));
```
AlexanderSher: `rdt.GetDocument`?

src/Analysis/Ast/Impl/Modules/PythonModule.cs
MikhailArkhipov: Sure

src/LanguageServer/Test/RdtTests.cs
MikhailArkhipov: Yeah, it was just so they all look the same since I know uri for for first but not others

src/LanguageServer/Test/RdtTests.cs
MikhailArkhipov: But I do. You may want to look into it. If I remove it, test fails b/c 
```
var docLc2 = rdt.First(d => d.Name.Contains("LockCount2"));
```
`Sequence contains no elements`

PR #706, opened 2019-03-06
Fix nullref during testing

jakebailey:
Running analysis tests after #687 fails because there is no progress reporting service.


MikhailArkhipov:
Yeah, fixed it while working on another issue.

Issue #620, opened 2019-02-15
Handle complex __all__ assignments


jakebailey:
We have basic `__all__` handling, but won't be able to handle more complicated expressions. This will likely be required for #129.

Example:

https://github.com/django/django/blob/741ce81a426e4d0cd7581d98d3b8e3290f513f09/django/db/models/__init__.py#L35-L48

```python
__all__ = aggregates_all + constraints_all + fields_all + indexes_all
__all__ += [
    'ObjectDoesNotExist', 'signals',
    'CASCADE', 'DO_NOTHING', 'PROTECT', 'SET', 'SET_DEFAULT', 'SET_NULL',
    'ProtectedError',
    'Case', 'Exists', 'Expression', 'ExpressionList', 'ExpressionWrapper', 'F',
    'Func', 'OuterRef', 'RowRange', 'Subquery', 'Value', 'ValueRange', 'When',
    'Window', 'WindowFrame',
    'FileField', 'ImageField', 'OrderWrt', 'Lookup', 'Transform', 'Manager',
    'Prefetch', 'Q', 'QuerySet', 'prefetch_related_objects', 'DEFERRED', 'Model',
    'FilteredRelation',
    'ForeignKey', 'ForeignObject', 'OneToOneField', 'ManyToManyField',
    'ManyToOneRel', 'ManyToManyRel', 'OneToOneRel',
]
```

jakebailey:
More examples:

- Torch: https://github.com/Microsoft/python-language-server/issues/690#issuecomment-469890261
- Numpy: https://github.com/Microsoft/python-language-server/issues/642#issuecomment-469890953

Issue #619, opened 2019-02-15
__all__ not being respected from an `import *`


brettcannon:
Microsoft/vscode-python#494

```python
# A.py
import B
B. # intellisense shows `foo` and `bar`, but should only show `foo`.

# B.py
from C import *

# C.py
__all__ = ['foo']
foo = 42
bar = 24
```

jakebailey:
Works in the new language server.

![image](https://user-images.githubusercontent.com/5341706/52825823-2ca69e00-3073-11e9-96c7-d38794d14a9f.png)

We have basic `__all__` support, though not for more complicated expressions (like those which combine multiple modules' `__all__`s into a single list).

jakebailey:
An unfortunate side effect of handling this but only partially means that we may _under_ complete in some situations (especially projects like Django which do lots of fun stuff). I've opened #620 to track improving the support.

jakebailey:
Repoening, the code which filters `__all__` has been temporarily removed as commonly used libraries are using it in ways we can't handle yet.

PR #708, opened 2019-03-06
Unfilter __all__

### LINKED ISSUE #642: Missing numpy and tensorflow completion, members not showing up

jakebailey:
Fixes #642.
Reopens #619.
Related #620.

This unfilters `__all__` except for things which are directly from `typing` or are imported modules, though even that may be too strong. The old LS didn't look at `__all__` whatsoever, so maybe all of these should be allowed to come through; you tell me.

Numpy:

image

### DIFF
--- a/src/Analysis/Ast/Impl/Modules/PythonModule.cs
+++ b/src/Analysis/Ast/Impl/Modules/PythonModule.cs
@@ -149,20 +149,11 @@ namespace Microsoft.Python.Analysis.Modules {
         #region IMemberContainer
         public virtual IMember GetMember(string name) => Analysis.GlobalScope.Variables[name]?.Value;
         public virtual IEnumerable<string> GetMemberNames() {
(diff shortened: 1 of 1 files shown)


jakebailey:
Amending my statement: Even though the old LS didn't look at `__all__` doesn't mean that it didn't filter what modules exported. I haven't looked at that code to see what it did, as I assume it didn't just pass _everything_ through. Probably did what I'm changing to above (no modules or typing).

jakebailey:
Old LS exposed variable defs, which I think would be about the same as what I've written. I'll merge this and we can tweak it later; at least numpy works now. 😄

Issue #699, opened 2019-03-05
Find All References in new LS


MikhailArkhipov:
Precursor for https://github.com/Microsoft/python-language-server/issues/577. Needs #682 as currently in the sequence of
```python
x = 1
x = 2
x = 3
```
only the last assignment is known.

Issue #639, opened 2019-02-20
Implement detection of undefined variables


MikhailArkhipov:
Replacement using new LS

Issue #577, opened 2019-02-04
LS Rename using new analysis


MikhailArkhipov:
Rename needs information on all documents, so this entails to
- Find definition of the symbol being renamed #567
- Get dependencies of the module that contains the definition #432
- Run lightweight analysis on the dependencies not analyzed yet. Lightweight means ignoring library imports for faster search since symbol being renamed is in the user code.

MikhailArkhipov:
Heavily depends on dependency search and resolution and specific imports mode.

MikhailArkhipov:
Requires https://github.com/Microsoft/python-language-server/issues/577

PR #693, opened 2019-03-04
Linter for undefined variables

MikhailArkhipov:
Fixes #639 
Fixes #395 
Fixes #461 
Fixes #475 
Fixes #476 
Fixes #513
Fixes #682
Fixes #734

- Walk AST post analysis and use analysis to find out declared variables. Special filtering and handling of comprehensions. Includes #711
- Adds support for analysis of comprehensions. Fixes issue when comprehension variables were not declared in scope. Adds handling of tuple expressions in `for in` loops.
- Adds support for lambdas. Enables related tests.

PR #711, opened 2019-03-06
Implement storing of variable definition locations

### LINKED ISSUE #682: Implement tracking variable values by location
For better intellisense we need to track variable assignment locations so in

```python
x = 1
x. # yields completions for int
x = 's'
x. # yields completions for str
```
see also #668 for `isinstance` so

```python
from typing import Union, List

class Foo: pass
class Bar: pass

FB = Union[Foo, List[Bar]]

def do(fb: FB):
    if isinstance(fb, list):
        fb. # completions for list
        return fb[0]
    else:
        fb. # completions for Foo
```

MikhailArkhipov:
Fixes #682

Adds collection of variable assignment locations. For #639, #699 and #577

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/AnalysisWalker.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/AnalysisWalker.cs
@@ -66,11 +66,7 @@ namespace Microsoft.Python.Analysis.Analyzer {
             return false;
         }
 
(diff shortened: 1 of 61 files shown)


jakebailey:
The example from #682 doesn't work for me:

image

MikhailArkhipov:
in favor of #693

PR #718, opened 2019-03-07
Reduce amount of async code in PythonAnalyzer

AlexanderSher:
Stabilization fixes:
- IoC is replaced with direct list of dependencies
- All operations in `ProjectAnalyzerEntry` now happen under individual lock
- Cancellation of analysis can't abort graph change in the middle of the process
- Allow access to AST instance only with corresponding module instance
- Update IPythonModule instance when entry is invalidated

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
@@ -113,6 +113,7 @@ namespace Microsoft.Python.Analysis.Analyzer {
         public void EnqueueDocumentForAnalysis(IPythonModule module, ImmutableArray<IPythonModule> dependencies) {
             var key = new ModuleKey(module);
             PythonAnalyzerEntry entry;
+            int version;
             lock (_syncObj) {
                 if (!_analysisEntries.TryGetValue(key, out entry)) {
                     return;
@@ -123,10 +124,11 @@ namespace Microsoft.Python.Analysis.Analyzer {
                     return;
                 }
 
-                entry.Invalidate(significantDependencies, _version + 1);
+                version = _version + 1;
+                entry.Invalidate(significantDependencies, version);
             }
 
-            AnalyzeDocumentAsync(key, entry, default).DoNotWait();
+            AnalyzeDocumentAsync(key, entry, version, default).DoNotWait();
 
             bool IsSignificant(IPythonModule m)
                 => m != entry.Module && ((m.ModuleType == ModuleType.User && m.Analysis.Version < entry.AnalysisVersion) || m.Analysis is EmptyAnalysis);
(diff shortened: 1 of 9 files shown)


src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -281,19 +283,89 @@ private void StartAnalysis(IDependencyChainNode<PythonAnalyzerEntry> node, int v
                 var analysis = new DocumentAnalysis((IDocument)module, version, walker.GlobalScope, walker.Eval);
 
                 (module as IAnalyzable)?.NotifyAnalysisComplete(analysis);
-                node.Value.TrySetAnalysis(analysis, version, _syncObj);
+                entry.TrySetAnalysis(analysis, version);
+                node.Commit();
 
-                _log?.Log(TraceEventType.Verbose, $"Analysis of {module.Name}({module.ModuleType}) complete in {(stopWatch.Elapsed - startTime).TotalMilliseconds} ms.");
+                _log?.Log(TraceEventType.Verbose, $"Analysis of {module.Name}({module.ModuleType}) is completed in {(stopWatch.Elapsed - startTime).TotalMilliseconds} ms.");
```
jakebailey: Nit: "Analysis of _ completed in _" is more grammatically correct ("analysis of _ complete in _" was also fine).

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -281,19 +283,89 @@ private void StartAnalysis(IDependencyChainNode<PythonAnalyzerEntry> node, int v
...
             } catch (OperationCanceledException oce) {
-                node.Value.TryCancel(oce, version, _syncObj);
+                node.Value.TryCancel(oce, version);
+                node.Skip();
+                var module = node.Value.Module;
+                _log?.Log(TraceEventType.Verbose, $"Analysis of {module.Name}({module.ModuleType}) is canceled.");
```
jakebailey: Nit: Just "analysis of _ canceled" to fit with the above comment.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -281,19 +283,89 @@ private void StartAnalysis(IDependencyChainNode<PythonAnalyzerEntry> node, int v
...
             } catch (Exception exception) {
-                node.Value.TrySetException(exception, version, _syncObj);
+                var module = node.Value.Module;
+                node.Value.TrySetException(exception, version);
+                node.Commit();
+                _log?.Log(TraceEventType.Verbose, $"Analysis of {module.Name}({module.ModuleType}) has failed.");
```
jakebailey: Forgive another nitpick, but given the other two, "analysis of _ failed" is more consistent, this one is even more minor though.

Should this print elapsed time as well?

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
jakebailey: I think this should also print the elapsed time, it might be useful.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
AlexanderSher: Elapsed time will be inaccurate at that point, cause building/catching exception is a long operation by itself.

jakebailey: Fair enough.

jakebailey:
After that most recent commit, the analysis deadlocks (using a file which does `import numpy`).

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -281,19 +296,97 @@ private void StartAnalysis(IDependencyChainNode<PythonAnalyzerEntry> node, int v
...
+
+            var dependenciesHashSet = new HashSet<ModuleKey>();
+            var isTypeshed = module is StubPythonModule stub && stub.IsTypeshed;
+            var moduleResolution = module.Interpreter.ModuleResolution;
+            var pathResolver = isTypeshed
+                ? module.Interpreter.TypeshedResolution.CurrentPathResolver
+                : moduleResolution.CurrentPathResolver;
+
+            if (module.Stub != null) {
+                dependenciesHashSet.Add(new ModuleKey(module.Stub));
```
MikhailArkhipov: Do we need stubs as dependencies? I think we used to forcibly analyze and attach them at import time without going through actual dependency analysis.

AlexanderSher: Stub must be analyzed before the module. So graph must be aware of this dependency. Preserving this dependency is important in cases when stub has its own analysis dependency on another stub.

PR #722, opened 2019-03-08
Various import completion fixes

AlexanderSher:
- Import search now is always happening through PathResolver
- Filtering empty name at the end of module path
- Fixes #716: Language Server 0.2.16 crashes (Request textDocument/completion failed.)
- Fixes #594: Completion support in 'from . import'

PR #733, opened 2019-03-11
Fix #719: Scraper produces incorrect output for the Anaconda 2X builtin

### LINKED ISSUE #719: Scraper produces incorrect output for the Anaconda 2X builtin
Testing the beta channel of the language server I ran into an issue starting the server. So far I've seen it happen with multiple projects a couple of seconds after a windows reload.

VSCode Info
Version: 1.32.0
Commit: 507312a3e3b34b084b467dfd983263bc7c9d87e0
Date: 2019-03-06T18:53:01.021Z
Electron: 3.1.3
Chrome: 66.0.3359.181
Node.js: 10.2.0
V8: 6.6.346.32
OS: Linux x64 4.18.0-16-generic

Python 2.7.15
Virtualenv

```
[Error - 10:14:51 PM] Server initialization failed.
  Message: Object reference not set to an instance of an object.
  Code: -32000 
   at Microsoft.Python.Analysis.Modules.Resolution.MainModuleResolution.CreateModule(String name)
   at Microsoft.Python.Analysis.Modules.Resolution.ModuleResolutionBase.ModuleRef.GetOrCreate(String name, ModuleResolutionBase mrb)
   at Microsoft.Python.Analysis.Modules.Resolution.ModuleResolutionBase.GetOrLoadModule(String name)
   at Microsoft.Python.Analysis.Analyzer.PythonAnalyzer.LoadMissingDocuments(IPythonInterpreter interpreter, ImmutableArray`1 missingKeys)
   at Microsoft.Python.Analysis.Analyzer.PythonAnalyzer.AnalyzeDocumentAsync(ModuleKey key, PythonAnalyzerEntry entry, CancellationToken cancellationToken)
   at Microsoft.Python.Core.TaskExtensions.DoNotWait(Task task)
   at Microsoft.Python.Analysis.Analyzer.PythonAnalyzer.EnqueueDocumentForAnalysis(IPythonModule module, PythonAst ast, Int32 bufferVersion, CancellationToken cancellationToken)
   at Microsoft.Python.Analysis.Modules.PythonModule.Parse(CancellationToken cancellationToken)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot)
--- End of stack trace from previous location where exception was thrown ---
   at Microsoft.Python.Analysis.Modules.PythonModule.GetAstAsync(CancellationToken cancellationToken)
   at Microsoft.Python.Analysis.Modules.PythonModule.LoadAndAnalyzeAsync(CancellationToken cancellationToken)
   at Microsoft.Python.Analysis.Modules.Resolution.MainModuleResolution.LoadBuiltinTypesAsync(CancellationToken cancellationToken)
   at Microsoft.Python.Analysis.Analyzer.PythonInterpreter.LoadBuiltinTypesAsync(String root, IServiceManager sm, CancellationToken cancellationToken)
   at Microsoft.Python.Analysis.Analyzer.PythonInterpreter.CreateAsync(InterpreterConfiguration configuration, String root, IServiceManager sm, CancellationToken cancellationToken)
   at Microsoft.Python.LanguageServer.Implementation.Server.InitializeAsync(InitializeParams params, CancellationToken cancellationToken)
   at Microsoft.Python.LanguageServer.Implementation.LanguageServer.Initialize(JToken token, CancellationToken cancellationToken)
```

AlexanderSher:
Also:
- fixed messaging for relative imports that go beyond top-level 
- made amount of tasks in `PythonAnalyzer` dependent on the number of CPU cores

### DIFF
--- a/src/Parsing/Test/PythonVersions.cs
+++ b/src/Parsing/Test/PythonVersions.cs
@@ -36,12 +36,16 @@ namespace Microsoft.Python.Parsing.Tests {
         public static readonly InterpreterConfiguration Anaconda27_x64 = GetAnacondaVersion(PythonLanguageVersion.V27, InterpreterArchitecture.x64);
         public static readonly InterpreterConfiguration Anaconda36 = GetAnacondaVersion(PythonLanguageVersion.V36, InterpreterArchitecture.x86);
         public static readonly InterpreterConfiguration Anaconda36_x64 = GetAnacondaVersion(PythonLanguageVersion.V36, InterpreterArchitecture.x64);
+        public static readonly InterpreterConfiguration Anaconda37 = GetAnacondaVersion(PythonLanguageVersion.V37, InterpreterArchitecture.x86);
+        public static readonly InterpreterConfiguration Anaconda37_x64 = GetAnacondaVersion(PythonLanguageVersion.V37, InterpreterArchitecture.x64);
         public static readonly InterpreterConfiguration IronPython27_x64 = GetIronPythonVersion(true);
         public static readonly InterpreterConfiguration Jython27 = GetJythonVersion(PythonLanguageVersion.V27);
 
         public static IEnumerable<InterpreterConfiguration> AnacondaVersions => GetVersions(
             Anaconda36,
             Anaconda36_x64,
+            Anaconda37,
+            Anaconda37_x64,
             Anaconda27,
             Anaconda27_x64);
 
@@ -82,6 +86,8 @@ namespace Microsoft.Python.Parsing.Tests {
             Python27_x64).FirstOrDefault() ?? NotInstalled("v2");
 
         public static InterpreterConfiguration LatestAnaconda3X => GetVersions(
+            Anaconda37,
+            Anaconda37_x64,
             Anaconda36,
             Anaconda36_x64).FirstOrDefault() ?? NotInstalled("Anaconda v3");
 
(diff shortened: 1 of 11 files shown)


src/Parsing/Test/PythonVersions.cs
```diff
@@ -81,6 +81,14 @@ public static class PythonVersions {
             Python27,
             Python27_x64).FirstOrDefault() ?? NotInstalled("v2");
 
+        public static InterpreterConfiguration LatestAnaconda3X => GetVersions(
+            Anaconda36,
```
jakebailey: The latest Anaconda is at Python 3.7, can that be added as an option?

jakebailey:
So with my example without an `__init__.py`, this now works:

image

However, if you type `from .` you don't get any completion for `B`. (It also shows the same unresolved-import message with an empty string like in #722, but I'm not sure if that's in scope.)

src/Parsing/Test/PythonVersions.cs
AlexanderSher: Sure

AlexanderSher:
No, the change was for the cases like  `from ..abc` in the root.

jakebailey:
You're right, my example works in `master`. Sorry for the noise.

jakebailey:
This is likely also out of scope, but the exception that gets thrown when you do something like `from ..B import b` otu of the root gets eaten completely. It ends up causing an exception over in the indexer. Do we know why we're eating these exceptions?

AlexanderSher:
I'm not sure I understand what exception do you mean. Analysis error is shown in UI (red squiggles are from pylint):
image

jakebailey:
I mean before this patch, on `master`. If you were to write that same code, then an exception is thrown _somewhere_ and never logged, and then causes exceptions elsewhere.

PR #741, opened 2019-03-13
Empty lambda before end of file not reporting any error

brianbok:
We can now report when we don't find colon separator for expression body.
Before, when we had a lambda token followed by end of file, we were not reporting anything.


src/Parsing/Test/ParserTests.cs
```diff
@@ -2170,12 +2170,60 @@ public class ParserTests {
                     CheckSuite(
                         CheckLambdaStmt(new[] { CheckParameter("x") }, One),
                         CheckLambdaStmt(new[] { CheckParameter("x", ParameterKind.List) }, One),
-                        CheckLambdaStmt(new[] { CheckParameter("x", ParameterKind.Dictionary) }, One)
+                        CheckLambdaStmt(new[] { CheckParameter("x", ParameterKind.Dictionary) }, One),
+                        CheckLambdaStmt(new Action<Parameter>[] { }, One)
```
jakebailey: `NoParameters`

src/Parsing/Test/ParserTests.cs
```diff
@@ -2170,12 +2170,60 @@ public class ParserTests {
...
+        }
+
+        [TestMethod, Priority(0)]
+        public void LambdaErrors() {
+            foreach (var version in AllVersions) {
+                var errors = new CollectingErrorSink();
+                CheckAst(
+                    ParseFile("LambdaErrors.py", errors, version),
+                    CheckSuite(
+                        CheckLambdaStmt(new Action<Parameter>[] { CheckParameter(null, ParameterKind.Normal, null, null) }, CheckErrorExpr()),
```
jakebailey: The other tests in this file do `new[]`, which is probably preferred.

PR #742, opened 2019-03-13
Fix performance bottleneck fetching module content

MikhailArkhipov:
- Cache document content. Converting string builder every time is a time sink and a memory waste.

Tensorflow is down to 50s
image

Original
image

### DIFF
--- a/src/Analysis/Ast/Impl/Documents/DocumentBuffer.cs
+++ b/src/Analysis/Ast/Impl/Documents/DocumentBuffer.cs
@@ -17,17 +17,15 @@ using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Text;
-using System.Threading;
-using Microsoft.Python.Core.Diagnostics;
 using Microsoft.Python.Parsing;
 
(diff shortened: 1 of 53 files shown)


src/Analysis/Ast/Impl/Documents/DocumentBuffer.cs
```diff
@@ -17,27 +17,27 @@
...
             Version = version;
             _sb.Clear();
             if (!string.IsNullOrEmpty(content)) {
                 _sb.Append(content);
             }
+            _cachedText = null;
         }
 
         public void Update(IEnumerable<DocumentChange> changes) {
+
```
AlexanderSher: `SplitLines` can also be optimized, cause it always calls `sb.ToString()`, but I'm not sure if that would be a huge improvement.

MikhailArkhipov: Yes, I can just borrow https://github.com/Microsoft/RTVS/tree/master/src/VsCode/LanguageServer/Impl/Text too.  But it is not visible in the updated profile so it is low pri.

PR #743, opened 2019-03-13
Optimize scope search and variable lookup

MikhailArkhipov:
Another ~20%

image

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Scopes.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Scopes.cs
@@ -14,6 +14,7 @@
 // permissions and limitations under the License.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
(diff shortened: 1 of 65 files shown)


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Scopes.cs
```diff
@@ -136,9 +155,12 @@ public IPythonType GetTypeFromAnnotation(Expression expr, LookupOptions options
...
 
             if (node.Parent != null) {
-                fromScope = gs
-                    .TraverseBreadthFirst(s => s.Children.OfType<Scope>())
-                    .FirstOrDefault(s => s.Node == node.Parent);
+                if (!_scopeLookupCache.TryGetValue(node.Parent, out fromScope)) {
+                    fromScope = gs
+                        .TraverseDepthFirst(s => s.Children.OfType<Scope>())
+                        .FirstOrDefault(s => s.Node == node.Parent);
+                    _scopeLookupCache[node.Parent] = fromScope;
```
AlexanderSher: Can't we add the scope to the dictionary when it is created below?

MikhailArkhipov: Good point

PR #747, opened 2019-03-13
Update to support IntelliCode

### LINKED ISSUE #536: Update IntelliCode with new LS

MikhailArkhipov:
Fixes #536 

- Extend completion item for IC since it needs associated types.
- Reactivate extension loading.

### DIFF
--- a/src/LanguageServer/Impl/Completion/CompletionItemSource.cs
+++ b/src/LanguageServer/Impl/Completion/CompletionItemSource.cs
@@ -38,7 +38,7 @@ namespace Microsoft.Python.LanguageServer.Completion {
         public CompletionItem CreateCompletionItem(string text, IMember member, IPythonType self = null, string label = null)
             => CreateCompletionItem(text, ToCompletionItemKind(member?.MemberType ?? PythonMemberType.Class), member, self, label);
 
-        public CompletionItem CreateCompletionItem(string text, CompletionItemKind kind, IMember member, IPythonType self = null, string label = null) {
+        public CompletionItemEx CreateCompletionItem(string text, CompletionItemKind kind, IMember member, IPythonType self = null, string label = null) {
             var t = member?.GetPythonType();
             var docFormat = _docSource.DocumentationFormat;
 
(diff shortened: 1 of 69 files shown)


src/LanguageServer/Impl/Completion/CompletionItemSource.cs
```diff
@@ -47,19 +47,23 @@ public CompletionItem CreateCompletionItem(string text, IMember member, IPythonT
...
                 insertText = text,
                 insertTextFormat = docFormat,
                 // Place regular items first, advanced entries last
                 sortText = char.IsLetter(text, 0) ? "1" : "2",
                 kind = kind,
-                documentation = !t.IsUnknown() ? _docSource.GetHover(label ?? text, member, self) : null
+                documentation = !t.IsUnknown() ? _docSource.GetHover(label ?? text, member, self) : null,
+                // Custom fields used by the LS extensions that may modify
+                // the completion list. Not passed to the client.
+                Member = member,
```
jakebailey: Does this mean that they're going to continue to be reliant on the internal details of the analyzer?

MikhailArkhipov: These are public interfaces, i,e, `IMember`, `IPythonType`, `IDocumentAnalysis`

jakebailey: I assume that if we modify those types, then they'd break? I had just thought that us splitting apart the analysis, the language server, and the parser into their own assemblies was to prevent breakage, but I guess if they need it then there's probably no recourse.

MikhailArkhipov: We dont modify public interfaces after they are published. It would be `IDocumentAnalysis2` etc

src/LanguageServer/Impl/Implementation/Server.Editor.cs
```diff
@@ -14,11 +14,13 @@
 // permissions and limitations under the License.
 
 using System;
+using System.ComponentModel.Design;
```
jakebailey: What is this using for?

MikhailArkhipov: Auto-inserted by resharper I guess. Not used, cleaned up.

PR #749, opened 2019-03-13
Stability improvements

AlexanderSher:
- Reanalyze user module out of the order if required. Helps providing completions earlier (before full initial analysis is completed)
- Skip walker if it is outdated and doesn't have a lot of unanalyzed modules.
- Initialize PriorityProducerConsumer<IDependencyChainNode<TValue>> only when GetNextAsync is called for the first time
- Fix ProcessServices so that error stream doesn't hang the process
- Full Anaconda 2X analysis passes

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
@@ -16,6 +16,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Python.Analysis.Core.DependencyResolution;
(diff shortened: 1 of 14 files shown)


src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -315,121 +316,95 @@ private void StartAnalysis(IDependencyChainNode<PythonAnalyzerEntry> node, int v
...
-        private bool TryFindDependencies(PythonAnalyzerEntry entry, int version, CancellationToken cancellationToken, out ImmutableArray<ModuleKey> dependencies) {
-            IPythonModule module;
-            PythonAst ast;
-            lock (_syncObj) {
-                if (!entry.IsValidVersion(version, out module, out ast)) {
-                    dependencies = ImmutableArray<ModuleKey>.Empty;
-                    return false;
-                }
-            }
+        private void StartAnalysis(PythonAnalyzerEntry entry, int version, CancellationToken cancellationToken)
```
MikhailArkhipov: Do you want this to be Task too as above?

PR #759, opened 2019-03-15
More reliable progress indication

MikhailArkhipov:


src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -221,23 +223,21 @@ public sealed class PythonAnalyzer : IPythonAnalyzer, IDisposable {
...
-                    _progress?.ReportRemaining(remaining);
 
                     if (_log != null) {
-                        if (remaining == 0) {
-                            _log.Log(TraceEventType.Verbose, $"Analysis version {walker.Version} of {originalRemaining} entries has been completed in {stopWatch.Elapsed.TotalMilliseconds} ms.");
-                        } else if (remaining < originalRemaining) {
-                            _log.Log(TraceEventType.Verbose, $"Analysis version {walker.Version} has been completed in {stopWatch.Elapsed.TotalMilliseconds} ms with {originalRemaining - remaining} entries analyzed and {remaining} entries skipped.");
-                        } else {
-                            _log.Log(TraceEventType.Verbose, $"Analysis version {walker.Version} of {originalRemaining} entries has been canceled after {stopWatch.Elapsed.TotalMilliseconds}.");
+                        if (walker.Remaining == 0) {
```
AlexanderSher: We need a log even if `walker.Remaining` is not zero

MikhailArkhipov: Weird merge issue

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -284,10 +284,9 @@ public sealed class PythonAnalyzer : IPythonAnalyzer, IDisposable {
         private void LoadMissingDocuments(IPythonInterpreter interpreter, ImmutableArray<AnalysisModuleKey> missingKeys) {
             foreach (var (moduleName, _, isTypeshed) in missingKeys) {
                 var moduleResolution = isTypeshed ? interpreter.TypeshedResolution : interpreter.ModuleResolution;
-                moduleResolution.GetOrLoadModule(moduleName);
+                var m = moduleResolution.GetOrLoadModule(moduleName);
+                _queueTracker.Add(m.FilePath);
```
AlexanderSher: `moduleResolution.GetOrLoadModule` doesn't invalidate module analysis. It only ensures that module (if it is available) is loaded.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -190,6 +189,8 @@ public sealed class PythonAnalyzer : IPythonAnalyzer, IDisposable {
                     }
                 }
 
+                _queueTracker.Add(entry.Module.FilePath);
```
AlexanderSher: The only source of truth of how many modules are left to be analyzed is `walker.Remaining` of the walker that has `_version == walker.Version`. So this is the parameter that must be checked before every `_progress.ReportRemaining()` call, and that should be enough.

MikhailArkhipov: Basically I track `queued` vs `completed` pairs in the log for the same path and they seem to match.  Let me try your suggestion though.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -284,45 +285,46 @@ public sealed class PythonAnalyzer : IPythonAnalyzer, IDisposable {
         private void LoadMissingDocuments(IPythonInterpreter interpreter, ImmutableArray<AnalysisModuleKey> missingKeys) {
             foreach (var (moduleName, _, isTypeshed) in missingKeys) {
                 var moduleResolution = isTypeshed ? interpreter.TypeshedResolution : interpreter.ModuleResolution;
-                moduleResolution.GetOrLoadModule(moduleName);
+                var m = moduleResolution.GetOrLoadModule(moduleName);
```
AlexanderSher: `m` isn't needed

PR #763, opened 2019-03-15
Handle linting enable/disable option change

### LINKED ISSUE #754: Handle change in linter on/off without redoing analysis
Add separate entry to just re-run the linter

MikhailArkhipov:
Fixes #754


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.cs
```diff
@@ -78,6 +79,9 @@ internal sealed partial class ExpressionEval: IExpressionEvaluator {
             }
         }
 
+        public void RemoveDiagnostics(DiagnosticSource source) 
+            => _diagnostics = _diagnostics.Where(d => d.Source != source).ToList();
```
AlexanderSher: Consider using `ImmutableArray` here

src/LanguageServer/Impl/LanguageServer.Configuration.cs
```diff
@@ -0,0 +1,101 @@
...
+using Microsoft.Python.Core;
+using Microsoft.Python.LanguageServer.Protocol;
+using Newtonsoft.Json.Linq;
+using StreamJsonRpc;
+
+namespace Microsoft.Python.LanguageServer.Implementation {
+    public sealed partial class LanguageServer {
+        [JsonRpcMethod("workspace/didChangeConfiguration")]
+        public async Task DidChangeConfiguration(JToken token, CancellationToken cancellationToken) {
+            using (await _prioritizer.ConfigurationPriorityAsync(cancellationToken)) {
```
AlexanderSher: All this code is to handle one boolean parameter change?

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -162,6 +163,24 @@ public sealed class PythonAnalyzer : IPythonAnalyzer, IDisposable {
...
+                return;
+            }
+
+            var optionsProvider = _services.GetService<IAnalysisOptionsProvider>();
+            if (optionsProvider?.Options?.LintingEnabled != false) {
+                ((ExpressionEval)module.Analysis.ExpressionEvaluator).RemoveDiagnostics(DiagnosticSource.Linter);
+                var linter = new LinterAggregator();
+                linter.Lint(module.Analysis, _services);
+
+                var afterLinter = module.Analysis.Diagnostics.ToArray();
```
AlexanderSher: If linter has its own `DiagnosticSource`, why do we care about `Analysis.Diagnostics`?

MikhailArkhipov: Currently it adds to the analysis diagnostics. However... come to think of it - maybe it should be separate?

jakebailey:
If I have a file which only contains `x = y` (so `y` does is undefined), the message shows up. Then I change `python.linting.enabled` to `false`, and it goes away. But, if I change it back to `true`, then the linting message doesn't come back until I modify the file again. Is this the expected behavior?

PR #766, opened 2019-03-15
Split searchPaths by semicolon to work around extension before config cleanup

jakebailey:
Temporary fix until these configurations and the extension are cleaned up together.


src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -117,7 +117,7 @@ public sealed partial class Server : IDisposable {
             // TODO: multi-root workspaces.
             var rootDir = @params.rootUri != null ? @params.rootUri.ToAbsolutePath() : PathUtils.NormalizePath(@params.rootPath);
             var configuration = InterpreterConfiguration.FromDictionary(@params.initializationOptions.interpreter.properties);
-            configuration.SearchPaths = @params.initializationOptions.searchPaths;
+            configuration.SearchPaths = @params.initializationOptions.searchPaths.Select(p => p.Split(';')).SelectMany().Where(p => !string.IsNullOrEmpty(p)).ToList();
```
MikhailArkhipov: `.Split(';', StringSplitOptions.RemoveEmptyEntries)`

PR #796, opened 2019-03-19
Async adding workspace files to index

brianbok:
Instead of synchronously adding workspace files to the index, do it in a separate Task.

### DIFF
--- a/src/LanguageServer/Impl/Implementation/Server.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.cs
@@ -128,7 +128,7 @@ namespace Microsoft.Python.LanguageServer.Implementation {
                                             @params.initializationOptions.includeFiles,
                                             @params.initializationOptions.excludeFiles,
                                             _services.GetService<IIdleTimeService>());
-            _indexManager.IndexWorkspace(cancellationToken).DoNotWait();
+            _indexManager.IndexWorkspace().DoNotWait();
             _services.AddService(_indexManager);
             _disposableBag.Add(_indexManager);
 
--- a/src/LanguageServer/Impl/Implementation/Server.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.cs
@@ -128,7 +128,7 @@ namespace Microsoft.Python.LanguageServer.Implementation {
                                             @params.initializationOptions.includeFiles,
                                             @params.initializationOptions.excludeFiles,
                                             _services.GetService<IIdleTimeService>());
-            _indexManager.IndexWorkspace().DoNotWait();
+            _indexManager.IndexWorkspace(cancellationToken).DoNotWait();
             _services.AddService(_indexManager);
             _disposableBag.Add(_indexManager);
 
--- a/src/LanguageServer/Impl/Implementation/Server.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.cs
@@ -128,6 +128,7 @@ namespace Microsoft.Python.LanguageServer.Implementation {
                                             @params.initializationOptions.includeFiles,
                                             @params.initializationOptions.excludeFiles,
                                             _services.GetService<IIdleTimeService>());
+            _indexManager.IndexWorkspace().DoNotWait();
             _services.AddService(_indexManager);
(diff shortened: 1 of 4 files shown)


jakebailey:
If it fails, then what happens?

src/LanguageServer/Impl/Indexing/IndexManager.cs
```diff
@@ -57,18 +57,18 @@ internal class IndexManager : IIndexManager {
...
         public int ReIndexingDelay { get; set; } = DefaultReIndexDelay;
 
-        private void StartAddRootDir() {
-            foreach (var fileInfo in WorkspaceFiles()) {
-                if (ModulePath.IsPythonSourceFile(fileInfo.FullName)) {
-                    _symbolIndex.Parse(fileInfo.FullName);
+        public Task IndexWorkspace(CancellationToken ct = default) {
+            return Task.Run(() => {
+                foreach (var fileInfo in WorkspaceFiles()) {
+                    if (ModulePath.IsPythonSourceFile(fileInfo.FullName)) {
```
MikhailArkhipov: Doesn't seem to be using the `ct`. Should be doing then `ct.ThrowIfCancellationRequested` in the loop.

src/LanguageServer/Test/IndexManagerTests.cs
```diff
@@ -201,6 +204,7 @@ public class IndexManagerTests : LanguageServerTestBase {
             var pythonTestFilePath = context.FileWithXVarInRootDir();
 
             var indexManager = context.GetDefaultIndexManager();
+            await WaitForWorkspaceAddedAsync(indexManager);
```
MikhailArkhipov: All async methods should have cancellation

src/LanguageServer/Test/IndexManagerTests.cs
```diff
@@ -341,7 +345,7 @@ private class IndexTestContext : IDisposable {
         }
 
         private static async Task WaitForWorkspaceAddedAsync(IIndexManager indexManager) {
-            await indexManager.WorkspaceSymbolsAsync("", 1000);
+            await indexManager.IndexWorkspace();
```
MikhailArkhipov: How do we cancel this

brianbok:
> 
> 
> If it fails, then what happens?

You are right, it would still throw the exception to the user.

src/LanguageServer/Impl/Indexing/IndexManager.cs
```diff
@@ -57,18 +57,19 @@ internal class IndexManager : IIndexManager {
...
-            StartAddRootDir();
         }
 
         public int ReIndexingDelay { get; set; } = DefaultReIndexDelay;
 
-        private void StartAddRootDir() {
-            foreach (var fileInfo in WorkspaceFiles()) {
-                if (ModulePath.IsPythonSourceFile(fileInfo.FullName)) {
-                    _symbolIndex.Parse(fileInfo.FullName);
+        public Task IndexWorkspace(CancellationToken ct = default) {
```
jakebailey: You're adding a cancellation token, but I don't see that it's actually used anywhere; it's always `default`. Is there any reason to add it if nobody cancels it?

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -128,6 +128,7 @@ public sealed partial class Server : IDisposable {
                                             @params.initializationOptions.includeFiles,
                                             @params.initializationOptions.excludeFiles,
                                             _services.GetService<IIdleTimeService>());
+            _indexManager.IndexWorkspace(cancellationToken).DoNotWait();
```
jakebailey: I'm not sure that this `cancellationToken` is the right one to use, unfortunately. I believe this would be cancelled once the initialization occurs.

src/LanguageServer/Impl/Indexing/IndexManager.cs
brianbok: Even if it's not used, I'd have it there in case someone moves the call to IndexWorkspace

Issue #810, opened 2019-03-21
Focus is constantly switched to the Output window


DonJayamanne:
@luabud commented on [Wed Mar 20 2019](https://github.com/Microsoft/vscode-python/issues/4868)

Microsoft Python Language Server version: 0.2.29.0
Python extension version: 2019.3.5917-rc

If you watch the GIF below, you'll see the output window opens even before the package finishes installing in the terminal:

![LanguageServerOutputBug](https://user-images.githubusercontent.com/45497113/54725159-484a0c00-4b2b-11e9-95b7-dd6939b21c52.gif)

This is happening quite constantly and it can be very annoying when you need to use the terminal

jakebailey:
The exception is a bug which should be filed separately, but VS Code switching panels if we print something that's an error doesn't seem like something we have any control over (and probably want to make sure the error becomes visible).

DonJayamanne:
If thats the case, please do not flag it as an error or similar.
Please file an issue with VSC, as from a users perspective we don't want the panel stealing focus from the user (when they're in the terminal).

jakebailey:
I don't understand; this is an error in the language server that we want to know about and fix. No exceptions should be thrown unless there is something wrong; silently hiding them does not seem like a good idea.

luabud:
I opened #811 for the exception

DonJayamanne:
> this is an error in the language server that we want to know about and fix. 

I think we still need to pay attention to the over all `User Experience`

> No exceptions should be thrown unless there is something wrong; silently hiding them does not seem like a good idea.

* Over all, this is a PM decision
* Here's another problem:
	* Users open terminal, and pip install a package
	* Output panel appears and steals focus
	* User is confused and files a bug
	* Now they have to live with this poor UX for weeks til the issue is resolved.

> silently hiding them does not seem like a good idea.

We might want to re-evaluate how errors are surfaced to the user, in a non-intrusive way. But that's my opinion.

jakebailey:
I understand, but also know that we might not be able to control the level of that message. This is an exception which gets caught by the JSON-RPC library we use, which sends a JSON-RPC error response for the request being served, which the vscode-languageserver client then does something with. We might be able to modify the error code (similarly to how we modify the error body to work around other vscode-languageserver shortcomings), but we may be breaking the JSON-RPC specification by doing so.

MikhailArkhipov:
[...]

Issue #809, opened 2019-03-21
Latest version of LS fails to load


DonJayamanne:
* `0.2.30`
* Open the `pytest` repo in VS Code
* LS fails to start

<img width="792" alt="Screen Shot 2019-03-21 at 10 19 25 AM" src="https://user-images.githubusercontent.com/1948812/54771781-36b44300-4bc3-11e9-916e-27b49e50b09d.png">

jakebailey:
Is this with my PR to the extension? The language server is showing nothing in the search paths, which isn't good.

jakebailey:
Not that the PR causes this; the search path logging happens after we call our script to get the search paths.

jakebailey:
So I think what's happening is that the interpreter path specified doesn't exist. There's a log message missing which is printed after the interpreter is checked which is not in your output above.

DonJayamanne:
Oops, agreed no interpreter was selected (in fact a non-existent interpreter was selected), & the extension does display a warning message about that & i ignored it.

MikhailArkhipov:
Should now show clear error

PR #812, opened 2019-03-21
Log when interpreter is not found

jakebailey:
For #809.

If this happens we probably need some better handling, but we should at least print something.

I made it an error, which steals focus as in #810. I can change it to info or similar, but the LS will likely fail to initialize when there is no interpreter (at the moment).

### DIFF
--- a/src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
+++ b/src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
@@ -136,7 +136,7 @@ namespace Microsoft.Python.Analysis.Modules.Resolution {
         private async Task<IReadOnlyList<string>> GetInterpreterSearchPathsAsync(CancellationToken cancellationToken = default) {
             if (!_fs.FileExists(Configuration.InterpreterPath)) {
                 _log?.Log(TraceEventType.Warning, "Interpreter does not exist:", Configuration.InterpreterPath);
-                _ui?.ShowMessageAsync($"Interpreter does not exist; analysis will not be available. See the output panel for more information.", TraceEventType.Error);
+                _ui?.ShowMessageAsync(Resources.InterpreterNotFound, TraceEventType.Error);
                 return Array.Empty<string>();
             }
 
@@ -149,7 +149,7 @@ namespace Microsoft.Python.Analysis.Modules.Resolution {
                 return paths.MaybeEnumerate().Select(p => p.Path).ToArray();
             } catch (InvalidOperationException ex) {
                 _log?.Log(TraceEventType.Warning, "Exception getting search paths", ex);
-                _ui?.ShowMessageAsync($"An exception occured while discovering search paths; analysis will not be available. See the output panel for more information.", TraceEventType.Error);
+                _ui?.ShowMessageAsync(Resources.ExceptionGettingSearchPaths, TraceEventType.Error);
                 return Array.Empty<string>();
             }
         }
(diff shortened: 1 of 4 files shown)


src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
```diff
@@ -135,6 +135,7 @@ public MainModuleResolution(string root, IServiceContainer services)
 
         private async Task<IReadOnlyList<string>> GetInterpreterSearchPathsAsync(CancellationToken cancellationToken = default) {
             if (!_fs.FileExists(Configuration.InterpreterPath)) {
+                _log?.Log(TraceEventType.Error, "Interpreter does not exist:", Configuration.InterpreterPath);
```
AlexanderSher: We should probably add similar logging to the `try/catch` block below.

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
```diff
@@ -135,7 +135,7 @@ public MainModuleResolution(string root, IServiceContainer services)
 
         private async Task<IReadOnlyList<string>> GetInterpreterSearchPathsAsync(CancellationToken cancellationToken = default) {
             if (!_fs.FileExists(Configuration.InterpreterPath)) {
-                _log?.Log(TraceEventType.Error, "Interpreter does not exist:", Configuration.InterpreterPath);
+                _log?.Log(TraceEventType.Warning, "Interpreter does not exist:", Configuration.InterpreterPath);
```
MikhailArkhipov: We should pop up error saying that LS can't continue. I.e. the `IUIService.ShowMessageAsync` thing.

jakebailey: image

This alright? (For the exception, as an example.)

jakebailey: Only nit about my wording is "getting" feels weird.

MikhailArkhipov: IMO explicit message that interpreter is missing is more straightforward. Also, it should be an error as LS is not functional without it.

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
```diff
@@ -135,6 +135,8 @@ public MainModuleResolution(string root, IServiceContainer services)
 
         private async Task<IReadOnlyList<string>> GetInterpreterSearchPathsAsync(CancellationToken cancellationToken = default) {
             if (!_fs.FileExists(Configuration.InterpreterPath)) {
+                _log?.Log(TraceEventType.Warning, "Interpreter does not exist:", Configuration.InterpreterPath);
+                _ui?.ShowMessageAsync($"Interpreter does not exist; analysis may not be available. See the output panel for more information.", TraceEventType.Warning);
```
MikhailArkhipov: I'd make it big red error

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
```diff
@@ -145,7 +147,9 @@ public MainModuleResolution(string root, IServiceContainer services)
                 var paths = await PythonLibraryPath.GetSearchPathsAsync(Configuration, fs, ps, cancellationToken);
                 cancellationToken.ThrowIfCancellationRequested();
                 return paths.MaybeEnumerate().Select(p => p.Path).ToArray();
-            } catch (InvalidOperationException) {
+            } catch (InvalidOperationException ex) {
+                _log?.Log(TraceEventType.Warning, "Exception getting search paths", ex);
+                _ui?.ShowMessageAsync($"An exception occured while getting the search paths; analysis may not be available. See the output panel for more information.", TraceEventType.Warning);
```
MikhailArkhipov: And you can put it in resources as it is localizable thing (as opposed to tracing stuff).

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
jakebailey: OK, you probably missed my comment that it was an example of the caught exception; this is what you'd get if it were the interpreter missing:

image

MikhailArkhipov: I've seen it but it is a warning. Its not like "may not be available" :-) it definitely won't it and most probably we'll just crash.

PR #816, opened 2019-03-22
Fix #731: Autocomplete can't list all attributes in numpy.array even shape

### LINKED ISSUE #731: Autocomplete can't list all attributes in numpy.array even .shape
@kondongx commented on [Sat Mar 09 2019](https://github.com/Microsoft/vscode-python/issues/4698)

## Environment data

- VS Code version: 1.32.1 
- Extension version: 2019.2.5558
- OS and version: macOS Mojave 10.14.3
- Python version: 3.7.2; miniconda(python=3.6.3). Both installed by pyenv
- Type of virtual environment used: none venv; conda 
- Relevant/affected Python packages and their versions: NumPy 1.16.2

## Expected behaviour

When I type `np.array().`, I want autocomplete list `reshape()` or `shape` or many other attributes, just like PyCharm:
屏幕快照 2019-03-09 21 02 36
屏幕快照 2019-03-09 21 02 40
屏幕快照 2019-03-09 21 02 44

## Actual behaviour

But I got only `dtype` and some `__xxx__` in VSCode:
屏幕快照 2019-03-09 20 55 57
Note that yellow box, no more attributes.
屏幕快照 2019-03-09 21 01 12

## Steps to reproduce:
1. Type these code:
    ```
    import numpy as np
    np.array().
    ```
2. Then press `shift + enter` to trigger autocomplete.

## Logs
```

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)

Microsoft Python Language Server version 0.1.80.0
Initializing for /Users/kondongx/.pyenv/shims/python
```

---

@artemvalieiev commented on [Sun Mar 10 2019](https://github.com/Microsoft/vscode-python/issues/4698#issuecomment-471338743)

It also happens to me, Microsoft Language Server for Python incorrectly shows return type for numpy.array() as numpy.recarray instead of numpy.ndarray

AlexanderSher:
image

Fix contains of two parts:
- `__all__` variable is handled for the most simple case (single assignment)
- Stub is added for the `numpy.ndarray` and some of the `numpy` module functions


AlexanderSher:
More stubs can be added later if needed

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -15,22 +15,36 @@
...
+
         public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst ast)
-            : base(services, module, ast) {
+            : base(new ExpressionEval(services, module, ast)) {
             _stubAnalysis = Module.Stub is IDocument doc ? doc.GetAnyAnalysis() : null;
+            ExportedMemberNames = ImmutableArray<string>.Empty;
+        }
+
+        public override bool Walk(NameExpression node) {
+            if (Eval.CurrentScope == Eval.GlobalScope && node.Name == AllVariableName) {
```
MikhailArkhipov: You can also do `is IGlobalScope`

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -83,9 +97,18 @@ public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst
             SymbolTable.EvaluateAll();
             SymbolTable.ReplacedByStubs.Clear();
             MergeStub();
+
+            if (_allReferencesCount == 1 && GlobalScope.Variables.TryGetVariable(AllVariableName, out var variable) && variable?.Value is IPythonCollection collection) {
```
MikhailArkhipov: I guess @jakebailey and https://github.com/microsoft/python-language-server/issues/620 should be extending this.

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -15,22 +15,36 @@
...
+        private int _allReferencesCount;
+
         public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst ast)
-            : base(services, module, ast) {
+            : base(new ExpressionEval(services, module, ast)) {
             _stubAnalysis = Module.Stub is IDocument doc ? doc.GetAnyAnalysis() : null;
+            ExportedMemberNames = ImmutableArray<string>.Empty;
+        }
+
+        public override bool Walk(NameExpression node) {
```
jakebailey: This would mark uses; it's probably better to check for it in an assignment or augmented assignment node's left.

AlexanderSher: It can be also updated using `__all__.append(...)`, so I have to count simple uses as well.

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -83,9 +97,18 @@ public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst
             SymbolTable.EvaluateAll();
             SymbolTable.ReplacedByStubs.Clear();
             MergeStub();
+
+            if (_allReferencesCount == 1 && GlobalScope.Variables.TryGetVariable(AllVariableName, out var variable) && variable?.Value is IPythonCollection collection) {
+                ExportedMemberNames = collection.Contents
```
jakebailey: I'm not sure I really like the idea of exposing a public property which may or may not contain the true list of exports from a module. I think keeping this as an implementation detail for `GetMemberNames` makes more sense, especially since I'm going to be handling more things and better implementing the list of exports (and now that IC is back linking to the analyzer assembly).

There are three scenarios:

1. `__all__` was never assigned.
2. `__all__` was assigned in a way we can handle.
3. `__all__` was assigned in a way we can't handle.
4. `__all__` is defined by `__all__ = []` (a subset of 2).

With this PR, case 1 means `ExportedMemberNames` is empty; case 2 means `ExportedMemberNames` is not empty and `__all__` is an assigned variable in the scope; case 3 means `ExportedMemberNames` is empty but `__all__` is still potentially assigned; case 4 has an empty `ExportedMemberNames` and `__all__` is defined and has a value.

This feels a little inconsistent, whereas previously there was a single source of truth, `GetMemberNames`. Is there something `ExportedMemberNames` is doing that `GetMemberNames` was not, other than that it implies `_allReferencesCount == ` earlier than `GetMemberNames` would determine?

AlexanderSher: I'm not sure I understand the question. `IDocumentAnalysis` doesn't have `GetMemberNames` - it has `GlobalScope`, and `PythonModule.GetMemberNames()` is still a single source of truth.

We can make implementation of `PythonModule.GetMemberNames()` look like:
```cs
public virtual IEnumerable<string> GetMemberNames() 
    => Analysis.ExportedMemberNames;
```
and assign variables from GlobalScope to it in `ModuleWalker.Complete`, something like:
```cs
if (_allReferencesCount == 1 && GlobalScope.Variables.TryGetVariable(AllVariableName, out var variable) && variable?.Value is IPythonCollection collection) {
    ExportedMemberNames = collection.Contents
        .OfType<IPythonConstant>()
        .Select(c => c.GetString())
        .Where(s => !string.IsNullOrEmpty(s))
        .ToImmutableArray();
} else {
    ExportedMemberNames = GlobalScope.Variables
        .Where(v => !(v.Value?.GetPythonType() is PythonModule)
                    && !(v.Value?.GetPythonType().DeclaringModule is TypingModule && !(this is TypingModule)))
        .Select(v => v.Name)
        .ToImmutableArray();
}
```
I didn't think much about it cause there is #620 and current implementation is temporary, but we definitely can make `ExportedMemberNames` a list of exported variables for all cases.

MikhailArkhipov: Yeah, my point in comments was that #620 is primary and this implementation is temporary and basic.

PR #819, opened 2019-03-24
Find all references and rename

### LINKED ISSUE #454: Wrong file referenced in Find/List All References and Rename Symbol
## Environment data

- VS Code version: 1.29.1
- Language server version: 0.1.65.0
- OS and version: Win 10 x64 
- Python version (& distribution if applicable, e.g. Anaconda): Anaconda 3.7.1 x64
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): conda
- "python.jediEnabled": false
## Steps to reproduce:
- Create 2 files:
``` python
# foo.py

from bar import baz


class spam:
    __bug__ = 0


def eggs(ham: spam):
    return baz(ham.__bug__)

```
``` python
# bar.py

# comments ########################################
# more commments ##################################
# even more comments ##############################


def baz(quux):
    pass
```
- Hover over `__bug__` and Find All References
- Completely unrelated file `bar.py` is referenced at offset 112 which is exactly the offset of `__bug__` in `foo.py` at line 11
reference_bug
- List All References behavior is identical
- Trying to rename F2  `__bug__` to `some_random_variable` works fine for `foo.py` but changes `bar.py`:
``` python
# bar.py

# comments ########################################
# more commments ##############################some_random_variable even more comments ##############################


def baz(quux):
    pass

```

### LINKED ISSUE #477: Rename variable sometimes fails when using the LanguageServer
@cipri-tom commented on [Thu Dec 06 2018](https://github.com/Microsoft/vscode-python/issues/3574)

## Environment data

- VS Code version: 1.29.1
- Extension version (available under the Extensions sidebar): 2018.11.0
- OS and version: macOS Mojave 10.14.1
- Python version (& distribution if applicable, e.g. Anaconda): python 3.6.4 64-bit conda
- Relevant/affected Python packages and their versions:
   - Microsoft Python Language Server version 0.1.60.0

## Steps to reproduce:
1. Make new file, `test.py`
2. paste the following:
```python
def iadd_no_overflow(a, b):
    """Perform `a += b` on integer numpy arrays without overflowing"""
    dest += src
    dest[dest < src] = np.iinfo(dest.dtype).max
```
Note the file is dirty (not saved). The same happens if the file is saved.

3. Place cursor on `dest` and hit <kbd>F2</kbd> or choose `Rename symbol`. Enter `a` as name.

## Expected behaviour
All occurences of `dest` are changed into `a`

## Actual behaviour

Got a tooltip `Cannot rename`, and an error in the Output window:
## Logs
Output for `Python` in the `Output` panel (`View`→`Output`, change the drop-down the upper-right of the `Output` panel to `Python`)

```
Microsoft Python Language Server version 0.1.60.0
Initializing for /Users/ciprian/miniconda3/bin/python
[Error - 12:12:43 PM] Request textDocument/rename failed.
  Message: Cannot rename
  Code: -32000 
   at Microsoft.Python.LanguageServer.Implementation.Server.Rename(RenameParams params, CancellationToken cancellationToken)
   at Microsoft.Python.LanguageServer.Implementation.LanguageServer.Rename(JToken token, CancellationToken cancellationToken)
```

Output from `Console` under the `Developer Tools` panel (toggle Developer Tools on under `Help`):

Nothing. I mean, there are some `ERR` but from before. I put a `console.log` and then reproduce the error and nothing new appears in the console.

### LINKED ISSUE #577: LS Rename using new analysis

### LINKED ISSUE #645: Peeking References for a extended class
@manutero commented on [Thu Feb 21 2019](https://github.com/Microsoft/vscode-python/issues/4481)

## Environment data

- VS Code version:  1.31.1 (1.31.1)
- Extension version (available under the Extensions sidebar): 1b8e8302e405050205e69b59abb3559592bb9e60
- OS and version: macOS Mojave 10.14.3
- Python version (& distribution if applicable, e.g. Anaconda): 3.7.1 64-bit
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): virtualenv 
- Relevant/affected Python packages and their versions: 
    - `Python` 2019.1.0 (microsoft)
    - `python-snippets` 0.2.0 (cstrap)

## Expected behaviour

Peeking References for a Python class might show any class extending it. 

```
Class A():

...

Class B(A): # A is referenced here
```

## Actual behaviour

Picking references of a Python class *does not* list any ussage of that class as parent class.

## Steps to reproduce:

___Prerequisites___

In the code base, we have at least two classes, A & B. B extends A.

***

1. Cursor over Python class A
2. Peek references (via command palette or shortcut or right clicking)
3. A ussage from B class is *not* shown

## Logs

Output for `Python` in the `Output` panel (`View`→`Output`, change the drop-down the upper-right of the `Output` panel to `Python`)

None.

Output from `Console` under the `Developer Tools` panel (toggle Developer Tools on under `Help`; turn on source maps to make any tracebacks be useful by running `Enable source map support for extension debugging`)

```
2console.ts:134 [Extension Host] Python Extension: Cached data exists getEnvironmentVariables, /Users/martero/workspace/tuenti/flow/flow_core/flow_core/tasks/basetask.py
```

## Relevant user config

```
  "python.linting.lintOnSave": true,
  "python.linting.pep8CategorySeverity.E": "Warning",
  "python.linting.pylintEnabled": true,
  "python.venvPath": "~/.virtualenvs",
  // "python.jediEnabled": false, // TRIED BOTH, TRUE & FALSE
```

### LINKED ISSUE #699: Find All References in new LS

### LINKED ISSUE #729: Renaming file will burn down LSP
@hujianxin commented on [Sun Mar 10 2019](https://github.com/Microsoft/vscode-python/issues/4703)

## Environment data

- VS Code 1.32.1: 
- Extension version (available under the Extensions sidebar): 2019.2.5558 (6 March 2019)
- OS and version: Ubuntu 16.04
- Python version (& distribution if applicable, e.g. Anaconda): 3.7.1
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): pipenv 

## Expected behaviour

Rename a python file in VSCode sidebar, goto definition of LSP works fine

## Actual behaviour

LSP does not work, and a progress bar above editor is always running

## Steps to reproduce:
1. XXX
touch a.py, b.py

import b.py from a.py

change b.py to c.py

### LINKED ISSUE #829: mspls can not handle this: [year, month] = date.split("-")
If I get variable `year` and `month` from a list like:  [year, month] = date.split("-")

then, vscode will complain `Undefined variable: 'year'`

### LINKED ISSUE #830: Undefined variables in list comprehensions after latest update
After the latest update to the Python extension that dropped today we're seeing a bunch of undefined variables showing up in list comprehensions. I'm assuming it's to do with some update in the language server. This snippet of code demonstrates it:

```python
def func_a(value):
    return value


def func_b():
    list_a = [1, 2, 3, 4]

    return [func_a(value=elem) for elem in list_a]
```

In the above code, `elem` in the last line will be undefined inside the function call to `func_a`.

Version of the extension used: 2019.3.6139.

### LINKED ISSUE #843: Variable appears in completion outside its scope
image

### LINKED ISSUE #865: NullReferenceException in Server.Completion
Seeing this one at the top now.

```
   at Microsoft.Python.LanguageServer.Implementation.Server.Completion(CompletionParams params, CancellationToken cancellationToken)
   at Microsoft.Python.LanguageServer.Implementation.LanguageServer.Completion(JToken token, CancellationToken cancellationToken)
```

MikhailArkhipov:
Fixes #699 
Fixes #454
Fixes #577
Fixes #645
Fixes #477 
Fixes #829
Fixes #729
Fixes #830
Fixes #843
Fixes #865

- Adds definition and reference tracking to `ILocatedMember`.
- Eliminates creation of `LocationInfo` up front instead relying of remember originating expressions.
- Add `LocatedMember` base to python types.
- Track 'parent' definition of imported variables so there is definition in the current file (the import statement) and there is parent definition in the imported module (to go to on F12 in the import statement definition). There may be a chain of parents.
- Instances do not track location.
- Open, parse and inspect imports for candidates.
- Load and analyze candidate files.
- Analysis is now disposable and upon dispose cleans references to the module variables.
- Tests.

Although the change is large, 90% is plumbing ILocatedMember to types. The remaining is calling `ILocatedMember.AddReference` in strategic places.


jakebailey:
In this test case, going to the definition of `compile` goes to the wrong place in the file.

```python
import re

x = re.compile(r'hello', re.IGNORECASE)
```

src/Analysis/Ast/Impl/Analyzer/Handlers/AssignmentHandler.cs
```diff
@@ -44,10 +45,10 @@ internal sealed class AssignmentHandler : StatementHandler {
                 value = Eval.UnknownType;
             }
 
-            if (node.Left.FirstOrDefault() is TupleExpression lhs) {
+            if (node.Left.FirstOrDefault() is SequenceExpression seq) {
```
brianbok: What happens if you have several complicated expressions on the left side of an assignment?
I think you can't look only at node.Left.First()
((x, r), y, z) = (1, 1), '', ''

src/Analysis/Ast/Impl/Analyzer/Handlers/AssignmentHandler.cs
```diff
@@ -60,26 +61,18 @@ internal sealed class AssignmentHandler : StatementHandler {
             foreach (var ne in node.Left.OfType<NameExpression>()) {
```
brianbok: Consider just " (x, (y)) = 1, '' "

src/Analysis/Ast/Test/AssignmentTests.cs
```diff
@@ -160,7 +160,16 @@ class test():
...
 ";
-            var analysis = await GetAnalysisAsync(code, PythonVersions.LatestAvailable3X);
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Should().HaveVariable("x").And.HaveVariable("y").And.HaveVariable("z");
+        }
+
+        [TestMethod, Priority(0)]
+        public async Task TupleSingleValue() {
+            const string code = @"
+(x, (y, (z))) = 1,
```
brianbok: Did you choose this example for a specific reason?
I would use a working python example like "(x, (y, (z))) = (1, (1., ''))" and check for the types returned by analysis

MikhailArkhipov: The re is another one added below. This one has error in the RHS and a single value so it is a separate example

src/Analysis/Ast/Impl/Analyzer/Handlers/SequenceExpressionHandler.cs
```diff
@@ -93,6 +92,22 @@ internal sealed class SequenceExpressionHandler : StatementHandler {
             return names;
         }
 
+        private static IEnumerable<IMember> ValuesFromSequenceExpression(IEnumerable<Expression> items, ExpressionEval eval) {
+            var names = new List<IMember>();
```
brianbok: values instead of names?

jakebailey:
If you have master merged, you should also remove the references/rename item from the known issues list in `TROUBLESHOOTING.md`. If not, we can just remove it in another PR.

jakebailey:
In a folder with just these two files:

`a.py`:
```python
foo = 123
bar = "xyz"
baz = 1.23
fux = 1j

def cool():
    """This is cool"""
    return 123

x = ["foo"]
y = ["bar"]
z = ["fux"]

__all__ = x + y + z
__all__ += ["cool"]
```

`b.py`
```python
from .a import cool

cool()
```

`cool` is shown to have 19 some references in `b.py`, many of which are broken.

src/LanguageServer/Impl/Program.cs
```diff
@@ -13,7 +13,7 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
-// #define WAIT_FOR_DEBUGGER
+#define WAIT_FOR_DEBUGGER
```
jakebailey: Don't forget to remove this.

MikhailArkhipov: Removed

tiangolo:
I understand this fixed several issues like https://github.com/Microsoft/python-language-server/issues/577, but I guess it hasn't been released yet, right?

Where can I check releases to know when it is available (and if it is the case that I have a problem in my own environment)?

MikhailArkhipov:
It is released in the daily channel, but not in beta or stable yet as we'd like a bit more testing.

tiangolo:
Ok, thank you.

PR #827, opened 2019-03-26
Named Expressions

### LINKED ISSUE #580: Support for PEP 572 assignment expressions
Python 3.8 ([release planned for October](https://www.python.org/dev/peps/pep-0569/)) is adding support for "assignment expressions" (see [PEP 572](https://www.python.org/dev/peps/pep-0572/)).  I just wanted to make sure the syntax is supported by the language server.

brianbok:
Updates parser to recognize named expressions. Syntax specified in PEP 572.
However, this doesn't implement TargetScopeErrors (part of the same PEP).

Closes #580 .


src/LanguageServer/Test/HoverTests.cs
```diff
@@ -224,6 +224,18 @@ class Box(Generic[_T]):
             AssertHover(hs, analysis, new SourceLocation(6, 8), @"some: str", new SourceSpan(6, 8, 6, 12));
         }
 
+        [TestMethod, Priority(0)]
+        public async Task AssignmentExpressions() {
+            const string code = @"
+a = 5
```
jakebailey: You probably don't want this extra assignment; if the analysis were to skip over the `:=` then the test may pass because the normal assignment caused `a` to be an `int`.

jakebailey:
You'll also want to add the token to the line formatter along with a test. Theoretically, the default case will add spaces around the `:=` so it will work fine, but the code explicitly looked at every token so would be better off still doing that.

jakebailey:
To be clear, this really just means adding `:=` to the normal operators case, and a few lines for a test to make sure that `:=` gets spaces around it.

Issue #832, opened 2019-03-27
High memory usage


gramster:
@suiahaw commented on [Tue Mar 26 2019](https://github.com/Microsoft/vscode-python/issues/4952)

Issue Type: <b>Bug</b>

### Some questions about the python plugin encountered when building python library indexes
I strongly hope that **the python plugin does not `read the information into memory in real time `when creating the python library index**, **but `instead saves the index file`** in order to speed up the time and reduce memory overhead. 

The python library is really too big. Sometimes I have to wait a few minutes for writing a small amount of code.

Extension version: 2019.3.6139
VS Code version: Code 1.32.3 (a3db5be9b5c6ba46bb7555ec5d60178ecc2eaae4, 2019-03-14T23:43:35.476Z)
OS version: Windows_NT x64 10.0.17763

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|Intel(R) Core(TM) i5-7200U CPU @ 2.50GHz (4 x 2712)|
|GPU Status|2d_canvas: enabled<br>checker_imaging: disabled_off<br>flash_3d: enabled<br>flash_stage3d: enabled<br>flash_stage3d_baseline: enabled<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>native_gpu_memory_buffers: disabled_software<br>rasterization: enabled<br>surface_synchronization: enabled_on<br>video_decode: enabled<br>webgl: enabled<br>webgl2: enabled|
|Memory (System)|15.86GB (8.60GB free)|
|Process Argv||
|Screen Reader|no|
|VM|46%|

</details>

---

@algo99 commented on [Wed Mar 27 2019](https://github.com/Microsoft/vscode-python/issues/4952#issuecomment-477034361)

I can confirm, on Windows 10.

After update (yesterday) of Python Language Server and/or vscode_python extension VSCode  not more usable. See the disk I/O and memory consumption:

![vscode_python_language_server](https://user-images.githubusercontent.com/28513219/55060755-7c18ba80-5072-11e9-8b79-5e5566d9dd69.jpg)

The extension proceeds after VSCode start:
> Analyzing in background, 6879 items left ...

and this process does not stop.
I tried several times to restart VSCode, but it does not help.

---

@algo99 commented on [Wed Mar 27 2019](https://github.com/Microsoft/vscode-python/issues/4952#issuecomment-477038799)

Just noted in Python  output pane:
```log
...
[...]

PR #835, opened 2019-03-28
[WIP] Limit concurrency

MikhailArkhipov:
For #832 

Temporary solution, offers limit to ~2.5GB from 10GB on tensorflow without significant loss of speed. @jakebailey - could you try this out please.


MikhailArkhipov:
Merging per offline discussion with @AlexanderSher

PR #836, opened 2019-03-28
Changing Expression's nodenames

brianbok:
All nodes have a default NodeName implementation, which is just the type's name.
Nodenames can be used to report syntax errors. (See for example ExpressionWithAnnotation.cs) 
I will use them to report syntax errors related to named expressions.

This PR implements missing nodenames so that errors match what CPython prints.

### DIFF
--- a/src/Parsing/Impl/Ast/BackQuoteExpression.cs
+++ b/src/Parsing/Impl/Ast/BackQuoteExpression.cs
@@ -26,6 +26,8 @@ namespace Microsoft.Python.Parsing.Ast {
 
         public Expression Expression { get; }
 
+        public override string NodeName => "string conversion";
+
         public override IEnumerable<Node> GetChildNodes() {
             if (Expression != null) yield return Expression;
         }
--- a/src/Parsing/Impl/Ast/CallExpression.cs
+++ b/src/Parsing/Impl/Ast/CallExpression.cs
@@ -70,6 +70,8 @@ namespace Microsoft.Python.Parsing.Ast {
 
         internal override string CheckDelete() => "can't delete function call";
 
+        public override string NodeName => "function call";
+
         public override IEnumerable<Node> GetChildNodes() {
             if (Target != null) yield return Target;
             foreach (var arg in Args) {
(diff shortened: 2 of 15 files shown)


jakebailey:
Out of curiosity, if these are from what CPython prints, are there any that already exist in the parser that are "wrong"?

brianbok:
> 
> 
> Out of curiosity, if these are from what CPython prints, are there any that already exist in the parser that are "wrong"?

I didn't find any "wrong" Nodename. 
I should leave the reference somewhere. They match CPython/Python/ast.c get_expr_name

brianbok:
> 
> 
> > Out of curiosity, if these are from what CPython prints, are there any that already exist in the parser that are "wrong"?
> 
> I didn't find any "wrong" Nodename.
> I should leave the reference somewhere. They match CPython/Python/ast.c get_expr_name
> See this commit
> [python/cpython@d8b3a98](https://github.com/python/cpython/commit/d8b3a98c9098c66a714fd5593e1928af0ffbc631)

> 
> 
> Out of curiosity, if these are from what CPython prints, are there any that already exist in the parser that are "wrong"?

Now that you mention it, ConstantExpression special cases are not covered by this PR.
A ConstantExpr holding False, should have "False" as Nodename

jakebailey:
What is `NodeName` used for, in general? Making the node's name dependent on its value rather than the type seems weird, but I know that True/False/None/Ellipsis are sort of special.

brianbok:
> 
> 
> What is `NodeName` used for, in general? Making the node's name dependent on its value rather than the type seems weird, but I know that True/False/None/Ellipsis are sort of special.

It's only being used by:
* SymbolIndexWalker, to add symbols inside "scope" named by the comprehension type (dict comprehension, set comprehension)
*  Reporting errors, as in Expression.cs, CheckDelete default implementation:  
CheckDelete() => "can't delete " + NodeName;

jakebailey:
Okay. I'd be more inclined to have just the changes in this PR for now and maybe edit it for those special constants later on if we need it. (For example, check out ConstantExpression and how a `null` value changes all sorts of strings.)

PR #863, opened 2019-04-01
__all__ evaluation

### LINKED ISSUE #619: __all__ not being respected from an `import *`

### LINKED ISSUE #620: Handle complex __all__ assignments

jakebailey:
Fixes #620.
Fixes #619.

This adds support for concatenating lists using `+`, doing `+=` on `__all__`, and calling `append` and `extend` on `__all__`. If something goes wrong (an unsupported operation on `__all__` or some unsupported value), then the old behavior continues to be used. I don't track uses of `__all__` indirectly (i.e. passing `__all__` to something that modifies it), only direct actions. If `__all__` is in a more complicated lvar (like `__all__, foo = ...`), then it is ignored. This can be improved later on when we fix up our multiple assignment issues.

This works well for Django models (see #620), but `numpy`'s import cycles prevent this from having an effect, so the old behavior will be used.

~Tests are WIP.~ I'll need to rebase/merge master when the refs per gets merged.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
@@ -13,8 +13,11 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
+using System;
+using System.Collections.Generic;
 using Microsoft.Python.Analysis.Modules;
 using Microsoft.Python.Analysis.Types;
+using Microsoft.Python.Analysis.Types.Collections;
 using Microsoft.Python.Analysis.Values;
 using Microsoft.Python.Parsing;
 using Microsoft.Python.Parsing.Ast;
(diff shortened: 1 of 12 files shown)


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
```diff
@@ -123,6 +126,22 @@ internal sealed partial class ExpressionEval {
                 return left;
             }
 
+            if (binop.Operator == PythonOperator.Add
```
MikhailArkhipov: May be worth making a separate method like `HandleListAdd`

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -47,6 +51,105 @@ public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst
...
+            Eval.LookupNameInScopes(AllVariableName, out var scope, LookupOptions.Normal);
+            if (scope == null) {
+                return;
+            }
+
+            var loc = Eval.GetLoc(declNode);
+
+            var allContents = (scope.Variables[AllVariableName].Value as IPythonCollection)?.Contents;
+
+            var contents = new List<IMember>(allContents ?? Array.Empty<IMember>());
```
MikhailArkhipov: This three lines appear to repeat, perhaps method or an extension?

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
jakebailey: With the collection type stuff pulled out into its own function and some tweaks to shorten this, I think it's probably short enough to leave where it is (just a couple lines).

PR #876, opened 2019-04-03
Send telemetry event on analysis complete

### LINKED ISSUE #851: Add telemetry for analysis time.
As the title says, we should be gathering this in our telemetry so we can track improvements over time.

jakebailey:
Closes #851.

Using the same definition of "complete" as the log message. If it did fewer than 100 items, then nothing is sent to prevent spamming for every small change. Open to suggestions for another arbitrary condition instead.

I'd include the system memory, but that'd require a bunch of code to make it work cross platform (or waiting for dotnet/corefx#22660).

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerSession.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerSession.cs
@@ -148,36 +148,51 @@ namespace Microsoft.Python.Analysis.Analyzer {
 
             var elapsed = stopWatch.Elapsed.TotalMilliseconds;
 
-            if (_telemetry != null && remaining == 0 && originalRemaining > 100) {
-                double privateMB;
-                double peakPagedMB;
+            SendTelemetry(elapsed, originalRemaining, remaining, walker.Version);
+            LogResults(elapsed, originalRemaining, remaining, walker.Version);
+        }
 
-                using (var proc = Process.GetCurrentProcess()) {
-                    privateMB = proc.PrivateMemorySize64 / 1e+6;
-                    peakPagedMB = proc.PeakPagedMemorySize64 / 1e+6;
-                }
-
-                var e = new TelemetryEvent() {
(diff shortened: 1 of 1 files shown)


src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerSession.cs
```diff
@@ -143,13 +145,37 @@ internal sealed class PythonAnalyzerSession {
                     _progress.ReportRemaining(walker.Remaining);
                 }
 
+                var elapsed = stopWatch.Elapsed.TotalMilliseconds;
```
AlexanderSher: Can we move it outside of the `finally` block and put into its individual method?

jakebailey: What do you mean? This needs to be accessed after `stopWatch.Stop()`, right?

AlexanderSher: It will be stopped after `finally`. Code in `finally` blocks is always executed (even in case of `Thread.Abort`), so I'd prefer not to call heavy methods from it.

jakebailey: Oh, you meant to move it below, not above, I see.

jakebailey: I moved things outward; let me know if that's what you wanted.

AlexanderSher: Yep. And move that code into separate methods, something like `SendTelemetry` & `LogResults`.

jakebailey: I moved them out, including their checks. I can move the if statements back out, if you'd prefer.

PR #878, opened 2019-04-03
LineFormatter: Spaces around ':='

brianbok:
We treat ':=' the same as normal assignment operator.

### DIFF
--- a/src/LanguageServer/Impl/Formatting/LineFormatter.cs
+++ b/src/LanguageServer/Impl/Formatting/LineFormatter.cs
@@ -197,7 +197,6 @@ namespace Microsoft.Python.LanguageServer.Formatting {
                     case TokenKind.BitwiseAndEqual:
                     case TokenKind.BitwiseOrEqual:
                     case TokenKind.ExclusiveOrEqual:
-
                         AppendTokenEnsureWhiteSpacesAround(builder, token);
                         break;
 
--- a/src/LanguageServer/Impl/Formatting/LineFormatter.cs
+++ b/src/LanguageServer/Impl/Formatting/LineFormatter.cs
@@ -180,7 +180,6 @@ namespace Microsoft.Python.LanguageServer.Formatting {
                         break;
 
                     case TokenKind.Assign:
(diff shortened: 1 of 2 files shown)


src/LanguageServer/Impl/Formatting/LineFormatter.cs
```diff
@@ -180,6 +180,7 @@ public class LineFormatter {
                         break;
 
                     case TokenKind.Assign:
+                    case TokenKind.ColonEqual:
```
jakebailey: It's functionally the same, but given `:=` is an operator which appears in an expression like `+`, I would probably have put it with the other normal operators.

brianbok: Moved it

jakebailey: You moved it with the augmented assignment operators; I meant to move it with "Operators" a few lines down.

brianbok: Fixed

src/LanguageServer/Impl/Formatting/LineFormatter.cs
```diff
@@ -197,6 +197,9 @@ public class LineFormatter {
                     case TokenKind.BitwiseAndEqual:
                     case TokenKind.BitwiseOrEqual:
                     case TokenKind.ExclusiveOrEqual:
+                    // Used in expressions as an operator
```
jakebailey: Drop the comment; it doesn't add anything (because all of these are operators used in expressions...)

src/LanguageServer/Impl/Formatting/LineFormatter.cs
```diff
@@ -197,6 +197,7 @@ public class LineFormatter {
                     case TokenKind.BitwiseAndEqual:
                     case TokenKind.BitwiseOrEqual:
                     case TokenKind.ExclusiveOrEqual:
+
```
jakebailey: There's a random extra empty line added here

Issue #248, opened 2018-10-12
Fix \r\n assumptions in round-trip tests.


jakebailey:
The round trip tests assume that the C# file they're written in (thanks to `@""` strings) is the same as `Environment.NewLine`. The file's line endings are `\r\n` since the file was created on Windows, so the tests fail on platforms other than Windows.

Test cases should be normalized, and potentially tested against both `\n` and `\r\n`.

jakebailey:
In this case, I'm specifically referring to the round-trip tests, which do a character-for-character check to see that the tokenizer and parser will give back exactly what they were given (so long as they were configured to be in verbatim mode).

Honestly, reading over ParserRoundTripTest makes me want to just trim the whole thing down; there's a decent amount of code there that's just for pretty printing string differences that would be better written as a single line of FluentAssertions.

MikhailArkhipov:
Ah, OK. I thought about baselines, I believe RTVS normalized baseline files to just LF.

jakebailey:
They could also be converted to DataTests at the same time, though there are _so many_ test cases that it might look really silly in the test explorer. 😛

jakebailey:
Refocusing this to be about testing both LF and CRLF with CodeFormattingOptions, since #256 will effectively "fix" the tests not passing.

MikhailArkhipov:
Doesn't have to be done this sprint

PR #885, opened 2019-04-04
Remove check for indexes in parser tests

brianbok:
Tests that check errors reported by parser were expecting specific index numbers.
Index numbers differ by OS (based on usage of CRLF/LF, etc) so I removed the check fixing the tests for other platforms. Notice that index numbers remain there unused.
This prepares us for a different commit which would delete ErrorInfo and all those numbers from our tests.
I also added tokenizer tests for different newlines.


brianbok:
I'm not sure if this closes #248 or you had something different in mind when you filed that issue @jakebailey .

jakebailey:
No, not particularly. That has more to do with the roundtrips being the same when the code is reprinted with different newline styles.

PR #886, opened 2019-04-04
Add IsExact to collection interface, check when deciding to use __all__

### LINKED ISSUE #881: No exports from tensorflow
From #869:

image

tensorflow's `__all__` should be getting marked as unusable, but it may be something else.

jakebailey:
Fixes #881.

This adds `IsExact` to `IPythonCollection`, which is true when `Contents` contains an exact representation of the contents, meaning that the values are set and are known. This boolean is propagated around, and used when looking up `__all__`.

Tensorflow broke because it assigns `__all__` via a comprehension; the comprehension evaluator doesn't actually populate a real collection, instead producing a collection which contains a single member. This could be improved in the future, but by not setting `IsExact` true when that collection is produced, that incomplete value won't be used when looking at `__all__`.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
@@ -98,7 +98,7 @@ namespace Microsoft.Python.Analysis.Analyzer {
             if (node.Args.Count == 0) {
                 return;
             }
-            
+
             var arg = node.Args[0].Expression;
             var v = Eval.GetValueFromExpression(arg);
             if (v == null) {
--- a/src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
@@ -76,14 +76,14 @@ namespace Microsoft.Python.Analysis.Analyzer {
             }
 
             var rightVar = Eval.GetValueFromExpression(node.Right);
-            var rightContents = (rightVar as IPythonCollection)?.Contents;
+            var right = rightVar as IPythonCollection;
 
-            if (rightContents == null) {
+            if (right == null) {
                 _allIsUsable = false;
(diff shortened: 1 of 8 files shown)


jakebailey:
This too will conflict with the refs PR due to the LocationInfo uses.

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -98,42 +98,44 @@ public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst
             if (node.Args.Count == 0) {
                 return;
             }
-
-            IReadOnlyList<IMember> contents = null;
-            var v = Eval.GetValueFromExpression(node.Args[0].Expression);
+
```
brianbok: nit: extra space?

jakebailey: Oops.

src/Analysis/Ast/Impl/Values/Collections/PythonCollection.cs
```diff
@@ -72,5 +75,7 @@ internal class PythonCollection : PythonInstance, IPythonCollection {
                     return 0;
             }
         }
+
+        public bool IsExact { get; protected set; }
```
AlexanderSher: Why `protected set`? Looks like we only set it in constructor

jakebailey: I made it have the same level as `Contents` did, so it could be set in the same places, but that one seems like it could drop `set` as well without issue if you wanted me to make that change.

jakebailey: Contents needed `protected set`, but I've dropped `protected set` from `IsExact` (but we can add it back later if we come up with some more interesting use case).

PR #890, opened 2019-04-05
Perf improvement. Delays creation of the walker until it is actually required.

AlexanderSher:


src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerSession.cs
```diff
@@ -161,20 +163,20 @@ internal sealed class PythonAnalyzerSession {
                 return;
             }
 
-            double privateMB;
-            double peakPagedMB;
+            double privateMb;
```
jakebailey: `MB` is a unit, hence why I didn't do `Mb`.

MikhailArkhipov: +1

AlexanderSher: Sorry, autoformatting :(

PR #913, opened 2019-04-09
Limit size of exact collections

### LINKED ISSUE #877: probablepeople library causes LS to peg a CPU core
@jakebailey figured out that this is caused by the large number of dictionary entries in that library

jakebailey:
Fixes #877.

Limit the size of exact collections to 1000 items (when they are created from the expressions). This is set high in order to not affect potentially large `__all__` definitions.

probablepeople has a dictionary with more than 150,000 key/value pairs, which takes a very long time to process; 1000 doesn't take very long.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
@@ -23,6 +23,8 @@ using Microsoft.Python.Parsing.Ast;
 
 namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
     internal sealed partial class ExpressionEval {
+        private const int MaxCollectionSize = 1000;
+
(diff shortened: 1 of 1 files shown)


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
```diff
@@ -23,6 +23,8 @@
 
 namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
     internal sealed partial class ExpressionEval {
+        private const int MaxCollectionSize = 1000;
```
MikhailArkhipov: This may be made settable for tests for you can set it to 5 and make a test for it. Up yo you.

jakebailey: Yeah, this is somewhat deep in the analysis so there's no good direct way to test it (given you need to provide it a module and such). I'll just leave it as-is for now.

PR #919, opened 2019-04-10
Fix #914: Multiple fast opening of files causes GetAnalysisAsync to return old analysis

### LINKED ISSUE #914: Multiple fast opening of files causes GetAnalysisAsync to return old analysis
Or `WaitForCompleteAnalysisAsync` to hang. Continuation of #577 and #699

AlexanderSher:


src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -212,7 +212,7 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
             lock (AnalysisLock) {
                 LoadContent(content, version);
 
-                var startParse = ContentState < State.Parsing && _parsingTask == null;
+                var startParse = ContentState < State.Parsing && (_parsingTask == null || version > 0);
```
MikhailArkhipov: OK to run multiple?

Issue #849, opened 2019-04-01
Language Server does not work with PYTHONPATH=./


DonJayamanne:
@DonJayamanne commented on [Sun Mar 31 2019](https://github.com/Microsoft/vscode-python/issues/5036)

* Open the [behave](https://github.com/behave/behave) repo
* Use python 3.7 
* OS Mac
* Enable the following in `settings.json`
```js
	"python.analysis.downloadChannel": "daily",
	"python.analysis.logLevel": "Trace",
```
* Add a `.env` file with the following contents:
```
PYTHONPATH=./
```

## **Delete the above `.env` file then it works**

**None of the Language Server functionality works**

<img width="1127" alt="Screen Shot 2019-03-31 at 7 32 02 PM" src="https://user-images.githubusercontent.com/1948812/55300341-d8216d00-53eb-11e9-95ff-a3b277d1fe25.png">

vivlim:
I encountered this with a project that had two adjacent delimiters in PYTHONPATH, i.e. `PYTHONPATH=::`.
There was no clear way to tell what was happening, all I could observe was that I had a cpu core pinned by the language server & no completions were working. That core continues to be 100% utilized until the vscode window closes. I've observed it running for at least 20 minutes of cpu time, though it's probably run for longer.

(I ended up having adjacent delimiters because I was trying have a single PYTHONPATH that would work across windows & linux which both use different os.pathsep.)

jakebailey:
@vivlim Did you also encounter this on Mac/Linux?

vivlimmsft:
@jakebailey yes, I saw this on an Ubuntu 18.04 machine

jakebailey:
This should be fixed in 0.2.55+, if you'd like to retest.

vivlimmsft:
Sure, I'll try it tonight. Thanks!!

vivlim:
@jakebailey On 0.2.58, the language server no longer pins a cpu core at 100% when `PYTHONPATH=::`, which is a better, but it still seems to be stuck initializing. I do not get any completions or any analysis status in the status bar.

```
Starting Microsoft Python language server.

Linter 'pep8' is not installed. Please install it or select another linter".
Error: Module 'pep8' not installed.

Linter 'pylint' is not installed. Please install it or select another linter".
Error: Module 'pylint' not installed.
[...]

PR #962, opened 2019-04-15
Ignore root and dot in search paths

jakebailey:
I've been seeing these in user screenshots/logs.

- `/` is the root on *nix systems, and it showing up in the search paths seems like an error or misconfiguration more than anything. There shouldn't be code there, so I think we can skip it and prevent lookups to odd places which shouldn't be valid.
- `.` is the current directory, which is generally the workspace root, which is already a place we search for user code, so it could be ignored as well.


jakebailey:
I'm going to rework this to handle #849.

PR #966, opened 2019-04-16
Preprocess search paths

### LINKED ISSUE #849: Language Server does not work with PYTHONPATH=./

jakebailey:
Fixes #849.

This does some preprocessing of searchPaths, removing redundant entries and rooting relative paths.

### DIFF
--- a/src/LanguageServer/Impl/Implementation/Server.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.cs
@@ -15,6 +15,7 @@
 
 using System;
 using System.Diagnostics;
+using System.IO;
 using System.Linq;
 using System.Reflection;
 using System.Threading;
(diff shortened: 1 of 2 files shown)


jakebailey:
An option question on this: Should I be also filtering out paths that are just `/`? For example, see #849, where we end up with a search path of `/`. This, to me, seems like an extension bug I don't really want to work around (lest the language server gets run in something like a docker container or CI where `/` actually contains something), but wouldn't be difficult to ignore.

PR #971, opened 2019-04-16
Optimize scope lookup + improve function evaluation performance

### LINKED ISSUE #970: Scope search affects analysis performance
`OpenScope` takes a while since it traverses scopes looking for the scope for a particular AST node.

MikhailArkhipov:
Fixes #970

- There was already cache of scopes but it was populated AFTER traversal. Now scopes are added to it at the creation time and no traversal is needed.

- Additional improvements in function evaluation. If function is not a stub and not specialized and evaluation with specific arguments failed we now assume that eval with arguments will always fail and no longer evaluate it. This saves a lot of time in the large library analysis.


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Callables.cs
```diff
@@ -143,16 +143,26 @@ internal sealed partial class ExpressionEval {
...
                 if (args == null) {
                     return UnknownType;
                 }
-                fd = fn.Overloads.Count > 0 ? fn.Overloads[args.OverloadIndex].FunctionDefinition : null;
+                overload = fn.Overloads.Count > 0 ? fn.Overloads[args.OverloadIndex] : null;
+                fd = overload?.FunctionDefinition;
+            }
+
+            // Stubs are coming from another module and their return type is static.
+            var isStub = fn.DeclaringModule is IDocument doc && fd?.Ast != doc.GetAnyAst();
```
AlexanderSher: Why `fn.DeclaringModule.ModuleType == ModuleType.Stub` doesn't work?

PR #972, opened 2019-04-16
Fix #923: Runaway memory usage when working with msticpy

### LINKED ISSUE #923: Runaway memory usage when working with msticpy
To reproduce:

1. Clone https://github.com/Microsoft/msticpy
2. Install dependencies in a venv (`pip install -r requirements.txt`)
3. Open `msticpy\nbtools\__init__.py`.

Memory is stable until a point (I think when analyzing plotly) when the memory usage just grows and grows.

AlexanderSher:


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.cs
```diff
@@ -27,6 +23,9 @@
 using Microsoft.Python.Core.Logging;
 using Microsoft.Python.Core.Text;
 using Microsoft.Python.Parsing.Ast;
+using System;
```
jakebailey: Nit: sorting

PR #974, opened 2019-04-17
Filter out empty names

### LINKED ISSUE #969: ArgumentOutOfRange exception while creating completion item
```
   at System.Char.IsLetter(String s, Int32 index)
   at Microsoft.Python.LanguageServer.Completion.CompletionItemSource.CreateCompletionItem(String text, CompletionItemKind kind)
   at System.Linq.Enumerable.SelectEnumerableIterator`2.MoveNext()
   at System.Collections.Generic.LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
   at System.Collections.Generic.EnumerableHelpers.ToArray[T](IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Microsoft.Python.LanguageServer.Completion.ImportCompletion.TryGetCompletions(ImportStatement import, CompletionContext context)
   at Microsoft.Python.LanguageServer.Completion.CompletionSource.GetCompletions(IDocumentAnalysis analysis, SourceLocation location)
   at Microsoft.Python.LanguageServer.Implementation.Server.Completion(CompletionParams params, CancellationToken cancellationToken)
   at Microsoft.Python.LanguageServer.Implementation.LanguageServer.Completion(JToken token, CancellationToken cancellationToken)
```

This means `text` is empty, which doesn't seem like a case we want. I've seen this exception with a few different traces (as all of these CreateCompletionItem variants do the same IsLetter call).

MikhailArkhipov:
Fix #969


src/LanguageServer/Impl/Completion/ImportCompletion.cs
```diff
@@ -136,7 +136,9 @@ internal class ImportCompletion {
         private static IEnumerable<CompletionItem> GetAllImportableModules(CompletionContext context) {
             var mres = context.Analysis.Document.Interpreter.ModuleResolution;
             var modules = mres.CurrentPathResolver.GetAllModuleNames().Distinct();
-            return modules.Select(n => CompletionItemSource.CreateCompletionItem(n, CompletionItemKind.Module));
+            return modules
+                .Where(n => !string.IsNullOrEmpty(n))
```
AlexanderSher: It would be a bit faster to have `Where` before `Distinct`, but up to you

src/LanguageServer/Impl/Completion/ImportCompletion.cs
```diff
@@ -164,7 +166,9 @@ internal class ImportCompletion {
             }
 
             if (module != null) {
-                completions.AddRange(module.GetMemberNames().Select(n => context.ItemSource.CreateCompletionItem(n, module?.GetMember(n))));
+                completions.AddRange(module.GetMemberNames()
+                    .Where(n => !string.IsNullOrEmpty(n))
+                    .Select(n => context.ItemSource.CreateCompletionItem(n, module?.GetMember(n))));
```
AlexanderSher: `module` can't be null here.

Issue #979, opened 2019-04-17
Library function is renamable


jakebailey:
Open `proj` in VS Code. The values from `lib` should be resolved.

Rename `make_foo` in main.py succeeds, but should not.

[librename.zip](https://github.com/Microsoft/python-language-server/files/3091225/librename.zip)

jakebailey:
Debugged the call, `make_foo`'s declaring module is null, which means that it gets treated as an empty analysis, which means that it isn't library code. It should have a declaring module.

MikhailArkhipov:
Goto definition is acting up too.

PR #988, opened 2019-04-18
Cache function eval result per argument set

MikhailArkhipov:
Performance


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Callables.cs
```diff
@@ -207,9 +207,26 @@ internal sealed partial class ExpressionEval {
             return instance.Call(p.Name, ArgumentSet.Empty);
         }
 
-        private IMember TryEvaluateWithArguments(IPythonModule module, FunctionDefinition fd, IArgumentSet args) {
+
+        private readonly Dictionary<int, IMember> _argEvalCache = new Dictionary<int, IMember>();
+
+        private IMember TryEvaluateWithArguments(IPythonFunctionType fn, IArgumentSet args) {
+            var name = fn.DeclaringType != null ? $"{fn.DeclaringModule.Name}.{fn.Name}" : fn.Name;
```
jakebailey: This might be affected by #979 where the DeclaringModule of a function I had was `null`, so all functions of that style with the same name would get cached together.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Callables.cs
```diff
@@ -207,9 +207,26 @@ internal sealed partial class ExpressionEval {
...
+            var argHash = args
+                .Arguments
+                .Select(a => a.Name.GetHashCode() ^ 397 * (a.Value?.GetHashCode() ?? 0))
+                .Aggregate(0, (current, d) => 31 * current ^ d);
+            var key = fn.DeclaringModule.Name.GetHashCode() ^ name.GetHashCode() ^ (397 * argHash);
+
+            if (_argEvalCache.TryGetValue(key, out var result)) {
+                return result;
```
jakebailey: Won't hash collisions here make some unrelated functions have the same evaluation?

MikhailArkhipov: Functions have fully qualified names like module.class.method which should have unique hashes

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Callables.cs
MikhailArkhipov: Yes, but it is a bug. I'll work on it right after this. All types should have declaring module, there are many many places where we would crash otherwise.

PR #1011, opened 2019-04-24
Call methods for binary operators, better handle builtin type conversions

### LINKED ISSUE #860: "datetime - datetime" is timedelta, not datetime
(from https://github.com/Microsoft/vscode-python/issues/4983)

The result of subtracting one datetime.datetime from another is a datetime.timedelta.  However, the language server is inferring that it is actually a datetime.

```python
from datetime import datetime
date1 = datetime.strptime('2019/3/12', '%Y/%m/%d')
date2 = datetime.strptime('2019/3/19', '%Y/%m/%d')
diffDate = date2-date1  # Intellisense shows this as a datetime.
```

I'm guessing that the default inference for subtraction is the same type as the two arguments.  FYI, there may be other types beside datetime.datetime where that does not hold true.

jakebailey:
Fixes #860.

This should cover the bulk of the operators; some cases which aren't parsed as BinaryExpressions are not yet handled (e.g. `pow()` should call `__pow__`). Builtins are done by hand, as conversions mean that we can't trust their operator methods. I wrote a script to help me write that code; I'll throw it somewhere when I get the chance.

As far as I've measured, there doesn't seem to be any major performance impact when expanding the operators.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
@@ -116,22 +116,35 @@ namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
                 return Interpreter.GetBuiltinType(leftTypeId);
             }
 
-            if (IsOperableBuiltin(leftTypeId) && IsOperableBuiltin(rightTypeId)) {
+            var leftIsOperable = IsOperableBuiltin(leftTypeId);
+            var rightIsOperable = IsOperableBuiltin(rightTypeId);
+
+            if (leftIsOperable && rightIsOperable) {
                 if (TryGetValueFromBuiltinBinaryOp(op, leftTypeId, rightTypeId, Interpreter.LanguageVersion.Is3x(), out var member)) {
                     return member;
                 }
             }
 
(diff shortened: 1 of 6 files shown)


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
```diff
@@ -79,9 +80,105 @@ internal sealed partial class ExpressionEval {
...
+                    return Interpreter.GetBuiltinType(BuiltinTypeId.Bool);
+                }
+
+                return UnknownType;
+            }
+
+            var leftType = left.GetPythonType();
+            var rightType = right.GetPythonType();
+
+            if (leftType == null) {
```
MikhailArkhipov: Probably never null since you do `?? UnknownType`

jakebailey: I think it could still be null if I were trying to compare a module to something else (or some other case the GetPythonType extension isn't handling). I'll test it and see.

jakebailey: I tested things, and these don't appear to ever be null.

PR #1020, opened 2019-04-26
Fix a few misc bugs

jakebailey:
Found via Sonar.


src/Analysis/Ast/Impl/Types/ArgumentSet.cs
```diff
@@ -406,7 +406,7 @@ private sealed class DictArg : IDictionaryArgument {
             public DictArg(string name, Expression expression, Node location) {
                 Name = name;
                 Expression = expression;
-                Location = Location;
+                Location = location;
```
AlexanderSher: This is a good one!

PR #1029, opened 2019-04-30
Misc performance improvements (AnalysisModuleKey hash, dependency graph list Adds)

jakebailey:
`AnalysisModuleKey.GetHashCode` was getting called a lot; it does nontrivial work producing a hash for the file path. Precompute it at initialization time (as it will always be called; it's used as a key in multiple `HashSet`s).

image

Next, avoid calling `Add` repeatedly on `ImmutableArray` instances by using a normal `List` with a capacity hint, and a final `ToImmutableArray` when populated. A lot of the time was being spent allocating new arrays over and over.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
@@ -26,16 +26,20 @@ namespace Microsoft.Python.Analysis.Analyzer {
         public string FilePath { get; }
         public bool IsTypeshed { get; }
 
+        private readonly int _hashCode;
+
         public AnalysisModuleKey(IPythonModule module) {
             Name = module.Name;
             FilePath = module.ModuleType == ModuleType.CompiledBuiltin ? null : module.FilePath;
             IsTypeshed = module is StubPythonModule stub && stub.IsTypeshed;
+            _hashCode = CreateHashCode(Name, FilePath, IsTypeshed);
         }
 
         public AnalysisModuleKey(string name, string filePath, bool isTypeshed) {
             Name = name;
             FilePath = filePath;
             IsTypeshed = isTypeshed;
+            _hashCode = CreateHashCode(Name, FilePath, IsTypeshed);
         }
 
         public bool Equals(AnalysisModuleKey other)
@@ -43,11 +47,13 @@ namespace Microsoft.Python.Analysis.Analyzer {
 
         public override bool Equals(object obj) => obj is AnalysisModuleKey other && Equals(other);
 
-        public override int GetHashCode() {
+        public override int GetHashCode() => _hashCode;
+
+        private static int CreateHashCode(string name, string filePath, bool isTypeshed) {
             unchecked {
-                var hashCode = (Name != null ? Name.GetHashCode() : 0);
-                hashCode = (hashCode * 397) ^ (FilePath != null ? FilePath.GetPathHashCode() : 0);
-                hashCode = (hashCode * 397) ^ IsTypeshed.GetHashCode();
+                var hashCode = (name != null ? name.GetHashCode() : 0);
+                hashCode = (hashCode * 397) ^ (filePath != null ? filePath.GetPathHashCode() : 0);
+                hashCode = (hashCode * 397) ^ isTypeshed.GetHashCode();
                 return hashCode;
             }
         }
--- a/src/Analysis/Ast/Impl/Dependencies/DependencyGraph.cs
+++ b/src/Analysis/Ast/Impl/Dependencies/DependencyGraph.cs
@@ -131,12 +131,12 @@ namespace Microsoft.Python.Analysis.Dependencies {
             var missingKeysHashSet = new HashSet<TKey>();
             for (var i = 0; i < _verticesByIndex.Count; i++) {
                 var vertex = _verticesByIndex[i];
-                var newIncoming = ImmutableArray<int>.Empty;
+                var newIncoming = new List<int>(vertex.IncomingKeys.Count);
                 var oldIncoming = vertex.Incoming;
 
                 foreach (var dependencyKey in vertex.IncomingKeys) {
                     if (_keyToVertexIndex.TryGetValue(dependencyKey, out var index)) {
-                        newIncoming = newIncoming.Add(index);
+                        newIncoming.Add(index);
                     } else {
                         missingKeysHashSet.Add(dependencyKey);
                         if (vertex.IsSealed) {
@@ -146,7 +146,7 @@ namespace Microsoft.Python.Analysis.Dependencies {
                     }
                 }
 
-                if (newIncoming.SequentiallyEquals(oldIncoming)) {
+                if (newIncoming.Count == oldIncoming.Count && newIncoming.SequenceEqual(oldIncoming)) {
                     continue;
                 }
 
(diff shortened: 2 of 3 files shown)


src/Analysis/Ast/Impl/Dependencies/DependencyGraph.cs
```diff
@@ -131,12 +131,13 @@ internal sealed class DependencyGraph<TKey, TValue> {
             var missingKeysHashSet = new HashSet<TKey>();
             for (var i = 0; i < _verticesByIndex.Count; i++) {
                 var vertex = _verticesByIndex[i];
-                var newIncoming = ImmutableArray<int>.Empty;
+                var newIncomingArr = vertex.IncomingKeys;
```
jakebailey: My only worry here is that this is a copy and has a performance impact; there's no way for me to refer to this array except to do a full name everywhere (which I can do), but when I profiled it, it didn't seem to matter one way or the other, so I left this for readability reasons. Happy to change if it's bad.

AlexanderSher:
There is a big change that I have that removes `DependencyGraph` as a separate type completely. `ImmutableArray<T>.Add` isn't much different from `List<T>.Add`: https://github.com/dotnet/coreclr/blob/379c3848ada079fce37d2691869b73531120c212/src/System.Private.CoreLib/shared/System/Collections/Generic/List.cs#L103 - it changes size of the list based on the incoming elements and reuses the same instance of the list when possible.

jakebailey:
`List` is a lot different, though... `List` has an underlying capacity which grows as needed, and is not related to the number of items in the list. When `List.Add` is called and the added item doesn't fit, it will call `EnsureCapacity`, which does the typical "double the size" to get amortized O(1) appends. Appending to `ImmutableList` is O(n) in any case because it's immutable and there is no underlying storage that can be resized and reused at will. That's where this code is slow; it appends to an immutable list (starting at empty) over and over until it's finished. Switching to a `List` just for the first part means that it can get a capacity hint (so doesn't need to grow up to a point), then gets the amortized boost later on. Then it gets converted to an immutable array to get used elsewhere.

If you're going to eliminate some of this code, then I can remove the bits you plan to change from this PR or do something later.

AlexanderSher:
Cool, you have found a bug in `ImmutableArray`. After fix, perf looks like this:

| Method | Mean | Error | StdDev |
| --- | --- | --- | --- |
| Add to List 10 items | 82.76 ns | 0.3918 ns | 0.3271 ns |
| Add to ImmutableArray 10 items |    162.40 ns |  0.7064 ns |  0.6608 ns |
| Add to List 100 items |    389.54 ns |  2.3554 ns |  1.9669 ns |
| Add to ImmutableArray 100 items |  1,216.86 ns | 23.7266 ns | 23.3027 ns |
| Add to List 1000 items |  2,623.46 ns | 14.3219 ns | 12.6960 ns |
| Add to ImmutableArray 1000 items | 11,033.17 ns | 73.1005 ns | 68.3783 ns |

There is still an additional cost, but it is due to copying of the structure itself, not the arrays

src/Analysis/Ast/Impl/Dependencies/DependencyResolver.cs
```diff
@@ -129,12 +130,12 @@ public int ChangeValue(TKey key, TValue value, params TKey[] incomingKeys)
         }
 
         private static ImmutableArray<WalkingVertex<TKey, TValue>> CreateWalkingGraph(ImmutableArray<DependencyVertex<TKey, TValue>> snapshot, ImmutableArray<DependencyVertex<TKey, TValue>> changedVertices) {
-            var analysisGraph = ImmutableArray<WalkingVertex<TKey, TValue>>.Empty;
+            var analysisGraph = new List<WalkingVertex<TKey, TValue>>(changedVertices.Count);
```
heejaechang: you can use ImmutableArray.Builder rather than use List and then do ToImmutableArray

src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
```diff
@@ -26,28 +26,34 @@ internal struct AnalysisModuleKey : IEquatable<AnalysisModuleKey> {
         public string FilePath { get; }
         public bool IsTypeshed { get; }
 
+        private readonly int _hashCode;
+
         public AnalysisModuleKey(IPythonModule module) {
             Name = module.Name;
             FilePath = module.ModuleType == ModuleType.CompiledBuiltin ? null : module.FilePath;
             IsTypeshed = module is StubPythonModule stub && stub.IsTypeshed;
+            _hashCode = CreateHashCode(Name, FilePath, IsTypeshed);
```
heejaechang: you might want to calculate the hash lazily if creating it multiple time was an issue at the first place?

src/Analysis/Ast/Impl/Dependencies/DependencyGraph.cs
heejaechang: this also looks like you can use ImmutableArray.Builder.

also, you can actually pool Builder to even remove some more cost?

src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
heejaechang: ah, sorry didn't see the description that saying HashCode is always get called.

src/Analysis/Ast/Impl/Dependencies/DependencyResolver.cs
jakebailey: There is no builder; this is our own implementation in our repo, not the widely available one.

src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
jakebailey: Lazily was what I'd have done, but unfortunately this is a struct so it'd need to be done in the initializer (unless there's something I'm missing about struct behaviors).

AlexanderSher: There is no benefit of doing it lazy. This type is made to be a key in the dictionary, so that is a proper fix.

src/Analysis/Ast/Impl/Dependencies/DependencyResolver.cs
heejaechang: I see. thank you for the answer. just curious. can I ask why you guys are not using the public immutable array? they also has ImmutableList that is baed on tree.

AlexanderSher: `ImmutableList` is slow. Even for enumeration and accessing element by index. `ImmutableArray` is fast on reading but very slow on modification. We know that most of our operations are either adding or iterating, so I've added a type that is fast for that.

heejaechang: by the way, since you guys are using .net core.

moving to new Span<T> - https://msdn.microsoft.com/en-us/magazine/mt814808.aspx - could be another way to improve perf around these?

AlexanderSher: Sure, but it isn't part of .Net Standard yet. We will start using it as soon as .net standard 2.1 is released.

jakebailey:
Closing; as of #1036 the immutable array is fixed and that hashcode function does not appear in the profile.

Issue #821, opened 2019-03-25
Language Server Analysis Time in Conda Environment


Anapo14:
VSCode Version: 1.32.3
Extension Version: 2019.2.5558
OS: Windows 10
Python 3.7.2
LS Version: 0.2.30.0
Using Conda virtual environment

I have a conda environment that includes the list of packages in this [attachment](https://github.com/Microsoft/python-language-server/files/3004298/packages_in_env.txt). However, it take the LS 17 minutes to complete analysis on this environment. 

![image](https://user-images.githubusercontent.com/10341823/54935534-bd2b9600-4edd-11e9-802b-f85d895a494b.png)

(Note the time that analysis starts and the time on my computer's clock...the screenshot was taken before analysis was fully completed).

I am working with this simple script:
```
import pandas as pd 
import datetime 
import pandas_datareader.data as web 

start = datetime.datetime(2010, 1,1)
end = datetime.datetime.now()

df = web.DataReader("MSFT", "yahoo", start, end)

df.reset_index(inplace = True )
df.set_index("Date", inplace = True)

import matplotlib.pyplot as plt
from matplotlib import style

style.use('fivethirtyeight')


df['High'].plot()

plt.legend()
plt.show()
```

luabud:
For me it didn't take long, but this is what I got on the output window:

``Starting Microsoft Python language server.
[Info  - 10:08:03 AM] GetCurrentSearchPaths C:\Users\luabud\AppData\Local\Continuum\Anaconda3\envs\env\python.exe 
[Info  - 10:08:03 AM] Python search paths:
[Info  - 10:08:03 AM]     c:\users\luabud\appdata\local\continuum\anaconda3\envs\env\dlls
[Info  - 10:08:03 AM]     c:\users\luabud\appdata\local\continuum\anaconda3\envs\env\lib
[Info  - 10:08:03 AM]     c:\users\luabud\appdata\local\continuum\anaconda3\envs\env
[Info  - 10:08:03 AM]     c:\users\luabud\appdata\roaming\python\python37\site-packages
[Info  - 10:08:03 AM]     c:\users\luabud\appdata\local\continuum\anaconda3\envs\env\lib\site-packages
[Info  - 10:08:03 AM] Configuration search paths:
[Info  - 10:08:03 AM]     C:\Users\luabud\AppData\Local\Continuum\Anaconda3\envs\env\python37.zip
[...]

PR #1044, opened 2019-05-06
Do not create scopes and variables in AST

MikhailArkhipov:
We do it in analysis. One syntax error will be lost, but can be restored in linter if really needed. Saves quite a bit of memory.


jakebailey:
Testing this on the example in #821 causes the analysis to hang for a very long time while analyzing PyQt5. I think this info might be getting used somewhere, or maybe there's a syntax error in the library (from our parser's POV) where we're currently skipping a bunch of work.

PR #1045, opened 2019-05-06
Make node names in AST const strings

MikhailArkhipov:


jakebailey:
I'm surprised this does something; do these strings not get interned and have only one instance?

AlexanderSher:
Yeah, that's weird. All string constants are interned automatically. You will get new instance of the string if you do something like `new StringBuilder().Append(str).ToString();`.

MikhailArkhipov:
It's a part of several other things I am going to do, so yeah, its minor.

PR #1050, opened 2019-05-06
Do not create list of overloads for every function type

MikhailArkhipov:
Since most functions only have a single overload, we don't need to create list in each and every case.


src/Analysis/Ast/Impl/Types/PythonFunctionType.cs
```diff
@@ -141,8 +142,14 @@ public override PythonMemberType MemberType
         internal ImmutableArray<string> Dependencies { get; private set; } = ImmutableArray<string>.Empty;
 
         internal void AddOverload(IPythonFunctionOverload overload) {
-            lock (_lock) {
+            if (_overload != null) {
+                _overloads = new List<IPythonFunctionOverload> { _overload };
```
jakebailey: It'd be a bit faster to create the list as `{ _overload, overload }` and return early, rather than doing an `Add` later on the fresh list.

MikhailArkhipov: Sure

PR #1051, opened 2019-05-06
Separate imported and plain variables

### LINKED ISSUE #1053: Remove fields only used in some types from base classes
Specifically, `MemberType` from `LocatedMember`, as well as `Parent` that is only used in imported variables and stays null otherwise.

MikhailArkhipov:
We declare a lot of variables. Currently each variable is a `LocatedMember` which has `Parent` field. However, only imported items actually use this property, all others have it null. This change is to introduce `IImportedMember` and `ImportedVariable` hence reducing size of 90% of the variables in scopes.

In addition, LocatedMember no longer stores `MemberType` and instead derived classes should override the property and return value specific to the type hence removing another field from use in all types.

Fixes #1053


src/LanguageServer/Impl/Sources/ReferenceSource.cs
```diff
@@ -146,10 +146,20 @@ internal sealed class ReferenceSource {
...
         private ILocatedMember GetRootDefinition(ILocatedMember lm) {
-            for (; lm.Parent != null; lm = lm.Parent) { }
-            return lm;
+            if (!(lm is IImportedMember im) || im.Parent == null) {
+                return lm;
+            }
+
+            var parent = im.Parent;
+            for (; parent != null;) {
+                if (!(parent is IImportedMember im1) || im1.Parent == null) {
```
jakebailey: The condition here is slightly different, I assume there's no way intended for a variable to not be the root but not implement IImportedMember? (Since before a `null` parent was the indicator, but now you could technically end early.)

MikhailArkhipov: Basically it looks for a root that does not have parent. In new code only imported member has parent, so yes, member that is not imported cannot have the parent.

PR #1055, opened 2019-05-07
Fix 1049: Opening second file with PyQt5 import reanalyzes more than it should

AlexanderSher:
Also use ImmutableArray instead of pair of fields in `PythonFunctionType`.

### DIFF
--- a/src/Analysis/Ast/Impl/Types/PythonFunctionType.cs
+++ b/src/Analysis/Ast/Impl/Types/PythonFunctionType.cs
@@ -26,8 +26,7 @@ using Microsoft.Python.Parsing.Ast;
 namespace Microsoft.Python.Analysis.Types {
     [DebuggerDisplay("Function {Name} ({TypeId})")]
     internal class PythonFunctionType : PythonType, IPythonFunctionType {
-        private IPythonFunctionOverload _overload;
-        private List<IPythonFunctionOverload> _overloads;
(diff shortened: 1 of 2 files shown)


src/Analysis/Ast/Impl/Types/PythonFunctionType.cs
```diff
@@ -141,19 +139,8 @@ public override PythonMemberType MemberType
...
-            }
-
-            if (_overloads != null) {
-                _overloads.Add(overload);
-            } else {
-                _overload = overload;
-            }
-        }
+        internal void AddOverload(IPythonFunctionOverload overload) 
+            => _overloads = _overloads.Count > 0 ? _overloads.Add(overload) : ImmutableArray<IPythonFunctionOverload>.Create(overload);
```
jakebailey: Is there any performance benefit to this rather than just calling `Add` unconditionally? I think it needs to allocate a new backing array in any case.

jakebailey: Hmm, I guess it skips over all of the compare exchange and such.

AlexanderSher: There is a memory benefit. By default, `ImmutableArray<T>` acts like a `List<T>` - creates an underlying array of 4 elements. Since it is a common case here that we have only one element (at least how I understand original optimization), we can start by creating an array with just one element.

PR #1063, opened 2019-05-09
Only use __all__ in star imports

### LINKED ISSUE #867: Only use __all__ in star imports
We're using the result of `__all__` to determine what a module exports; this isn't completely correct. `__all__` only impacts what is transferred in `from foo import *` style import. All other names are passed through (ignoring `_` prefixed names, other than specific ones like `__all__` or `__name__`, etc). See:

- https://docs.python.org/3.7/tutorial/modules.html#importing-from-a-package
- https://stackoverflow.com/questions/44834/can-someone-explain-all-in-python

There are likely other nuances to consider here.

jakebailey:
Fixes #867.

I had to change the DocumentAnalysis interface to swap out the ImmutableArray for something that's nullable, to distinguish when `__all__` isn't specified or is unusable. #867 mentions exposing this list through the module; I just accessed it directly through Analysis (since it was there), but I can move it out into its own function/property of the module if that feels better.

Needs a bit more testing, but as far as I can tell, numpy/tensorflow/django (major users of `__all__`) appear to be unaffected by this.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/DocumentAnalysis.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/DocumentAnalysis.cs
@@ -29,14 +29,14 @@ using Microsoft.Python.Parsing.Ast;
 
 namespace Microsoft.Python.Analysis.Analyzer {
     internal sealed class DocumentAnalysis : IDocumentAnalysis {
-        public DocumentAnalysis(IDocument document, int version, IGlobalScope globalScope, IExpressionEvaluator eval, ImmutableArray<string> exportedMemberNames) {
+        public DocumentAnalysis(IDocument document, int version, IGlobalScope globalScope, IExpressionEvaluator eval, IReadOnlyList<string> exportedMemberNames) {
             Check.ArgumentNotNull(nameof(document), document);
             Check.ArgumentNotNull(nameof(globalScope), globalScope);
             Document = document;
             Version = version;
             GlobalScope = globalScope;
             ExpressionEvaluator = eval;
-            ExportedMemberNames = exportedMemberNames;
+            StarImportMemberNames = exportedMemberNames;
         }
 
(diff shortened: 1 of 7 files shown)


src/Analysis/Ast/Impl/Analyzer/DocumentAnalysis.cs
```diff
@@ -29,14 +29,14 @@
...
-        public DocumentAnalysis(IDocument document, int version, IGlobalScope globalScope, IExpressionEvaluator eval, ImmutableArray<string> exportedMemberNames) {
+        public DocumentAnalysis(IDocument document, int version, IGlobalScope globalScope, IExpressionEvaluator eval, IReadOnlyList<string> exportedMemberNames) {
             Check.ArgumentNotNull(nameof(document), document);
             Check.ArgumentNotNull(nameof(globalScope), globalScope);
             Document = document;
             Version = version;
             GlobalScope = globalScope;
             ExpressionEvaluator = eval;
-            ExportedMemberNames = exportedMemberNames;
+            StarImportMemberNames = exportedMemberNames;
```
MikhailArkhipov: Probably change parameter name to match

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -209,7 +208,7 @@ public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst
         }
 
         public IGlobalScope GlobalScope => Eval.GlobalScope;
-        public ImmutableArray<string> ExportedMemberNames { get; private set; }
+        public IReadOnlyList<string> ExportedMemberNames { get; private set; }
```
AlexanderSher: You can also change `ModuleWalker.Complete`

PR #1069, opened 2019-05-11
Add telemetry event and timer for RPC requests

### LINKED ISSUE #1068: Add new telemetry event for RPC call timing/profiling
On top of exceptions, also send some timing information for profiling on select events (like completion). Not for every single invocation, just a couple per.

jakebailey:
Closes #1068.

I did this in `LanguageServer.cs` instead of further downstream in order to get the "true" request time, i.e. I want to see the time that was spent waiting in the prioritizer if any. Where required, some functions now await their result instead of returning the task, otherwise the timing will only capture the time to create a task and not to complete it.

To limit the number of events sent, it'll only send up to 10 per method. Another solution might be to debounce it using a few IdleTimeActions that reset a boolean per event.

I only added this to the methods I thought would be interesting to examine. Some were explicitly skipped (document changes, line formatting, etc). The `Timer` supports adding extra measures and properties, which are used to send the number of completion items, symbols, etc, for profiling.

### DIFF
--- a/src/LanguageServer/Impl/Telemetry/RequestTimer.cs
+++ b/src/LanguageServer/Impl/Telemetry/RequestTimer.cs
@@ -37,7 +37,7 @@ namespace Microsoft.Python.LanguageServer.Telemetry {
                 _events.TryGetValue(method, out var eventNum);
 
                 if (eventNum >= MaxEvents) {
-                    return new Timer();
+                    return Timer.Disabled;
                 }
 
                 _events[method] = eventNum + 1;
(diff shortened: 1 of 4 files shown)


src/LanguageServer/Impl/Telemetry/RequestTimer.cs
```diff
@@ -0,0 +1,105 @@
...
+        public RequestTimer(ITelemetryService telemetryService) {
+            _telemetryService = telemetryService;
+        }
+
+        public Timer Time(string method) {
+            lock (_lock) {
+                _events.TryGetValue(method, out var eventNum);
+
+                if (eventNum >= MaxEvents) {
+                    return new Timer();
```
MikhailArkhipov: Perhaps make static Timer.Disabled and return it, like we have in Core `Disposable.Empty`

PR #1090, opened 2019-05-15
Python inspector service

jakebailey:
Opening to get some feedback and thoughts.

The inspector is a partial JSON-RPC server, which takes requests from the LS. Adding more methods is pretty simple, just decorate a function.

~Right now, all it does is get the module member names and `__all__`. This is marginally useful (especially since it's not used anywhere), but we can add more methods to do something like the scraper for full type info, or check for installed versions, etc.~

Some things I'm not sure about are:
- If the process dies, it will be restarted, but the requests that were issued to the crashed process will fail and throw.
- The process should periodically be killed to save on memory.
- The scraper takes extra arguments for path modification which I'm not sure I can really handle due to the process persisting (see `initial_import` in `scrape_module.py`).
- Because this uses StreamJsonRpc, I have to implement it in the language server assembly, otherwise the analysis has to depend on that library. The best I can do is declare the interface outside of the LS, but where I've declared it is probably wrong.
- Because StreamJsonRpc is all async, using the inspector in the analyzer requires some weird tasking (I did a hack in PythonModule to call the inspector for GetModuleMembers, which meant `Services.GetService<IPythonInspector>()?.GetModuleMemberNames(Name)?.Result?.Members`; not thrilling).

### DIFF
--- a/src/Analysis/Ast/Impl/Inspection/IPythonInspector.cs
+++ b/src/Analysis/Ast/Impl/Inspection/IPythonInspector.cs
@@ -20,6 +20,11 @@ using System.Threading.Tasks;
 
 namespace Microsoft.Python.Analysis.Inspection {
     public interface IPythonInspector {
-        Task<IReadOnlyList<string>> GetModuleMemberNames(string moduleName);
+        Task<ModuleMemberNamesResponse> GetModuleMemberNames(string moduleName);
+    }
+
+    public class ModuleMemberNamesResponse {
+        public IReadOnlyList<string> Members;
+        public IReadOnlyList<string> All;
     }
 }
--- a/src/Analysis/Ast/Impl/Inspection/IPythonInspector.cs
+++ b/src/Analysis/Ast/Impl/Inspection/IPythonInspector.cs
@@ -21,6 +21,8 @@ using System.Threading.Tasks;
 namespace Microsoft.Python.Analysis.Inspection {
     public interface IPythonInspector {
         Task<ModuleMemberNamesResponse> GetModuleMemberNames(string moduleName);
+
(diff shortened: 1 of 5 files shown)


jakebailey:
Right now, this can give back module members/`__all__`, and try to find the version of a module through `pkg_resources`. If the process dies, then the pending requests will fail, but further requests will restart the process. The service could be disposed at any time to trigger a restart (though it's somewhat of a hack and not thread safe; it doesn't keep a pending request count to know it's safe to restart).

It's only used in its own tests. I have some stashed code that would go into `PythonModule` to grab info from the service, but it's a bit hacky and it'll be better to wait until this can be a provider for information (alongside the real analysis or database, etc).

src/Analysis/Ast/Impl/Inspection/IPythonInspector.cs
```diff
@@ -0,0 +1,32 @@
...
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.Python.Analysis.Inspection {
+    public interface IPythonInspector {
+        Task<ModuleMemberNamesResponse> GetModuleMemberNames(string moduleName);
```
MikhailArkhipov: Probably `Async` so .NET style cop won't bark. Can it handle cancellation?

jakebailey: I could modify the signature to accept a token and pass it through to the JsonRpc instance, though I'm not really sure what will happen given that I implemented the JSON-RPC server by hand in Python and it's single threaded. I'll test and see.

PR #1106, opened 2019-05-20
Prevent SO on documentation fetch

### LINKED ISSUE #1100: StackOverflowException with PyMySQL==0.7.11
The language server raises StackOverflowException and won't start up, if my code imports pymysql. The code itself runs well, just the language server keep crashes.

+ Environment: macOS Mojave 10.14.4
+ Python: 2.7.15 with pyenv and virtualenv
+ VSCode: 1.34.0 (1.34.0) a622c65b2c713c890fcf4fbf07cf34049d5fe758
+ Microsoft Python Language Server version 0.2.82.0
+ Installed PyMySQL==0.7.11 and import it from code.

Debug info from python outputs:
```
[Info  - 11:42:33 AM] Microsoft Python Language Server version 0.2.82.0
[Info  - 11:42:33 AM] Initializing for /Users/xxxx/.pyenv/versions/2.7.15/bin/python
Process is terminating due to StackOverflowException.
```

MikhailArkhipov:
Fixes #1100 

Also harden writing of scraped content as it may occasionally collide scraping same file (I've seen exceptions thrown inside CLR on writing to a closed file.

### DIFF
--- a/src/Analysis/Ast/Impl/Types/PythonClassType.cs
+++ b/src/Analysis/Ast/Impl/Types/PythonClassType.cs
@@ -17,7 +17,6 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-using System.Threading;
 using Microsoft.Python.Analysis.Modules;
 using Microsoft.Python.Analysis.Specializations.Typing;
 using Microsoft.Python.Analysis.Types.Collections;
@@ -33,8 +32,7 @@ namespace Microsoft.Python.Analysis.Types {
     [DebuggerDisplay("Class {Name}")]
     internal class PythonClassType : PythonType, IPythonClassType, IPythonTemplateType, IEquatable<IPythonClassType> {
         private static readonly string[] _classMethods = { "mro", "__dict__", @"__weakref__" };
-        private readonly object _lock = new object();
-        private readonly AsyncLocal<IPythonClassType> _processing = new AsyncLocal<IPythonClassType>();
+        private IPythonClassType _processing;
         private List<IPythonType> _bases;
         private IReadOnlyList<IPythonType> _mro;
         private Dictionary<string, IPythonType> _genericParameters;
@@ -60,9 +58,7 @@ namespace Microsoft.Python.Analysis.Types {
 
         public override IEnumerable<string> GetMemberNames() {
             var names = new HashSet<string>();
-            lock (_lock) {
-                names.UnionWith(Members.Keys);
-            }
+            names.UnionWith(Members.Keys);
             foreach (var m in Mro.Skip(1)) {
                 names.UnionWith(m.GetMemberNames());
             }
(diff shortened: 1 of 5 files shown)


src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -104,19 +105,27 @@ internal PythonClassType(string name, Location location)
...
+                    _documentation = (GetMember("__init__") as IPythonFunctionType)?.Documentation;
+                }
+                if (string.IsNullOrEmpty(_documentation) && Bases != null) {
+                    // If still not found, try bases. 
                     var o = DeclaringModule.Interpreter.GetBuiltinType(BuiltinTypeId.Object);
-                    doc = Bases.Except(Enumerable.Repeat(o, 1)).FirstOrDefault(b => !string.IsNullOrEmpty(b?.Documentation))?.Documentation;
+                    _documentation = Bases.FirstOrDefault(b => b != o && !string.IsNullOrEmpty(b?.Documentation))?.Documentation;
                 }
-                return doc;
+                return _documentation;
```
jakebailey: `Pop`

src/Core/Impl/Extensions/IOExtensions.cs
```diff
@@ -114,28 +114,28 @@ public static class IOExtensions {
...
-                        var bytes = Encoding.UTF8.GetBytes(text);
-                        stream.Write(bytes, 0, bytes.Length);
-                        return;
-                    }
+            var failed = false;
+            for (var retries = 100; retries > 0; --retries) {
+                try {
+                    fs.WriteAllText(filePath, text);
+                } catch (Exception ex) when (ex is IOException || ex is UnauthorizedAccessException) {
+                    failed = true;
```
AlexanderSher: If there was one exception and next retry succeeds, file will be deleted later. We need either return after `fs.WriteAllText` or set `failed` to `true`.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -104,19 +105,35 @@ internal PythonClassType(string name, Location location)
...
-                    doc = init?.DeclaringType == this ? init.Documentation : null;
+                if (!string.IsNullOrEmpty(_documentation)) {
+                    return _documentation;
                 }
-                // Try bases.
-                if (string.IsNullOrEmpty(doc) && Bases != null) {
-                    var o = DeclaringModule.Interpreter.GetBuiltinType(BuiltinTypeId.Object);
-                    doc = Bases.Except(Enumerable.Repeat(o, 1)).FirstOrDefault(b => !string.IsNullOrEmpty(b?.Documentation))?.Documentation;
+                // Make sure we do not cycle through bases back here.
+                if (!Push(this)) {
```
jakebailey: Is it possible for this set to be in use for something else when the documentation is being calculated? Should there be a second one specifically for documentation accesses?

MikhailArkhipov: Push/Pop should be under lock

src/Analysis/Ast/Impl/Types/PythonClassType.cs
MikhailArkhipov: Done

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -372,40 +391,44 @@ public bool Equals(IPythonClassType other)
...
+
+                    // Create specific types for generic type bases
+                    // it for generic types but not Generic[T, ...] itself.
+                    foreach (var gt in genericTypeBases) {
+                        var st = gt.Parameters
+                            .Select(p => classType.GenericParameters.TryGetValue(p.Name, out var t) ? t : null)
+                            .Where(p => !p.IsUnknown())
+                            .ToArray();
+                        if (st.Length > 0) {
+                            var type = gt.CreateSpecificType(new ArgumentSet(st));
```
AlexanderSher: Is this a recursion under lock?

MikhailArkhipov: Yes as with other push/pops. However, specifically here it is always single task since we don't run analysis and access it at the same time. Actually I am wondering if locks here are needed at all...

src/Analysis/Ast/Impl/Types/PythonClassType.cs
jakebailey: FWIW my comment here wasn't to do with locking, it was if Documentation was accessed during another call that was using Push/Pop to prevent infinite recursion where some class is visited, then somewhere down the callstack its documentation accessed (which will fail, because now Push returns false).

PR #1111, opened 2019-05-21
Update signature parameter handling to LSP 3.14

### LINKED ISSUE #679: Intellisense - Function arguments bolded in wrong order
@JeffreyCA commented on [Fri Feb 08 2019](https://github.com/Microsoft/vscode-python/issues/4331)

## Environment data

- VS Code version: 1.32.0-insider
- Extension version (available under the Extensions sidebar): 2019.1.0
- OS and version: Windows 10
- Python version (& distribution if applicable, e.g. Anaconda): 3.6.8
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): virtualenv
- Relevant/affected Python packages and their versions: tensorflow 1.12.0

## Expected behaviour
Function arguments should be bolded in listed order

## Actual behaviour
Function arguments are not bolded in order. In addition only a substring of an argument is bolded in some cases.

## Steps to reproduce:
1. Have the following piece of code:

```py
import tensorflow as tf
from tensorflow import keras

model = keras.Sequential([
    tf.layers.Dense(64, activation = tf.nn.relu, input_shape = [ ]),
    tf.layers.Dense(1)
])
optimizer = tf.keras.optimizers.RMSprop(0.001)
```

Then begin typing the following, filling in the arguments:

```py
mode.compile(
```

## Demo

20190208at1046am

---

@DonJayamanne commented on [Tue Feb 19 2019](https://github.com/Microsoft/vscode-python/issues/4331#issuecomment-465196426)

Are you using Jedi or language server?
Please check the output of your `Python` output panel and let me know what you get when you have a python file open and the above issue.
Or let me know if you have the following setting enabled in your `local .vscode/settings.json` or `user settings.json`:
```js
"python.jediEnabled": true|false
```

---

@JeffreyCA commented on [Tue Feb 19 2019](https://github.com/Microsoft/vscode-python/issues/4331#issuecomment-465252156)

```
Microsoft Python Language Server version 0.1.78.0
Initializing for C:\venv\Scripts\python.exe
************* Module tensor
17,25,error,no-member:Module 'matplotlib.cm' has no 'binary' member

------------------------------------------------------------------

Your code has been rated at 9.11/10 (previous run: 9.11/10, +0.00)
```

I tried with `python.jediEnabled: true` and same issue happens.

### LINKED ISSUE #1038: Upgrade to LSP 3.0 for better signature handling
3.14.0+, per https://github.com/Microsoft/vscode/issues/68824, to fix https://github.com/Microsoft/python-language-server/issues/679 as well
```cs
		/**
		 * The client supports the following `SignatureInformation`
		 * specific properties.
		 */
		signatureInformation?: {
			/**
			 * The client supports the follow content formats for the documentation
			 * property. The order describes the preferred format of the client.
			 */
			documentationFormat?: MarkupKind[];

			/**
			 * Client capabilities specific to parameter information.
			 */
			parameterInformation?: {
				/**
				 * The client supports processing label offsets instead of a
				 * simple label string.
				 *
				 * Since 3.14.0
				 */
				labelOffsetSupport?: boolean;
```

MikhailArkhipov:
Fixes #1038
Fixes #679 

Inactive until extension upgrades to
```json
  "vscode-languageclient": "^5.2.1",
  "vscode-languageserver": "^5.2.1",
  "vscode-languageserver-protocol": "^3.14.1",
```

### DIFF
--- a/src/LanguageServer/Impl/Protocol/Classes.cs
+++ b/src/LanguageServer/Impl/Protocol/Classes.cs
@@ -278,13 +278,17 @@ namespace Microsoft.Python.LanguageServer.Protocol {
                 /// property.The order describes the preferred format of the client.
                 /// </summary>
                 public string[] documentationFormat;
-
+                
                 /// <summary>
-                /// When true, the label in the returned signature information will
-                /// only contain the function name. Otherwise, the label will contain
-                /// the full signature.
+                /// Client capabilities specific to parameter information.
                 /// </summary>
-                public bool? _shortLabel;
+                public struct ParameterInformationCapabilities {
+                    /// <summary>
+                    ///  The client supports processing label offsets instead of a simple label string
+                    /// </summary>
+                    public bool? labelOffsetSupport;
+                }
+                public ParameterInformationCapabilities? parameterInformation;
             }
             public SignatureInformationCapabilities? signatureInformation;
         }
@@ -567,10 +571,6 @@ namespace Microsoft.Python.LanguageServer.Protocol {
         /// The document version that range applies to.
         /// </summary>
         public int? _version;
-        /// <summary>
-        /// List of fully qualified type names for the expression
-        /// </summary>
-        public string[] _typeNames;
     }
 
     [Serializable]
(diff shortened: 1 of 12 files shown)


src/LanguageServer/Impl/Protocol/Classes.cs
```diff
@@ -585,13 +585,21 @@ public sealed class SignatureInformation {
...
+        /// The label of this signature.
+        /// </summary>
+        /// <remarks>
+        /// LSP before 3.14: string label.
+        /// LSP 3.14.0+: (int, int) range.
+        /// Either a string or inclusive start and exclusive end offsets within its containing
+        /// [signature label] (#SignatureInformation.label). *Note*: A label of type string must be
+        /// a substring of its containing signature information's [label](#SignatureInformation.label).
+        /// </remarks>
+        public object label;
```
AlexanderSher: So they have made a change that breaks backward compatibility?

MikhailArkhipov: Not really. In earlier than 3.14 it is `string`. 3.14+ it is [number, number]. In the initialization info the client tells if it supports offsets rather than strings and server can then provide them. Strings are still accepted in any case.

AlexanderSher: Do we have tests for both cases?

MikhailArkhipov: No. We actually never really verified bolding until now. Should we?

MikhailArkhipov: Let me try add one

MikhailArkhipov: Added

PR #1124, opened 2019-05-23
Fix #975: Analysis continues on old content if file content is replaced by unrelated code

### LINKED ISSUE #975: Analysis continues on old content if file content is replaced by unrelated code
Start with
```python
import plotly
```
when analysis starts running, paste over code from #960
```python
import sys
from collections import defaultdict

def xx():
    dmap = defaultdict(list)
    sizes = {db:int(sz)//(1 << 20) for sz, db in [ln.split() for ln in open(sys.argv[1]).readlines()]}
    return {srv:sum([sizes[d] for d in dbs]) for srv,dbs in dmap.items()}
```
Observed: analysis never completes. If there is progress it stops at the same number of items. Or it continues analysis `plotly`

image

AlexanderSher:


src/Analysis/Ast/Impl/Dependencies/IDependencyChainNode.cs
```diff
@@ -16,6 +16,7 @@
 namespace Microsoft.Python.Analysis.Dependencies {
     internal interface IDependencyChainNode<out TValue> {
         int VertexDepth { get; }
+        bool HasMissingDependencies { get; }
```
AlexanderSher: This property shows if node has any direct or indirect dependencies that aren't added to the graph

MikhailArkhipov: Add `///` comment?

PR #1133, opened 2019-05-28
Drop file content to conserve memory

### LINKED ISSUE #1132: Drop file content to conserve memory after analysis

MikhailArkhipov:
When analysis completes for libraries we no longer need file content. However, we do need to cache some strings for PEP hints to be available if new analysis is requested.

Fixes #1132

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/FunctionEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/FunctionEvaluator.cs
@@ -79,6 +79,10 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
                 if (ctor || annotationType.IsUnknown() || Module.ModuleType == ModuleType.User) {
                     // Return type from the annotation is sufficient for libraries and stubs, no need to walk the body.
                     FunctionDefinition.Body?.Walk(this);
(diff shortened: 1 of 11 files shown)


src/Analysis/Ast/Impl/Analyzer/Symbols/FunctionEvaluator.cs
```diff
@@ -79,6 +79,10 @@ public FunctionEvaluator(ExpressionEval eval, PythonFunctionOverload overload)
                 if (ctor || annotationType.IsUnknown() || Module.ModuleType == ModuleType.User) {
                     // Return type from the annotation is sufficient for libraries and stubs, no need to walk the body.
                     FunctionDefinition.Body?.Walk(this);
+                    // For libraries remove declared local function variables to free up some memory.
+                    if (Module.ModuleType != ModuleType.User) {
```
jakebailey: Will this break goto-def navigation?

MikhailArkhipov: Yes, inside code of the function in a library. Should not affect navigation to a function but navigation and other functionality on variables declared within the function - yes. I am wondering if this should be an option and how to name it then.

Issue #907, opened 2019-04-08
Stubs aren't analyzed before their modules


jakebailey:
This affects PyQt5 pretty heavily.

![image](https://user-images.githubusercontent.com/5341706/55762241-1a176800-5a17-11e9-8cdd-c8d135e29adb.png)

![image](https://user-images.githubusercontent.com/5341706/55762253-21d70c80-5a17-11e9-91b3-17c8fa7abafa.png)

This module has a stub next to it, which was added to the module by `TryCreateModuleStub`, but this stub was never analyzed (the state is at `Analyzing`) and has an empty analysis. This leaves the LS to spend all its in the ModuleSymbolTable with 30k+ evaluators to process.

jakebailey:
The underlying issue for PyQt5 actually seems to be the size; we'll probably start treating library stubs specially to help this (but this issue still stands).

AlexanderSher:
No repro so far. Feel free to reopen it if you see it.

MikhailArkhipov:
I can repro it in tests.

![Uploading image.png…]()

MikhailArkhipov:
I think this might be because producer in the walker has all modules in the same list and may happen to produce module before its stub. Perhaps separate higher priority queue of stubs would help or making graph of stubs for the graph of modules and processing it before modules?

AlexanderSher:
Modules can't be produced before stubs unless there is a loop from some stub to that module, which is unlikely. Analysis will be empty if there is an error in analysis itself. In this case, we don't crash LS or terminate analysis session, but there should be a record in the log.

MikhailArkhipov:
Events:
1. Load user module that has `import logging`
2. Module finds `logging` dependency.
3. `logging` dependency loads stub via 
```cs
            if (!TryCreateModuleStub(name, moduleImport.ModulePath, out var stub)) {
                // If nothing found, try Typeshed.
                stub = _interpreter.TypeshedResolution.GetOrLoadModule(moduleImport.IsBuiltin ? name : moduleImport.FullName);
            }
```
4. Stub is loaded and parsing starts. But since there is no await, parsing and analysis of `logging (Library)` starts in parallel.

MikhailArkhipov:
Add
```cs
Debug.Assert(!(_stubAnalysis is EmptyAnalysis))
```
to `ModuleWalker.MergeStub`. Test:

```python
import types
```
[...]

PR #1134, opened 2019-05-28
Drop AST for libraries after final analysis

### LINKED ISSUE #1088: Remove references to AST from analysis
This is needed for
- types that are backed by persistent objects that no longer have AST nodes
- further reduce memory consumption

MikhailArkhipov:
Fixes #1088 

Includes https://github.com/microsoft/python-language-server/pull/1133

- Remove AST nodes from types
- Cache AST nodes at module level so they cache can be cleared when analysis is done.
- Occasionally analysis is requested again even when previous was claimed to be final, so add code to reload and reparse the file. This may be related to #907.
- Keep AST for stubs generated by the scraper since some functions there require evaluation with arguments
- Remove reference to the AST root from nodes, reducing node size.
- Add `qualified name` to types in preparation for their identification in persistent models


src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerSession.cs
```diff
@@ -356,28 +364,24 @@ private Task StartAnalysis(IDependencyChainNode<PythonAnalyzerEntry> node, Async
             }
         }
 
-        private void LogCompleted(IPythonModule module, Stopwatch stopWatch, TimeSpan startTime) {
-            if (_log != null) {
```
jakebailey: These checks are here to prevent evaluation of all of the arguments and the string building if the logging is null (Alex added them in a previous PR).

MikhailArkhipov: Yes, needed to merge.

jakebailey: I think these still need to be restored.

MikhailArkhipov: Done

jakebailey:
One thing to check, CompletionTest.InImport is consistently failing for me.

```
 InImport
   Source: CompletionTests.cs line: 745
   Duration: 11 sec

  Message: 
    Expected completion list items to have label 'TestCase', but it has none.
  Stack Trace: 
    at LateBoundTestFramework.Throw(String message) in LateBoundTestFramework.cs line: 16
    at AssertionScope.FailWith(String message, Object[] args) in AssertionScope.cs line: 225
    at CompletionResultAssertions.HaveAttribute(IEnumerable`1 attributes, Func`2 attributeSelector, String itemNameSingle, String itemNamePlural, String because, Object[] reasonArgs) in CompletionResultAssertions.cs line: 113
    at CompletionResultAssertions.HaveLabels(IEnumerable`1 labels, String because, Object[] reasonArgs) in CompletionResultAssertions.cs line: 95
    at CompletionResultAssertions.HaveLabels(String[] labels) in CompletionResultAssertions.cs line: 91
    at CompletionTests.InImport() in CompletionTests.cs line: 789

   Open additional output for this result
```

MikhailArkhipov:
Passes on 2017 and 2019, although there is some flakiness in tests that involve imports, I had 3 tests (different) failed but passed on rerun (re compile, datetime delta and some other). This is actually b/c 'analysis ready' event on the module is set when one of the sessions completes. Hence if module is analyzed before stub, as we have in the bug, analysis may end up incorrect as await returns with the analysis without the stub and not with the final one, which happens later. This happens in master as well.

Issue #1142, opened 2019-05-29
Rerun gen.py tests and do comparison


jakebailey:
After #724.

Compare old LS to new LS, pass/fail/not applicable, and report bugs.

Splitting this into a second issue for tracking.

jakebailey:
Done; will compile and submit bugs with repros after discussion.

PR #1146, opened 2019-05-30
Specialize getattr()

### LINKED ISSUE #1140: Support getattr
The old LS supported getattr via a specialization, but this wasn't brought over to the new LS. It shows up via the builtins, but should be equivalent to actually calling GetMember on the argument (plus the overloads) so the correct type info gets carried over.

Found in #724.

jakebailey:
Fixes #1140.
Helps #1142.

### DIFF
--- a/src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
+++ b/src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
@@ -123,7 +123,7 @@ namespace Microsoft.Python.Analysis.Specializations {
             // which is incorrect more often than not.
 
             var args = argSet.Values<IMember>();
-            if (args.Count < 1) {
+            if (args.Count < 2) {
                 return null;
             }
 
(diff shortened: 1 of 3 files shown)


src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -116,5 +116,30 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
             return returnType != null ? new PythonInstance(returnType) : null;
         }
+
+        public static IMember GetAttr(IPythonModule module, IPythonFunctionOverload overload, IArgumentSet argSet) {
+            // TODO: Try __getattr__ first; this may not be as reliable in practice
+            // given we could be assuming that __getattr__ always returns the same type,
+            // which is incorrect more often than not.
+
+            var args = argSet.Values<IMember>();
+            if (args.Count < 1) {
```
MikhailArkhipov: `< 2` per below?

jakebailey: Oops, yes.

PR #1147, opened 2019-05-30
Relocate generated stub cache

### LINKED ISSUE #1094: Relocate scraped stub cache to user data folder
So it does not need to be re-created when extension updates.

MikhailArkhipov:
Fixes #1094 

- Make configurable setting for the scraper cache.
- Pick proper location depending on OS.
- Cache location will be used for analysis persistence.


src/Analysis/Ast/Impl/Caching/CacheFolders.cs
```diff
@@ -0,0 +1,127 @@
...
+
+                if(platform.IsWindows) {
+                    cachePath = defaultCachePath;
+                }
+
+                if (platform.IsMac) {
+                    if (CheckVariableSet(homeVarName, homeFolderPath, logger) 
+                        && CheckPathRooted(homeVarName, homeFolderPath, logger)
+                        && !string.IsNullOrWhiteSpace(homeFolderPath)) {
+                        cachePath = Path.Combine(homeFolderPath, "Library/Caches", plsSubfolder);
```
jakebailey: When we get to .NET standard 2.1, we should probably drop all uses of Combine for Join/TryJoin, as Combine can have strange behavior such that they recommend Join for user-provided inputs. I was going to comment to switch to it now and drop the "is rooted" checks, but then realized that it's not available to us quite yet...

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -48,14 +49,15 @@ public sealed class PythonAnalyzer : IPythonAnalyzer, IDisposable {
         private PythonAnalyzerSession _currentSession;
         private PythonAnalyzerSession _nextSession;
 
-        public PythonAnalyzer(IServiceManager services) {
+        public PythonAnalyzer(IServiceManager services, string cacheFolderPath = null) {
```
jakebailey: Should this argument be used at some `new PythonAnalyzer` site, i.e. with `DatabasePath` from the initialization options? My assumption is that we'd eventually have it removed from the extension in the future when the minimum version gets bumped up.

jakebailey: Or, in testing, since I know we set other cache locations per test as well.

MikhailArkhipov: Yes, setting was lost when I ported code from earlier work. Restored.

src/Analysis/Ast/Test/AnalysisTestBase.cs
```diff
@@ -58,13 +58,13 @@ public abstract class AnalysisTestBase {
...
+            => CreateServicesAsync(TestData.GetTestSpecificRootUri().AbsolutePath, configuration, null, sm);
 
-        protected async Task<IServiceManager> CreateServicesAsync(string root, InterpreterConfiguration configuration, IServiceManager sm = null) {
+        protected async Task<IServiceManager> CreateServicesAsync(string root, InterpreterConfiguration configuration, string stubCacheFolderPath = null, IServiceManager sm = null) {
             configuration = configuration ?? PythonVersions.LatestAvailable;
             configuration.AssertInstalled();
-            Trace.TraceInformation("Cache Path: " + configuration.DatabasePath);
-            configuration.DatabasePath = TestData.GetAstAnalysisCachePath(configuration.Version, true);
+            stubCacheFolderPath = stubCacheFolderPath ?? TestData.GetAstAnalysisCachePath(configuration.Version, true);
+            Trace.TraceInformation("Cache Path: " + stubCacheFolderPath);
```
jakebailey: I know this is here for testing, but can the same thing get printed during a regular run? (like the interpreter path, search paths, etc)

jakebailey: Ignore this, it's printed. I was running the wrong repo...

PR #1149, opened 2019-05-31
Expose event for 'analysis done'

### LINKED ISSUE #901: Create an event on IPythonAnalyzer (or similar) instead of accessing ITelemetryService from analyzer
There's a telemetry event for when the analysis completes on a large number of items; this shouldn't be in the analyzer but in the language server. An event can be added to `IPythonAnalyzer` (or similar) to expose that info. This should mitigate issues like those fixed in #900.

### LINKED ISSUE #1143: Reliable way to determine final analysis pass for all modules
Needed for #472 as well as for few other things such as fully working #1088 since currently some ASTs still linger. Should also improve performance a bit since we can delay some activities, such as linting run until everything is settled.

One possible way is to simulate analysis run, i.e.
- Run all sessions but with only parsing in order determine final session and then
- Run analysis without parsing repeating the above.

Technically end point should be the same.

MikhailArkhipov:
Fixes #901
Fixes #1143
For _ideal solution_ for #1143 we'd need to know completion per module. But I'll file a separate item since it is a bit more complicated.

### DIFF
--- a/src/LanguageServer/Impl/Implementation/Server.Telemetry.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.Telemetry.cs
@@ -1,0 +1,55 @@
+﻿// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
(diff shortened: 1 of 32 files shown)


src/LanguageServer/Impl/Program.cs
```diff
@@ -13,7 +13,7 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
-// #define WAIT_FOR_DEBUGGER
+#define WAIT_FOR_DEBUGGER
```
jakebailey: Remove

MikhailArkhipov: Done

src/LanguageServer/Impl/Implementation/Server.Telemetry.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            var te = new TelemetryEvent {
+                EventName = "python_language_server/analysis_complete", // TODO: Move this common prefix into Core.
+            };
+
+            te.Measurements["privateMB"] = privateMB;
+            te.Measurements["peakPagedMB"] = peakPagedMB;
+            te.Measurements["workingMB"] = workingMB;
+            te.Measurements["elapsedMs"] = e.MillisecondsElapsed;
+            te.Measurements["entries"] = e.ModuleCount;
```
jakebailey: Now that we report a module count instead of the number of entries, it might be a good idea to use `"moduleCount"` as the name here to distinguish.

PR #1153, opened 2019-06-02
Allow exporting of instances from modules and prefer real types to Any

### LINKED ISSUE #998: Intellisense Problems with Python Language Server and tkinter
@GaryA50 commented on [Thu Apr 18 2019](https://github.com/Microsoft/vscode-python/issues/5370)

## Environment data

- VS Code version: 1.33.1
- Extension version (available under the Extensions sidebar): 2019.3.6558
- OS and version: Win 10 Home 1809 and Ubuntu 18.04
- Python version (& distribution if applicable, e.g. Anaconda): 3.6.7
- Relevant/affected Python packages and their versions: tkinter.TkVersion 2.2.4

## Expected behaviour

Compared to Python Language Server behavior, with Jedi enabled, the squiggly lines disappear and hovering over the method names gives information about the method and the expected arguments.

## Actual behaviour

With Jedi disabled, configuration options entered in all caps, (FALSE, CENTER, EW, W, E), are underlined by a squiggly line. Hovering over any of those options causes the pop-up, "Undefined Variable". Also, hovering over any tkinter method name leading up to a left parens causes the pop-up message, "Any". Even with those messages, the script runs as expected.

```
#! /usr/bin/env python3

from tkinter import * 
from tkinter import ttk

root = Tk()

#Callback functions
def cbok():
    print('OK clicked')

def cbcancel():
    root.destroy()
    print('Cancel clicked')

if __name__ == '__main__':
    root.geometry("340x100")
    root.resizable(width=FALSE, height=FALSE)
    root.title("File Location Confirmation")

    lstatus = ttk.Label(root, text='Ensure CSV files are in the child folder named\nCSV_Files', 
        justify=CENTER, anchor=CENTER, borderwidth=5, relief='ridge')
    lstatus.grid_configure(column=0, row=0, padx=5, columnspan=2, sticky=EW)

    bok = ttk.Button(root, text='OK', command=cbok)
    bok.grid_configure(column=0, row=1, padx=5, pady=5, sticky=W)
    bcancel = ttk.Button(root, text='Cancel', command=cbcancel)
    bcancel.grid_configure(column=1, row=1, padx=5, pady=5, sticky=E)

    root.columnconfigure(0, weight=0, minsize=170)
    root.columnconfigure(1, weight=0, minsize=170)
    root.rowconfigure(0, weight=1, minsize=50)
    root.rowconfigure(1, weight=0)

    #WM_DELETE_WINDOW is a result of clicking the X in the upper right of root
    root.protocol("WM_DELETE_WINDOW", cbcancel)

    root.mainloop()

    pass
```

---

@GaryA50 commented on [Fri Apr 19 2019](https://github.com/Microsoft/vscode-python/issues/5370#issuecomment-484721481)

Don't know why the first 19 lines of the script were not picked up as code, the indents removed, and the leading and trailing double underscores deleted from __name__ and __main__. Sorry.

---

@karrtikr commented on [Fri Apr 19 2019](https://github.com/Microsoft/vscode-python/issues/5370#issuecomment-485054022)

Correct me if i am wrong here, but isn't `FALSE` supposed to be `False`. And `CENTER` is supposed to be wrapped around quotes - `'CENTER'`, (similar to `OK` in `text='OK'`).

---

@GaryA50 commented on [Sun Apr 21 2019](https://github.com/Microsoft/vscode-python/issues/5370#issuecomment-485282145)

I was not able to find specifications for what is allowed. As the next best option, I used the opinion of the Python interpreter, i.e. runtime behavior. If there were no runtime exceptions and the behavior was expected, I took it to mean that the options were accepted and processed properly. If no runtime exceptions and behavior was unexpected, I looked for my mistakes.

I did further testing and found that Python Language Server (PLS) will accept all of the upper case options I showed in my original message without flagging them so long as they are within quotes. I also found that it will accept lower case or mixed case in quotes without complaint. If "import tkinter as tk" is added (without the quotes), then it will recognize the form similar to tk.CENTER. All the different forms are accepted by Jedi, too, and run without exceptions.

---

@GaryA50 commented on [Sun Apr 21 2019](https://github.com/Microsoft/vscode-python/issues/5370#issuecomment-485282289)

I'm finding that the "Any" message I mentioned is appearing only with PLS for the root methods. Jedi gives informative messages in those cases, and more information for Label and Button.

MikhailArkhipov:
Fixes #998 

The problem were
- Module member filtering code did not allow export of instances.
- `tkinter` stub prototypes members as `Any` while we actually have better typing from the library (`str`).

Changes
- allow instance export while still restricting re-exporting `typing` members.
- Prefer specific type we discovered in library to `Any`.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -150,9 +150,6 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
             // Add members from this file
             var members = Eval.CurrentScope.Variables.Where(v => v.Source == VariableSource.Declaration || v.Source == VariableSource.Import);
             _class.AddMembers(members, false);
-            // Add members from stub
-            var stubClass = Eval.Module.Stub?.GetMember<IPythonClassType>(_class.Name);
-            _class.AddMembers(stubClass, false);
         }
 
         // Classes and functions are walked by their respective evaluators
--- a/src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
@@ -19,6 +19,7 @@ using System.Linq;
 using Microsoft.Python.Analysis.Analyzer.Evaluation;
 using Microsoft.Python.Analysis.Documents;
 using Microsoft.Python.Analysis.Modules;
+using Microsoft.Python.Analysis.Specializations.Typing;
 using Microsoft.Python.Analysis.Types;
 using Microsoft.Python.Analysis.Types.Collections;
 using Microsoft.Python.Analysis.Values;
@@ -239,6 +240,11 @@ namespace Microsoft.Python.Analysis.Analyzer {
 
                 var sourceVar = Eval.GlobalScope.Variables[v.Name];
                 var sourceType = sourceVar?.Value.GetPythonType();
+                
+                // If stub says 'Any' but we have better type, keep the current type.
+                if (!IsStubBetterType(sourceType, stubType)) {
+                    continue;;
+                }
 
                 // If types are the classes, merge members.
                 // Otherwise, replace type from one from the stub.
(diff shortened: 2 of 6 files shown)


src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -150,9 +150,6 @@ internal sealed class ClassEvaluator : MemberEvaluator {
             // Add members from this file
             var members = Eval.CurrentScope.Variables.Where(v => v.Source == VariableSource.Declaration || v.Source == VariableSource.Import);
             _class.AddMembers(members, false);
-            // Add members from stub
```
jakebailey: Why does this get removed? Shouldn't the stub be merged, or is this being done somewhere else?

MikhailArkhipov: It is added in the `MergeStub` as well. Adding it here was probably a leftover from early code and caused problem - instead of merging types with appropriate comparison this piece was hard-adding methods returning `Any` removing better types we discovered.

PR #1193, opened 2019-06-10
Handle null docstrings in docstring converter

### LINKED ISSUE #1192: null docstring exceptions when converting docstrings
```
Microsoft.Python.LanguageServer.Documentation.DocstringConverter.SplitDocstring(String) in LanguageServer\Impl\Documentation\DocstringConverter.cs:line 505
Microsoft.Python.LanguageServer.Documentation.DocstringConverter..ctor(String) in LanguageServer\Impl\Documentation\DocstringConverter.cs:line 82
Microsoft.Python.LanguageServer.Sources.MarkdownDocumentationSource.FormatDocumentation(String) in LanguageServer\Impl\Sources\MarkdownDocumentationSource.cs:line 71
Microsoft.Python.LanguageServer.Sources.SignatureSource.GetSignature(IDocumentAnalysis, SourceLocation) in LanguageServer\Impl\Sources\SignatureSource.cs:line 99
Microsoft.Python.LanguageServer.Implementation.Server.SignatureHelp(TextDocumentPositionParams, CancellationToken) in LanguageServer\Impl\Implementation\Server.Editor.cs:line 69
```

jakebailey:
Fixes #1192.

### DIFF
--- a/src/LanguageServer/Test/FluentAssertions/DocstringConverterAssertions.cs
+++ b/src/LanguageServer/Test/FluentAssertions/DocstringConverterAssertions.cs
@@ -23,12 +23,12 @@ namespace Microsoft.Python.LanguageServer.Tests.FluentAssertions {
     [ExcludeFromCodeCoverage]
     internal static class DocstringAssertions {
         public static AndConstraint<StringAssertions> ConvertToPlaintext(this StringAssertions a, string plaintext, string because = "", params object[] reasonArgs) {
-            DocstringConverter.ToPlaintext(a.Subject?.NormalizeLineEndings()).Should().Be(plaintext?.NormalizeLineEndings().TrimEnd(), because, reasonArgs);
+            DocstringConverter.ToPlaintext(a.Subject.NormalizeLineEndings()).Should().Be(plaintext.NormalizeLineEndings().TrimEnd(), because, reasonArgs);
             return new AndConstraint<StringAssertions>(a);
         }
 
         public static AndConstraint<StringAssertions> ConvertToMarkdown(this StringAssertions a, string markdown, string because = "", params object[] reasonArgs) {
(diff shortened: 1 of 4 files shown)


src/LanguageServer/Test/FluentAssertions/DocstringConverterAssertions.cs
```diff
@@ -23,12 +23,12 @@ namespace Microsoft.Python.LanguageServer.Tests.FluentAssertions {
     [ExcludeFromCodeCoverage]
     internal static class DocstringAssertions {
         public static AndConstraint<StringAssertions> ConvertToPlaintext(this StringAssertions a, string plaintext, string because = "", params object[] reasonArgs) {
-            DocstringConverter.ToPlaintext(a.Subject.NormalizeLineEndings()).Should().Be(plaintext.NormalizeLineEndings().TrimEnd(), because, reasonArgs);
+            DocstringConverter.ToPlaintext(a.Subject?.NormalizeLineEndings()).Should().Be(plaintext?.NormalizeLineEndings().TrimEnd(), because, reasonArgs);
```
AlexanderSher: Since this is the only place `NormalizeLineEndings` is used, we can check for `null` in the extension method itself.

AlexanderSher: `plaintext` is an expected string, we shouldn't change it in assertion.

AlexanderSher: something like `DocstringConverter.ToPlaintext(a.Subject.NormalizeLineEndings()).Should().Be(plaintext, because, reasonArgs);`

jakebailey: If I don't normalize the expected output, then all of my other tests will need to make this call, as they would depend on the line endings of the source file and the system line endings.

AlexanderSher: Hm... ok, let's leave it as it is.

PR #1224, opened 2019-06-19
Persistent analysis, part I

### LINKED ISSUE #1112: Proxy types backed by index database
To be used when declaring imported items when persistent model is available or when analysis is still running and we used Python informational process.

### LINKED ISSUE #1177: API to determine unique package id
Need a way to unique identify packages, such `name + version + ?` perhaps. If it is a version, then we needs means to reliably determine version of a package. If package is not versioned, perhaps hash the content?

### LINKED ISSUE #1178: Analysis DB schema (model)
Classes for the persistent analysis model.

### LINKED ISSUE #1205: Basic restore saved analysis from db
Initial test to save/restore analysis of builtins module

### LINKED ISSUE #1221: Tell between class/member chain vs module import
Specifically, handle `a.b.c` where `a` is a class with inner class `b` and method `c` vs. case where `a.b.c` is an package and `b` will not appear in members of `b` via `GetMemberNames`

MikhailArkhipov:
Closes #1112 
Closes #1177 
Closes #1178 
Closes #1205 
Closes #1221

Works in basic scenarios like `import sys`. Should be merged in inactive state - remove creation of the database service in Server initialization by commenting out

```cs
_services.AddService(new ModuleDatabase(_services));
```

- Introduces 'qualified name' to types. This is how type is identified in the module saved analysis. For example, `A.B.C` for method `C` inside method `B` in class `A`. `i:mod:z` is an instance of type `z` defined in module `mod`.
- Provides model classes for modules, classes, functions, properties, variables.
- Fixes issues when declaring module for `Union` and collections from `typing` was actually `builtins`.
- Uses LiteDb to save models to file. Currently uses one file per module for convenience of debugging.
- Splits `PythonModule` to partial classes.
- Provides code to recostruct types from qualified names.
- Implements database service to get module constructed from the db instead of running analysis.

See outstanding issues in the epic on what is not handled yet.


src/Analysis/Ast/Impl/Types/Collections/PythonCollectionType.cs
```diff
@@ -32,17 +32,17 @@ internal class PythonCollectionType : PythonTypeWrapper, IPythonCollectionType {
...
+        /// <param name="declaringModule">Declaring module.</param>
         /// <param name="isMutable">Indicates if collection is mutable (like list) or immutable (like tuple).</param>
         public PythonCollectionType(
             string typeName,
             BuiltinTypeId collectionTypeId,
-            IPythonInterpreter interpreter,
+            IPythonModule declaringModule,
             bool isMutable
-        ) : base(collectionTypeId, interpreter.ModuleResolution.BuiltinsModule) {
+            ) : base(collectionTypeId, declaringModule) {
```
AlexanderSher: Is this change related to databases or a separate bug fix?

src/Analysis/Ast/Impl/Specializations/Typing/Values/TypingType.cs
```diff
@@ -28,10 +28,12 @@ internal sealed class TypingType : LocatedMember, IPythonType {
         public TypingType(IPythonModule declaringModule, IPythonType type): base(declaringModule) {
             _type = type ?? throw new ArgumentNullException(nameof(type));
             Name = $"Type[{_type.Name}]";
+            QualifiedName = $"{DeclaringModule.Name}:Type[{_type.QualifiedName}]";
```
AlexanderSher: Can `DeclaringModule` be null or have no name?

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
```diff
@@ -110,14 +110,14 @@ internal sealed partial class ExpressionEval {
...
+                        return PythonCollectionType.CreateList(Module, new[] { v1 });
                     case SetComprehension sc:
                         var v2 = GetValueFromExpression(sc.Item) ?? UnknownType;
-                        return PythonCollectionType.CreateSet(Interpreter, new[] { v2 });
+                        return PythonCollectionType.CreateSet(Module, new[] { v2 });
                     case DictionaryComprehension dc:
                         var k = GetValueFromExpression(dc.Key) ?? UnknownType;
                         var v = GetValueFromExpression(dc.Value) ?? UnknownType;
-                        return new PythonDictionary(new PythonDictionaryType(Interpreter), new Dictionary<IMember, IMember> { { k, v } });
+                        return new PythonDictionary(new PythonDictionaryType(Interpreter.ModuleResolution.BuiltinsModule), new Dictionary<IMember, IMember> { { k, v } });
```
AlexanderSher: Why is it `BuiltinsModule` here and not the `Module`?

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -258,6 +260,10 @@ public ModuleWalker(IServiceContainer services, IPythonModule module)
 
                         var memberType = member?.GetPythonType();
                         var stubMemberType = stubMember.GetPythonType();
+
+                        if (builtins.Equals(memberType?.DeclaringModule) || builtins.Equals(stubMemberType.DeclaringModule)) {
```
AlexanderSher: Why not the `stubMemberType.IsBuiltin`?

src/Analysis/Ast/Impl/Caching/StubCache.cs
```diff
@@ -27,13 +27,15 @@ internal sealed class StubCache : IStubCache {
...
 
-        public StubCache(IServiceContainer services, string cacheRootFolder = null) {
+        public StubCache(IServiceContainer services) {
             _fs = services.GetService<IFileSystem>();
             _log = services.GetService<ILogger>();
+            _cfs = services.GetService<ICacheFolderService>();
 
-            cacheRootFolder = cacheRootFolder ?? CacheFolders.GetCacheFolder(services);
-            StubCacheFolder = Path.Combine(cacheRootFolder, $"stubs.v{_stubCacheFormatVersion}");
+            _cfs = services.GetService<ICacheFolderService>();
```
AlexanderSher: `ICacheFolderService` is accessed twice.

src/Analysis/Ast/Impl/Modules/PythonModule.Analysis.cs
```diff
@@ -0,0 +1,128 @@
...
+using Microsoft.Python.Analysis.Values;
+using Microsoft.Python.Core;
+
+namespace Microsoft.Python.Analysis.Modules {
+    /// <summary>
+    /// Primary base for all modules and user documents. Provides access
+    /// to AST and the module analysis.
+    /// </summary>
+    [DebuggerDisplay("{Name} : {ModuleType}")]
+    internal partial class PythonModule {
```
AlexanderSher: Does this file has any new code?

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
```diff
@@ -99,7 +99,12 @@ public MainModuleResolution(string root, IServiceContainer services)
                     return module;
                 }
             }
-            
+
+            var dbs = _services.GetService<IModuleDatabaseService>();
+            if (dbs != null && dbs.TryCreateModule(name, moduleImport.ModulePath, out var m) != ModuleStorageState.DoesNotExist && m != null) {
```
AlexanderSher: So we are still using this module for `partial`?

src/Analysis/Ast/Impl/Properties/AssemblyInfo.cs
```diff
@@ -15,5 +15,7 @@
...
 [assembly: InternalsVisibleTo("Microsoft.Python.Analysis.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010007d1fa57c4aed9f0a32e84aa0faefd0de9e8fd6aec8f87fb03766c834c99921eb23be79ad9d5dcc1dd9ad236132102900b723cf980957fc4e177108fc607774f29e8320e92ea05ece4e821c0a5efe8f1645c4c0c93c1ab99285d622caa652c1dfad63d745d6f2de5f17e5eaf0fc4963d261c8a12436518206dc093344d5ad293")]
+[assembly: InternalsVisibleTo("Microsoft.Python.Analysis.Caching.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010007d1fa57c4aed9f0a32e84aa0faefd0de9e8fd6aec8f87fb03766c834c99921eb23be79ad9d5dcc1dd9ad236132102900b723cf980957fc4e177108fc607774f29e8320e92ea05ece4e821c0a5efe8f1645c4c0c93c1ab99285d622caa652c1dfad63d745d6f2de5f17e5eaf0fc4963d261c8a12436518206dc093344d5ad293")]
```
AlexanderSher: Do we really need these internals?

src/Analysis/Ast/Impl/Specializations/Typing/Types/AnyType.cs
```diff
@@ -24,6 +24,7 @@ internal sealed class AnyType : LocatedMember, IPythonType {
 
         public override PythonMemberType MemberType => PythonMemberType.Class;
         public string Name => "Any";
+        public string QualifiedName => $"{DeclaringModule.Name}:{Name}";
```
AlexanderSher: Can we make it an extension method to `IPythonType`?

src/Analysis/Ast/Impl/Types/ParameterInfo.cs
```diff
@@ -29,7 +29,7 @@ public ParameterInfo(PythonAst ast, Parameter p, IPythonType type, IMember defau
         }
 
         public ParameterInfo(string name, IPythonType type, ParameterKind? kind, IMember defaultValue) {
-            Name = name ?? throw new ArgumentNullException(nameof(name));
+            Name = name ?? "*"; // ?? throw new ArgumentNullException(nameof(name));
```
AlexanderSher: Is it temporary?

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -196,7 +197,7 @@ internal PythonClassType(string name, Location location)
                 // During processing of builtins module some types may not be available yet.
                 // Specialization will attach proper base at the end.
                 Debug.Assert(!objectType.IsUnknown());
-                if (!bases.Any(b => objectType.Equals(b))) {
+                if (!bases.Any(b => b.Name == "object" && b.DeclaringModule.Name == @"builtins")) {
```
AlexanderSher: Why previous comparison didn't work? We can have two instances of `objectType`?

src/Analysis/Ast/Impl/Types/PythonUnionType.cs
```diff
@@ -42,15 +44,11 @@ internal sealed class PythonUnionType : LocatedMember, IPythonUnionType {
         #region IPythonType
 
         public string Name {
-            get {
-                lock (_lock) {
-                    return CodeFormatter.FormatSequence("Union", '[', _types.ToArray());
-                }
-            }
+            get { lock (_lock) { return CodeFormatter.FormatSequence("Union", '[', _types.ToArray()); } }
```
AlexanderSher: I can't say that it is easier to read one line.

src/Analysis/Ast/Impl/Values/Collections/PythonDictionary.cs
```diff
@@ -26,37 +26,37 @@ namespace Microsoft.Python.Analysis.Values.Collections {
...
-            _interpreter = interpreter;
+            _declaringModule = declaringModule;
         }
 
-        public PythonDictionary(IPythonInterpreter interpreter, IReadOnlyDictionary<IMember, IMember> contents, bool exact = false) :
-            this(new PythonDictionaryType(interpreter), contents, exact: exact) {
-            _interpreter = interpreter;
+        public PythonDictionary(IPythonModule declaringModule, IReadOnlyDictionary<IMember, IMember> contents, bool exact = false) :
+            this(new PythonDictionaryType(declaringModule), contents, exact: exact) {
+            _declaringModule = declaringModule;
```
AlexanderSher: I don't see where we use `_declaringModule` directly. Looks like it can be still `_interpreter = declaringModule.Interpreter`

src/Analysis/Ast/Test/Microsoft.Python.Analysis.Tests.csproj
```diff
@@ -42,6 +42,9 @@
     <ProjectReference Include="..\..\Core\Impl\Microsoft.Python.Analysis.Core.csproj" />
     <ProjectReference Include="..\Impl\Microsoft.Python.Analysis.csproj" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Properties\" />
```
AlexanderSher: Do we need this?

src/PLS.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27929.0
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29003.217
```
AlexanderSher: Are we switching to VS19?

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -142,6 +143,10 @@ public sealed partial class Server : IDisposable {
                 TypeshedPath = @params.initializationOptions.typeStubSearchPaths.FirstOrDefault()
             };
 
+            if (@params.initializationOptions.enableAnalysCache != false) {
+                //_services.AddService(new ModuleDatabase(_services));
```
AlexanderSher: I think we can uncomment this in the branch

src/LanguageServer/Impl/Documentation/DocstringConverter.cs
```diff
@@ -87,7 +87,7 @@ private int NextBlockIndent
         private string CurrentLineWithinBlock => CurrentLine.Substring(_blockIndent);
 
         private DocstringConverter(string input) {
-            _builder = new StringBuilder(input.Length);
+            _builder = new StringBuilder(input?.Length ?? 0);
```
AlexanderSher: This constructor is called only from:
```cs
        public static string ToMarkdown(string docstring) {
            if (string.IsNullOrWhiteSpace(docstring)) {
                return string.Empty;
            }

            return new DocstringConverter(docstring).Convert();
        }
```
so it shouldn't be null

src/Core/Impl/Extensions/StringExtensions.cs
```diff
@@ -302,5 +302,15 @@ public static bool TryGetNextNonEmptySpan(this string s, char separator, ref (in
             lineEnding = lineEnding ?? Environment.NewLine;
             return string.Join(lineEnding, s.SplitLines());
         }
+
+        public static int GetStableHash(this string s) {
```
AlexanderSher: Why not use SHA256?

src/Caching/Test/Microsoft.Python.Analysis.Caching.Tests.csproj
```diff
@@ -0,0 +1,43 @@
...
+    <LangVersion>7.2</LangVersion>
+  </PropertyGroup>
+  <Import Project="..\..\..\build\NetStandard.settings" />
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
+  <ItemGroup>
+    <None Remove="app.config" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.0.0" />
+    <PackageReference Include="Microsoft.Win32.Registry" Version="4.5.0" />
```
AlexanderSher: Registry?

src/LanguageServer/Impl/Documentation/DocstringConverter.cs
jakebailey: Probably just missing a merge of #1193.

src/Caching/Impl/TypeNames.cs
```diff
@@ -0,0 +1,117 @@
...
+
+        /// <summary>
+        /// Splits qualified type name in form of i:A(3.6).B.C into parts. as well as determines if
+        /// qualified name designates instance (prefixed with 'i:').
+        /// </summary>
+        /// <param name="qualifiedName">Qualified name to split. May include instance prefix.</param>
+        /// <param name="moduleName">Module name.</param>
+        /// <param name="memberNames">Module member names such as 'A', 'B', 'C' from module:A.B.C.</param>
+        /// <param name="isInstance">If true, the qualified name describes instance of a type.</param>
+        public static bool DeconstructQualifiedName(string qualifiedName, out string moduleName, out IReadOnlyList<string> memberNames, out bool isInstance) {
```
AlexanderSher: Should we have construction of qualified names here as well?

src/Caching/Impl/PythonDbModule.cs
```diff
@@ -0,0 +1,35 @@
...
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System.Collections.Generic;
+using Microsoft.Python.Analysis.Caching.Models;
+using Microsoft.Python.Analysis.Modules;
+using Microsoft.Python.Core;
+
+namespace Microsoft.Python.Analysis.Caching {
+    internal sealed class PythonDbModule : SpecializedModule {
```
AlexanderSher: Why we are deriving from `SpecializedModule` rather than from `PythonModule`?

src/Caching/Impl/ModuleUniqueId.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Linq;
+using System.Security.Cryptography;
+using Microsoft.Python.Analysis.Core.Interpreter;
+using Microsoft.Python.Analysis.Modules;
+using Microsoft.Python.Analysis.Types;
+using Microsoft.Python.Core;
+using Microsoft.Python.Core.IO;
+
+namespace Microsoft.Python.Analysis.Caching {
+    internal static class ModuleUniqueId {
```
AlexanderSher: Shouldn't it be part of `IModuleDatabaseService`?

src/Caching/Impl/ModuleDatabase.cs
```diff
@@ -0,0 +1,156 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Locates database file based on module information. Module is identified
+        /// by name, version, current Python interpreter version and/or hash of the
+        /// module content (typically file sizes).
+        /// </summary>
+        private string FindDatabaseFile(string moduleName, string filePath) {
+            var interpreter = _services.GetService<IPythonInterpreter>();
```
AlexanderSher: So now we always have one interpreter per LS, right?

MikhailArkhipov: We always actually had... There was notion to have multiple service containers for multi-root workspaces but how actually this is handled is that extension runs multiple language servers, one per root.

src/Caching/Impl/ModuleUniqueId.cs
MikhailArkhipov: Maybe... however, service is for analysis needs and unique id is not used anywhere outside of the caching assembly.

src/Caching/Impl/PythonDbModule.cs
MikhailArkhipov: Just seemed to be a reasonable choice since specialized is already handled in analysis as something that does not have content or AST and specialized functions are not expected to have bodies or otherwise evaluated.

src/Caching/Test/Microsoft.Python.Analysis.Caching.Tests.csproj
MikhailArkhipov: We actually have it in all test assemblies. Might be that test framework is bringing it in?

src/Caching/Impl/PythonDbModule.cs
AlexanderSher: Got it

src/Caching/Impl/ModuleDatabase.cs
AlexanderSher: Ok

src/Caching/Impl/Models/FunctionAttributes.cs
```diff
@@ -15,16 +15,12 @@
 
 using System;
 
-namespace Microsoft.Python.Analysis.Types {
-    public interface IPythonFile {
```
AlexanderSher: So this interface isn't needed any longer?

src/Caching/Impl/Models/ClassModel.cs
```diff
@@ -0,0 +1,96 @@
...
+            // Skip certain members in order to avoid infinite recursion.
+            foreach (var name in cls.GetMemberNames().Except(new[] { "__base__", "__bases__", "__class__", "mro" })) {
+                var m = cls.GetMember(name);
+
+                // Only take members from this class, skip members from bases.
+                if (m is IPythonClassMember cm && !cls.Equals(cm.DeclaringType)) {
+                    continue;
+                }
+
+                if (!_processing.Push(m)) {
```
AlexanderSher: Is it always single-threaded?

src/Core/Impl/Extensions/StringExtensions.cs
MikhailArkhipov: Need an int for Id field required by BSON/LiteDb. SHA256 yields `byte[]`, no? Would need to xor then anyway.

src/LanguageServer/Impl/Implementation/Server.cs
MikhailArkhipov: Done

src/PLS.sln
MikhailArkhipov: Everyone is on it AFAIK... I am on dogfood, works OK, some stuff is better/faster.

src/Analysis/Ast/Test/Microsoft.Python.Analysis.Tests.csproj
MikhailArkhipov: There is `AssemblyInfo` there with `InternalsVisibleTo("Microsoft.Python.Analysis.Caching.Tests)` for fluent things.

src/Core/Impl/Extensions/StringExtensions.cs
AlexanderSher: Ah, right. I thought it should be string.

src/Analysis/Ast/Impl/Values/Collections/PythonDictionary.cs
MikhailArkhipov: OK

src/Analysis/Ast/Impl/Types/PythonUnionType.cs
MikhailArkhipov: OK

src/Analysis/Ast/Impl/Specializations/Typing/Values/TypingType.cs
MikhailArkhipov: No, should never be null. And modules always have names.

src/Analysis/Ast/Impl/Caching/StubCache.cs
MikhailArkhipov: Fixed

src/Analysis/Ast/Impl/Modules/PythonModule.Analysis.cs
MikhailArkhipov: No, just split

src/Analysis/Ast/Impl/Specializations/Typing/Values/TypingType.cs
AlexanderSher: Can we get here from unsaved file?

src/Analysis/Ast/Impl/Properties/AssemblyInfo.cs
MikhailArkhipov: Fluent extensions come handy

src/Analysis/Ast/Impl/Specializations/Typing/Types/AnyType.cs
MikhailArkhipov: Yes

src/Caching/Impl/Models/FunctionAttributes.cs
MikhailArkhipov: It was never used anywhere, doesn't seem to provide any value.

AlexanderSher: Ok

src/Caching/Test/Microsoft.Python.Analysis.Caching.Tests.csproj
AlexanderSher: Maybe a copy/paste error.

src/PLS.sln
AlexanderSher: Ok, I was still on VS15

src/Analysis/Ast/Test/Microsoft.Python.Analysis.Tests.csproj
AlexanderSher: Ok

src/Analysis/Ast/Impl/Properties/AssemblyInfo.cs
AlexanderSher: Ok

src/Analysis/Ast/Impl/Modules/PythonModule.Analysis.cs
AlexanderSher: Ok

src/LanguageServer/Impl/Documentation/DocstringConverter.cs
MikhailArkhipov: Yes

src/Caching/Test/Microsoft.Python.Analysis.Caching.Tests.csproj
MikhailArkhipov: Removed.

src/Analysis/Ast/Impl/Types/Collections/PythonCollectionType.cs
MikhailArkhipov: Yes and yes. Collections were assumed to be builtins like `list`, but `List[str]` is actually `typing` and restoration from the db didn't work since we ended up with `builtins:List[str]` that does not make sense.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
MikhailArkhipov: `dict` technically should be `builtins:dict` since it is where the type is declared.

src/Analysis/Ast/Impl/Types/ParameterInfo.cs
MikhailArkhipov: The problem is that sometimes argument does not have name. This happens to `*` arguments since parser strips the `*`. So I'd like to have at least something.

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
MikhailArkhipov: `Partial` may or may not happen. The idea was to use Python process to obtain initial info on the module members quickly rather than wait for analysis. Consider `plotly`, for example. So we get quick members from Python, then, when/if analysis completes (user may as well close the app before it finishes), write the actual info. However, I am not 100% confident we'll do it, but we might still.

src/Analysis/Ast/Impl/Specializations/Typing/Values/TypingType.cs
MikhailArkhipov: Yes but it is specialized typing so it is always actually `typing`.

src/Caching/Impl/Models/ClassModel.cs
MikhailArkhipov: Should be as we should be writing as part of the analysis so as long as analysis pass is single and there is only one event leading to the db writing.

src/Caching/Impl/TypeNames.cs
MikhailArkhipov: We sort of do, but it is different for modules, types and class/functions.

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
MikhailArkhipov: Yep

src/Analysis/Ast/Impl/Types/PythonClassType.cs
MikhailArkhipov: This probably can be reverted. The problem was when real analysis had one builtins module and restored code had another one.

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
MikhailArkhipov: Actually no. We have a bit of a mess with what is `IsBuiltin` on the type. Ex `PythonModule` has `IsBuiltin => true` always. So does `PythonVariableModule`. Or `Union` when it has all types as builtin types... Modules assume they are builtin if `IsBuiltin => IsCompiled && ModulePath == null`.

This specific check is if member actually belongs to the `builtins` module. We should go clean this up, but it may be some other time.

src/Caching/Impl/TypeNames.cs
AlexanderSher: Ok

PR #1227, opened 2019-06-20
Lock document content 

MikhailArkhipov:
Since it is locked in other cases when we are resetting or reloading.


src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -235,7 +235,14 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
         /// <summary>
         /// Returns module content (code).
         /// </summary>
-        public string Content => _buffer.Text;
+        public string Content {
+            get {
+                lock (AnalysisLock) {
```
jakebailey: This seems spooky to me since `Reset` uses `Content` under `AnalysisLock`, but I guess if that's how C# locks work...

PR #1230, opened 2019-06-20
Adding diagnostic message when using TypeVar improperly

CTrando:


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.cs
```diff
@@ -108,7 +108,7 @@ public ExpressionEval(IServiceContainer services, IPythonModule module)
 
         public void ReportDiagnostics(Uri documentUri, DiagnosticsEntry entry) {
             // Do not add if module is library, etc. Only handle user code.
-            if (Module.ModuleType == ModuleType.User) {
+            if (Module.ModuleType == ModuleType.User || Module.ModuleType == ModuleType.Specialized) {
```
MikhailArkhipov: Should not be needed. We only report diagnostics in user code.

src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
```diff
@@ -25,5 +25,6 @@ public static class ErrorCodes {
         public const string UndefinedVariable = "undefined-variable";
         public const string VariableNotDefinedGlobally= "variable-not-defined-globally";
         public const string VariableNotDefinedNonLocal = "variable-not-defined-nonlocal";
+        public const string TypeVarNoArguments = "type-var-no-arguments";
```
MikhailArkhipov: A bit too generic since there may be diagnostics on constraints (TypeVar second argument). Probably `typevar` as a single word. `typevar-missing-first-argument` will do.

src/Analysis/Ast/Impl/Resources.resx
```diff
@@ -189,4 +189,7 @@
   <data name="UnableToDetermineCachePathException" xml:space="preserve">
     <value>Unable to determine analysis cache path. Exception: {0}. Using default '{1}'.</value>
   </data>
+  <data name="TypeVarNoArguments" xml:space="preserve">
+    <value>TypeVar must have at least one argument.</value>
```
MikhailArkhipov: Should be a bit more specific. What if user types `T = TypeVar(int, float, complex)` missing `T` per https://www.python.org/dev/peps/pep-0484? The message will then be confusing since there are 3 arguments there. It should be something like `TypeVar requires first argument to be a string with the same name as the variable.`. Jake may suggest something too.

src/Analysis/Ast/Test/GenericsTests.cs
```diff
@@ -642,5 +642,20 @@ class LoggedVar(Generic[T]):
             var analysis = await GetAnalysisAsync(code);
             analysis.Should().HaveVariable("x").OfType(BuiltinTypeId.Int);
         }
+
+        [TestMethod, Priority(0)]
```
MikhailArkhipov: Make separate test class (and file) like for `LintTypeVarTests`, like we have `LintUndefinedVarsTests`

src/Analysis/Ast/Test/GenericsTests.cs
```diff
@@ -642,5 +642,20 @@ class LoggedVar(Generic[T]):
...
+        public async Task TypeVarNoArguments() {
+            const string code = @"
+from typing import TypeVar
+
+T = TypeVar()
```
MikhailArkhipov: Should also check other variants like `T = TypeVar(int, float, complex)`, see PEP spec https://www.python.org/dev/peps/pep-0484. Consider `DataTestMethod` with `DataRow` and various code samples.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.cs
CTrando: Sure that's my bad.

src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
CTrando: Ok I can change the name.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
```diff
@@ -33,11 +35,21 @@ public GenericTypeParameter(string name, IPythonModule declaringModule, IReadOnl
...
-
         public static IPythonType FromTypeVar(IArgumentSet argSet, IPythonModule declaringModule, IndexSpan location = default) {
             var args = argSet.Values<IMember>();
+            var eval = argSet.Eval;
+
             if (args.Count == 0) {
-                // TODO: report that at least one argument is required.
+                var sourceLocation = argSet.CallExpression.GetLocation(eval.Module);
+
+                eval?.ReportDiagnostics(
```
AlexanderSher: we aren't checking `argSet.Eval` for null anywhere in this method.

src/Analysis/Ast/Impl/Resources.resx
CTrando: This is actually handled by another case, this diagnostic message is specifically for when TypeVar is called with no arguments.

src/Analysis/Ast/Test/GenericsTests.cs
CTrando: Sounds good.

src/Analysis/Ast/Impl/Resources.resx
MikhailArkhipov: May be too granular. Users probably don't need that much granularity.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
CTrando: Yup, I'll fix that

src/Analysis/Ast/Test/GenericsTests.cs
CTrando: Ok I added some more tests.

src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
```diff
@@ -25,5 +25,6 @@ public static class ErrorCodes {
         public const string UndefinedVariable = "undefined-variable";
         public const string VariableNotDefinedGlobally= "variable-not-defined-globally";
         public const string VariableNotDefinedNonLocal = "variable-not-defined-nonlocal";
+        public const string TypeVarLint = "typevar-linter";
```
MikhailArkhipov: Something more descriptive since this is what user would use to enable/disable the diagnostic. `typevar-arguments`

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
```diff
@@ -33,17 +35,35 @@ public GenericTypeParameter(string name, IPythonModule declaringModule, IReadOnl
...
         public override bool IsSpecialized => true;
 
-
         public static IPythonType FromTypeVar(IArgumentSet argSet, IPythonModule declaringModule, IndexSpan location = default) {
             var args = argSet.Values<IMember>();
+            var eval = argSet.Eval;
+            var callExpression = argSet.CallExpression; 
+            var callLocation = callExpression?.GetLocation(eval?.Module);
+
             if (args.Count == 0) {
```
MikhailArkhipov: Since method is getting long, how's about making a separate method like `CheckTypeVarArguments(...)`

src/Analysis/Ast/Impl/Types/Definitions/IArgumentSet.cs
```diff
@@ -141,5 +141,10 @@ public interface IArgumentSet {
         /// Evaluator associated with the set.
         /// </summary>
         IExpressionEvaluator Eval { get; }
+        
+        /// <summary>
+        /// Expression in how argument set was called
```
MikhailArkhipov: `Call expression that invoked function or method with this argument set`

CTrando: I agree, mine was pretty informal.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
CTrando: Done!

src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
CTrando: Done!

src/Analysis/Ast/Test/LintTypeVarTests.cs
```diff
@@ -0,0 +1,114 @@
...
+            const string code = @"
+from typing import TypeVar
+
+T = TypeVar()
+";
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Diagnostics.Should().HaveCount(1);
+
+            var diagnostic = analysis.Diagnostics.ElementAt(0);
+            diagnostic.ErrorCode.Should().BeEquivalentTo(Diagnostics.ErrorCodes.TypeVarArguments);
```
jakebailey: I think `Be` is enough here. Equivalent is for when you have to do extra work, like nested comparisons.

src/Analysis/Ast/Test/LintTypeVarTests.cs
```diff
@@ -0,0 +1,114 @@
...
+            diagnostic.Message.Should().BeEquivalentTo(Resources.TypeVarMissingFirstArgument);
+        }
+
+        [DataRow(Utils.TYPEVAR_IMPORT + @"
```
jakebailey: When I said you could simplify this, I had meant you could move this down into the function below, i.e. `code = "from typing import TypeVar\b" + code` or similar. Making a static constant in Utils doesn't seem like the right thing to do.

src/Analysis/Ast/Test/LintTypeVarTests.cs
```diff
@@ -0,0 +1,114 @@
...
+using Microsoft.Python.Analysis.Tests;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using TestUtilities;
+
+namespace Microsoft.Python.Analysis.Tests {
+
+    [TestClass]
+    public class LintTypeVarTests : AnalysisTestBase {
+
+        internal static class Utils {
```
jakebailey: I would delete this whole thing, but a couple style comments:

You can just put this at the top level, you don't need some hidden static class for this.

C# constants don't use the ALL_CAPS_NAME format, that's a Java thing. You'd probably do `TypeVarImport` and leave it at that.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
```diff
@@ -33,27 +35,67 @@ public GenericTypeParameter(string name, IPythonModule declaringModule, IReadOnl
...
+                    new DiagnosticsEntry(Resources.TypeVarFirstArgumentNotString,
+                    firstArgLocation?.Span ?? default,
+                    Diagnostics.ErrorCodes.TypeVarArguments,
+                    Severity.Warning, DiagnosticSource.Analysis)
+                );
+
+                return false;
+            }
+
+            if (args.Skip(1).Count() == 1) {
```
MikhailArkhipov: Since `ArgumentSet.Arguments` is `IReadOnlyList` count is known so you can simply do `args.Count-1 == 1`

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
```diff
@@ -33,27 +35,67 @@ public GenericTypeParameter(string name, IPythonModule declaringModule, IReadOnl
...
+        private static bool TypeVarArgumentsValid(IArgumentSet argSet) {
             var args = argSet.Values<IMember>();
+            var eval = argSet.Eval;
+            var callExpression = argSet.CallExpression;
+            var callLocation = callExpression?.GetLocation(eval?.Module);
+
             if (args.Count == 0) {
-                // TODO: report that at least one argument is required.
-                return declaringModule.Interpreter.UnknownType;
+                eval?.ReportDiagnostics(
```
MikhailArkhipov: I don't think it is possible to have `eval == null` here. It is not possible to run analysis without an evaluator.

src/Analysis/Ast/Test/LintTypeVarTests.cs
CTrando: Ok thanks for the tips!

src/Analysis/Ast/Test/LintTypeVarTests.cs
CTrando: I would still like to keep it as a constant so I don't have to redefine it in each test method, but I'll make a constant in the class level scope instead of a static class, don't really know why I didn't do that in the first place.

src/Analysis/Ast/Test/LintTypeVarTests.cs
CTrando: Thanks!

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
```diff
@@ -33,27 +35,67 @@ public GenericTypeParameter(string name, IPythonModule declaringModule, IReadOnl
...
+                    new DiagnosticsEntry(Resources.TypeVarFirstArgumentNotString,
+                    firstArgLocation?.Span ?? default,
+                    Diagnostics.ErrorCodes.TypeVarArguments,
+                    Severity.Warning, DiagnosticSource.Analysis)
+                );
+
+                return false;
+            }
+
+            if (args.Count - 1 == 1) {
```
jakebailey: Why not just `args.Count == 0`?

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
CTrando: Thanks.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
jakebailey: Err, I mean `== 2`.

CTrando: So when calling TypeVar, we have something like T = TypeVar('T', int). So the length of args is 2. Python gives a runtime error if there is only one constraint, (int in this case), so we exclude the first initial 'T' from our count, which is why we subtract 1.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
CTrando: Ok I have changed it.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
jakebailey: I know, I just mean from a style point of view, checking `x - 1 == 1` versus `x == 2`.

CTrando: Yup didn't see your earlier message, I changed it.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
```diff
@@ -33,27 +35,68 @@ public GenericTypeParameter(string name, IPythonModule declaringModule, IReadOnl
...
+            var name = (args[0] as IPythonConstant)?.GetString();
             var constraints = args.Skip(1).Select(a => {
                 // Type constraints may be specified as type name strings.
                 var typeString = (a as IPythonConstant)?.GetString();
                 return !string.IsNullOrEmpty(typeString) ? argSet.Eval.GetTypeFromString(typeString) : a.GetPythonType();
             }).ToArray();
+
             if (constraints.Any(c => c.IsUnknown())) {
-                // TODO: report that some constraints could not be resolved.
+                // TODO: report that some constraints could not be resolved.
```
jakebailey: Nit: Stray space changed at the end here.

src/Analysis/Ast/Impl/Resources.resx
jakebailey: I think having a no-arg case is alright.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
CTrando: Done!

CTrando:
After more thought, should leave this PR for now.

We have the current issue:

```
T = TypeVar('T', bound='Comparator')
```
```bound =  ...``` essentially means that T must be a subtype of whatever was passed in. I believe bound is actually a keyword only argument, which we do not handle in the instantiation of the Arguments in the constructor in ArgumentSet. 

Currently, bound is seen as a second parameter to TypeVar, which we flag as a bad type because (in our example) 'Comparator' is a string value and not the string type which is required. 

For an example look at test ```TypeVarStringConstraint ```.

**Comment by CTrando** [comment]
Delayed until #1263 merges.

File: src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs [hunk 17]
```diff
@@ -33,32 +36,63 @@ public GenericTypeParameter(string name, IPythonModule declaringModule, IReadOnl
...
+            var constraintArgs = argSet.ListArgument?.Values ?? Array.Empty<IMember>();
+
+            var name = (args[0].Value as IPythonConstant)?.GetString();
             var constraints = constraintArgs.Select(a => {
                 // Type constraints may be specified as type name strings.
                 var typeString = (a as IPythonConstant)?.GetString();
                 return !string.IsNullOrEmpty(typeString) ? argSet.Eval.GetTypeFromString(typeString) : a.GetPythonType();
             }).ToArray() ?? Array.Empty<IPythonType>();
 
-            if (constraints.Any(c => c.IsUnknown())) {
```
CTrando: Some constraints can be valid but still unknown, probably not worth reporting.

PR #1248, opened 2019-06-21
Give Diagnostic messages on improper usage of Generic

CTrando:
Gives a message when users do 

```
class M(Generic[T, T])
```

with duplicate type parameters or when they mix type parameters with concrete types.

```
class M(Generic[T, str])
```


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
```diff
@@ -59,40 +63,62 @@ internal sealed partial class ExpressionEval {
...
-            var genericTypeArgs = indices.OfType<IGenericTypeDefinition>().ToArray();
-            var specificTypes = indices.Where(i => !(i is IGenericTypeDefinition)).OfType<IPythonType>().ToArray();
-
-            if (genericTypeArgs.Length > 0 && genericTypeArgs.Length != indices.Count) {
-                // TODO: report that some type arguments are not declared with TypeVar.
+        private bool GenericClassParameterValid(IGenericTypeDefinition[] genericTypeArgs, IReadOnlyList<IMember> indices, Expression expr) {
+            if (genericTypeArgs.Length == 0) {
+                return false;
             }
-            if (specificTypes.Length > 0 && specificTypes.Length != indices.Count) {
```
CTrando: Really wasn't sure what error this TODO was talking about, could someone explain?

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
```diff
@@ -59,40 +63,62 @@ internal sealed partial class ExpressionEval {
...
+        /// </summary>
+        private IMember CreateSpecificTypeFromIndex(IGenericType gt, IReadOnlyList<IMember> args, Expression expr) {
+            var genericTypeArgs = args.OfType<IGenericTypeDefinition>().ToArray();
+
             if (gt.Name.EqualsOrdinal("Generic")) {
-                // Generic[T1, T2, ...] expression. Create generic base for the class.
-                if (genericTypeArgs.Length > 0) {
-                    return new GenericClassParameter(genericTypeArgs, Module);
-                } else {
-                    // TODO: report too few type arguments for Generic[].
```
CTrando: When there are no arguments to Generic[] then there is a syntax error.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
```diff
@@ -59,40 +63,62 @@ internal sealed partial class ExpressionEval {
...
+                return new GenericClassParameter(genericTypeArgs, Module);
             }
 
             // For other types just use supplied arguments
-            if (indices.Count > 0) {
-                return gt.CreateSpecificType(new ArgumentSet(indices));
+            if (args.Count > 0) {
+                return gt.CreateSpecificType(new ArgumentSet(args));
             }
-            // TODO: report too few type arguments for the generic expression.
```
CTrando: Same here, when there are no arguments to a generic structure like List[str] being List[] that is a syntax error.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
MikhailArkhipov: I've sent you an email. This one is a separate item. ex `class A(Generic[]):`

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
MikhailArkhipov: `List` should be handled in its specialization. See `CreateListType` and others

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
jakebailey: I said (and we observed) that it was a syntax error and probably didn't need to be handled, but during the analysis I think we'd actually have partial results and end up with an empty indexing (which we would want in order to provide completion), so my impression was wrong here.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
CTrando: Ok I've found it, thanks!

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
jakebailey: So long as we don't bail out of the generic code due to the missing expression, we shouldn't send a diagnostic for this. We looked at how the parser behaves, and the index expression is an error expression (and would have already had a diagnostic earlier). Doing it again probably doesn't provide any value.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
jakebailey: Is this question still active?

src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
```diff
@@ -25,5 +25,6 @@ public static class ErrorCodes {
         public const string UndefinedVariable = "undefined-variable";
         public const string VariableNotDefinedGlobally= "variable-not-defined-globally";
         public const string VariableNotDefinedNonLocal = "variable-not-defined-nonlocal";
+        public const string GenericArguments = "generic-arguments";
```
jakebailey: TypingGenericArguments and `typing-generic-arguments`.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
CTrando: I think it was just an oversight, I don't think it actually matters.

src/Analysis/Ast/Test/LintGenericTests.cs
```diff
@@ -0,0 +1,135 @@
...
+        pass
+")]
+        [DataTestMethod, Priority(0)]
+        public async Task GenericDuplicateArguments(string code) {
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Diagnostics.Should().HaveCount(1);
+
+            var diagnostic = analysis.Diagnostics.ElementAt(0);
+            diagnostic.ErrorCode.Should().Be(Diagnostics.ErrorCodes.TypingGenericArguments);
+            diagnostic.Message.Should().Be(Resources.GenericNotAllUnique);
```
jakebailey: Check span

src/Analysis/Ast/Test/LintGenericTests.cs
```diff
@@ -1,21 +1,25 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using FluentAssertions;
```
jakebailey: Usings should be sorted.

PR #1254, opened 2019-06-24
Adding diagnostic error on binary operations with incompatible types

CTrando:
I basically hooked into what Jake is doing; we agreed that it is probably best to only give diagnostic error only on the builtin Python types. I structured the error message to be nearly the same as the Python interpreter's runtime error message, but that is subject to change.

For testing, I feel that a lot of the functionality is related to Jake's parsing of binary operations, so I should rely on the tests for that part of the software to prove the correctness of the diagnostic message instead of trying to recreate new tests that may or may not be duplicates of his.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
@@ -25,16 +25,6 @@ using ErrorCodes = Microsoft.Python.Analysis.Diagnostics.ErrorCodes;
 
 namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
     internal sealed partial class ExpressionEval {
-
-        private void ReportOperatorDiagnostics(Expression expr, IPythonType leftType, IPythonType rightType, PythonOperator op) {
-            ReportDiagnostics(Module.Uri, new DiagnosticsEntry(
-                Resources.UnsupporedOperandType.FormatInvariant(op.ToCodeString(), leftType.Name, rightType.Name),
-                GetLocation(expr).Span,
-                ErrorCodes.UnsupportedOperandType,
-                Severity.Error,
-                DiagnosticSource.Analysis));
-        }
-
         private IMember GetValueFromUnaryOp(UnaryExpression expr) {
             switch (expr.Op) {
                 case PythonOperator.Not:
@@ -450,5 +440,14 @@ namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
 
             return (null, null);
         }
+        private void ReportOperatorDiagnostics(Expression expr, IPythonType leftType, IPythonType rightType, PythonOperator op) {
+            ReportDiagnostics(Module.Uri, new DiagnosticsEntry(
+                Resources.UnsupporedOperandType.FormatInvariant(op.ToCodeString(), leftType.Name, rightType.Name),
+                GetLocation(expr).Span,
+                ErrorCodes.UnsupportedOperandType,
+                Severity.Error,
+                DiagnosticSource.Analysis));
+        }
+
     }
 }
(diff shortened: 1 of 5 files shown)


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
```diff
@@ -13,16 +13,28 @@
...
 using Microsoft.Python.Analysis.Values;
+using Microsoft.Python.Core;
 using Microsoft.Python.Parsing;
 using Microsoft.Python.Parsing.Ast;
+using ErrorCodes = Microsoft.Python.Analysis.Diagnostics.ErrorCodes;
 
 namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
     internal sealed partial class ExpressionEval {
+
+        private void ReportOperatorDiagnostics(Expression expr, IPythonType leftType, IPythonType rightType, PythonOperator op) {
```
jakebailey: I'd move down to the bottom instead of having it at the top. When you open this file, you probably want to see the operator eval code first and not the diagnostics.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
```diff
@@ -13,16 +13,28 @@
...
-using System.Collections.Generic;
+using Microsoft.Python.Analysis.Diagnostics;
 using Microsoft.Python.Analysis.Modules;
 using Microsoft.Python.Analysis.Types;
 using Microsoft.Python.Analysis.Types.Collections;
 using Microsoft.Python.Analysis.Values;
+using Microsoft.Python.Core;
 using Microsoft.Python.Parsing;
 using Microsoft.Python.Parsing.Ast;
+using ErrorCodes = Microsoft.Python.Analysis.Diagnostics.ErrorCodes;
```
jakebailey: Does `ErrorCodes` not work directly?

src/Analysis/Ast/Impl/Resources.resx
```diff
@@ -189,4 +189,7 @@
   <data name="UnableToDetermineCachePathException" xml:space="preserve">
     <value>Unable to determine analysis cache path. Exception: {0}. Using default '{1}'.</value>
   </data>
+  <data name="UnsupporedOperandType" xml:space="preserve">
+    <value>Unsupported operand type(s) for '{0}': '{1}' and '{2}'</value>
```
jakebailey: There are always two types listed, so you can just write `types` instead of `type(s)`.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
CTrando: This is the same syntax used in other places where diagnostics are added.

src/Analysis/Ast/Impl/Resources.resx
CTrando: Ok!

src/Analysis/Ast/Test/LintOperatorTests.cs
```diff
@@ -0,0 +1,79 @@
...
+        public async Task IncompatibleTypesBasic() {
+            var code = $@"
+a = 5 + 'str'
+";
+
+            var analysis = await GetAnalysisAsync(code, PythonVersions.LatestAvailable3X);
+            analysis.Diagnostics.Should().HaveCount(1);
+
+            var diagnostic = analysis.Diagnostics.ElementAt(0);
+            diagnostic.ErrorCode.Should().Be(Diagnostics.ErrorCodes.UnsupportedOperandType);
```
jakebailey: I think you should also be checking that the diagnostic shows up in the right place with the right span.

CTrando: It was a bit annoying to do it but I ended up adding those tests.

PR #1258, opened 2019-06-26
add ICompletionExtension2 accepting CompletionList as parameter

### LINKED ISSUE #1256: [Feature Request] Ability to add completion items in an extension
I have been sorting and altering display texts of completion items with an Extension which implements ICompletionExtension.HandleCompletionAsync (used here: https://github.com/microsoft/python-language-server/blob/master/src/LanguageServer/Impl/Implementation/Server.Editor.cs#L45). It worked very well until now I need to insert some new completion items into the array. Since the completion item list is passed as an array: `res.items.OfType<CompletionItemEx>().ToArray()`, I cannot add new items into `res`.

The reason why I cannot adding new completion items in VSCode's extension side, is because I need to first check if the language server already supplies such completion to avoid duplicates.

I am thinking about adding a "HandleCompletionAsync2" which passes the `CompletionList` object instead of the array.

felixhao28:
close #1256

### DIFF
--- a/src/LanguageServer/Impl/Implementation/Server.Editor.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.Editor.cs
@@ -42,8 +42,14 @@ namespace Microsoft.Python.LanguageServer.Implementation {
                 var result = _completionSource.GetCompletions(analysis, @params.position);
                 res.items = result?.Completions?.ToArray() ?? Array.Empty<CompletionItem>();
 
-                await InvokeExtensionsAsync((ext, token)
-                    => (ext as ICompletionExtension)?.HandleCompletionAsync(analysis, @params.position, res.items.OfType<CompletionItemEx>().ToArray(), cancellationToken), cancellationToken);
+                await InvokeExtensionsAsync(async (ext, token)
+                    => {
+                        if (ext is ICompletionExtension2) {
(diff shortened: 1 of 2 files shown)


src/LanguageServer/Impl/Implementation/Server.Editor.cs
```diff
@@ -42,8 +42,14 @@ public sealed partial class Server {
                 var result = _completionSource.GetCompletions(analysis, @params.position);
                 res.items = result?.Completions?.ToArray() ?? Array.Empty<CompletionItem>();
 
-                await InvokeExtensionsAsync((ext, token)
-                    => (ext as ICompletionExtension)?.HandleCompletionAsync(analysis, @params.position, res.items.OfType<CompletionItemEx>().ToArray(), cancellationToken), cancellationToken);
+                await InvokeExtensionsAsync(async (ext, token)
+                    => {
+                        if (ext is ICompletionExtension2) {
```
jakebailey: A pattern matching switch case would probably be better here, rather checking `is` then `as`. (Or even `if (ext is ICompletionExtension e) { await e.Handle...`)

felixhao28: Refactored. Thx.

PR #1259, opened 2019-06-26
Trying to augment argument set with Evaluator and Expression context

CTrando:


src/Analysis/Ast/Impl/Values/Collections/PythonDictionary.cs
```diff
@@ -62,7 +62,7 @@ public IReadOnlyList<IPythonCollection> Items
             _contents.TryGetValue(key, out var value) ? value : UnknownType;
 
         public override IPythonIterator GetIterator() =>
-            Call(@"iterkeys", ArgumentSet.Empty) as IPythonIterator ?? new EmptyIterator(Type.DeclaringModule.Interpreter.UnknownType);
+            Call(@"iterkeys", null) as IPythonIterator ?? new EmptyIterator(Type.DeclaringModule.Interpreter.UnknownType);
```
jakebailey: null?

CTrando: ah my bad should be ArgumentSet.EmptyNoContext

src/Analysis/Ast/Impl/Types/ArgumentSet.cs
```diff
@@ -38,7 +39,7 @@ internal sealed class ArgumentSet : IArgumentSet {
         private readonly DictArg _dictArgument;
         private bool _evaluated;
 
-        public static IArgumentSet Empty = new ArgumentSet();
+        public static IArgumentSet EmptyNoContext = new ArgumentSet();
```
jakebailey: Perhaps just `WithoutContext`, such that it's used as `ArgumentSet.WithoutContext`.

src/Analysis/Ast/Impl/Types/ArgumentSet.cs
```diff
@@ -49,19 +50,38 @@ internal sealed class ArgumentSet : IArgumentSet {
...
+        /// </summary>
+        /// <param name="expr">Expression associated with argument set.</param>
+        /// <param name="eval">Evaluator for the expression involving the argument set.</param>
+        /// <returns></returns>
+        public static ArgumentSet Empty(Expression expr, IExpressionEvaluator eval) {
+            return new ArgumentSet(new List<IMember>(), expr, eval);
         }
+        
+        /// <summary>
+        /// Creates a set of arguments for a constructor call, an indexing call, etc.
```
jakebailey: Just "a call".

src/Analysis/Ast/Impl/Types/ArgumentSet.cs
CTrando: Done!

src/Analysis/Ast/Impl/Types/PythonFunctionOverload.cs
```diff
@@ -167,7 +167,7 @@ public PythonFunctionOverload(IPythonClassMember cm, FunctionDefinition fd, Loca
             }
 
             if (typeArgs != null) {
-                var specificReturnValue = returnClassType.CreateSpecificType(new ArgumentSet(typeArgs));
+                var specificReturnValue = returnClassType.CreateSpecificType(new ArgumentSet(typeArgs, args.Expression, args.Eval));
```
jakebailey: If you run MethodFromBaseClass, this fails here, seemingly because `args` is null.

jakebailey: GetReturnDocumentation calls with `null` args explicitly, which triggers the if statement above. I think it's probably on purpose, so maybe just `args?` here is good enough.

CTrando: Thanks! Fixed.

PR #1260, opened 2019-06-26
When NewType is called and the first argument is not a string, make a diagnostic message

CTrando:
For example:

```
from typing import NewType

T = NewType(5, int)
```

Now gives a diagnostic error. 

This PR relies on work done in PR #1259, so it should be merged after that.

### DIFF
--- a/src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
+++ b/src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
@@ -25,6 +25,6 @@ namespace Microsoft.Python.Analysis.Diagnostics {
         public const string UndefinedVariable = "undefined-variable";
         public const string VariableNotDefinedGlobally= "variable-not-defined-globally";
         public const string VariableNotDefinedNonLocal = "variable-not-defined-nonlocal";
-        public const string NewTypeArguments = "newtype-arguments";
+        public const string TypingNewTypeArguments = "typing-newtype-arguments";
     }
 }
--- a/src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
+++ b/src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
@@ -25,5 +25,6 @@ namespace Microsoft.Python.Analysis.Diagnostics {
         public const string UndefinedVariable = "undefined-variable";
         public const string VariableNotDefinedGlobally= "variable-not-defined-globally";
         public const string VariableNotDefinedNonLocal = "variable-not-defined-nonlocal";
+        public const string NewTypeArguments = "newtype-arguments";
     }
 }
--- a/src/Analysis/Ast/Impl/Resources.resx
+++ b/src/Analysis/Ast/Impl/Resources.resx
@@ -190,6 +190,6 @@
     <value>Unable to determine analysis cache path. Exception: {0}. Using default '{1}'.</value>
   </data>
   <data name="NewTypeFirstArgNotString" xml:space="preserve">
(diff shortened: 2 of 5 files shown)


CTrando:
Alright, all good to review!

src/Analysis/Ast/Impl/Resources.resx
```diff
@@ -189,4 +189,7 @@
   <data name="UnableToDetermineCachePathException" xml:space="preserve">
     <value>Unable to determine analysis cache path. Exception: {0}. Using default '{1}'.</value>
   </data>
+  <data name="NewTypeFirstArgNotString" xml:space="preserve">
+    <value>The first argument to NewType must be a string but is of type '{0}'.</value>
```
jakebailey: A comma after `be a string` would make it read better.

src/Analysis/Ast/Test/LintNewTypeTests.cs
```diff
@@ -0,0 +1,128 @@
...
+        }
+
+        [DataRow("'test'", "float")]
+        [DataRow("'testing'", "int")]
+        [DataTestMethod, Priority(0)]
+        public async Task NoDiagnosticOnStringFirstArg(string name, string type) {
+            string code = $@"
+from typing import NewType
+
+T = NewType({name}, {type})
```
jakebailey: I'd probably move the quotes down here instead of having them in the DataRows.

src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
```diff
@@ -25,5 +25,6 @@ public static class ErrorCodes {
         public const string UndefinedVariable = "undefined-variable";
         public const string VariableNotDefinedGlobally= "variable-not-defined-globally";
         public const string VariableNotDefinedNonLocal = "variable-not-defined-nonlocal";
+        public const string NewTypeArguments = "newtype-arguments";
```
jakebailey: I think we should prefix all of these, so do `typing-newtype-arguments` instead (and rename the constant to match).

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
```diff
@@ -217,13 +218,25 @@ private IPythonType CreateTupleType(IReadOnlyList<IPythonType> typeArgs)
...
-                // TODO: report incorrect first argument to NewVar
+
+                var firstArgType = (typeArgs[0] as PythonInstance)?.Type.Name;
+                var eval = args.Eval;
+                var expression = args.Expression;
+
+                eval.ReportDiagnostics(
+                    eval.Module?.Uri,
+                    new DiagnosticsEntry(Resources.NewTypeFirstArgNotString.FormatInvariant(firstArgType), 
+                        expression.GetLocation(eval.Module).Span,
```
CTrando: check for null here

src/Analysis/Ast/Test/LintNewTypeTests.cs
```diff
@@ -0,0 +1,128 @@
...
+
+T = NewType(5, int)
+
+";
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Diagnostics.Should().HaveCount(1);
+
+            var diagnostic = analysis.Diagnostics.ElementAt(0);
+            diagnostic.ErrorCode.Should().Be(Diagnostics.ErrorCodes.TypingNewTypeArguments);
+            diagnostic.Message.Should().Be(Resources.NewTypeFirstArgNotString.FormatInvariant("int"));
```
jakebailey: Spans here too (maybe skip the datarow'd ones for simplicity).

PR #1261, opened 2019-06-27
When returning in init method, report a diagnostic error

CTrando:


src/Analysis/Ast/Test/LintReturnInInitTests.cs
```diff
@@ -0,0 +1,76 @@
...
+";
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Diagnostics.Should().HaveCount(1);
+
+            var diagnostic = analysis.Diagnostics.ElementAt(0);
+            diagnostic.Severity.Should().Be(Severity.Warning);
+            diagnostic.ErrorCode.Should().Be(Diagnostics.ErrorCodes.ReturnInInit);
+            diagnostic.Message.Should().Be(Resources.ReturnInInit);
+        }
+    }
```
jakebailey: Can you add a test that looks like:

```python
class A:
    def __init__(self, x):
        self.x = x
        if x > 0:
            return
        self.x += 1
```

Or similar, to verify that bare returns/`None` returns are still valid?  I believe returns for control flow are alright, and the only thing that needs to hold true is that `__init__` returns `None`.

CTrando: Good catch. Added.

src/Analysis/Ast/Test/LintReturnInInitTests.cs
```diff
@@ -0,0 +1,124 @@
...
+class A:
+    def __init__(self, x):
+        self.x = x
+        self.x += 1
+        return None
+
+a = A(1)
+";
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Diagnostics.Should().HaveCount(0);
```
jakebailey: `BeEmpty()`

src/Analysis/Ast/Test/LintReturnInInitTests.cs
```diff
@@ -0,0 +1,124 @@
...
+        public async Task EmptyReturnInInit() {
+            const string code = @"
+class A:
+    def __init__(self, x):
+        self.x = x
+        return 
+a = A(1)
+";
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Diagnostics.Should().HaveCount(0);
```
jakebailey: `BeEmpty()`

src/Analysis/Ast/Test/LintReturnInInitTests.cs
```diff
@@ -0,0 +1,124 @@
...
+        self.area = width * height
+        return self.area
+
+r = Rectangle(10, 10)
+";
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Diagnostics.Should().HaveCount(1);
+
+            var diagnostic = analysis.Diagnostics.ElementAt(0);
+            diagnostic.Severity.Should().Be(Severity.Warning);
```
jakebailey: Can you also test the span of the diagnostics similarly to the other PR? (And do this going forward, so we're testing that the diagnostics show up in the right place.)

CTrando: Alright, sure I can do that.

PR #1263, opened 2019-06-28
Improving how arg set handles specialized function parameters

CTrando:
Making specialized functions more granular in the way that we can specify their parameters and have them handled just like normal functions, fixing test cases related to that, fixing documentation for TypeVar related to keyword only arguments.

I'll need this PR in first before #1230.


src/Analysis/Ast/Impl/Types/ArgumentSet.cs
```diff
@@ -113,10 +113,13 @@ internal sealed class ArgumentSet : IArgumentSet {
...
             var fd = overload.FunctionDefinition;
 
-            if (fn.IsSpecialized) {
-                // Typically specialized function, like TypeVar() that does not actually have AST definition.
-                // Make the arguments from the call expression. If argument does not have name, 
-                // try using name from the function definition based on the argument position.
+            // Some specialized functions have more complicated definitions, so we pass
+            // parameters to those, TypeVar() is an example, so we allow the latter logic to handle
+            // argument instatiation. For simple specialized functions, it is enough to handle here.
+            if (fn.IsSpecialized && !overload.HasParameters()) {
```
jakebailey: `overload.Parameters.Count == 0` and remove the function.

src/Analysis/Ast/Impl/Types/PythonFunctionOverload.cs
```diff
@@ -17,6 +17,7 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading.Tasks;
```
jakebailey: Remove

src/Analysis/Ast/Impl/Extensions/AnalysisExtensions.cs
```diff
@@ -44,9 +44,7 @@ public static IScope FindScope(this IDocumentAnalysis analysis, SourceLocation l
             var f = analysis.GetOrCreateFunction(name);
             if (f != null) {
                 foreach (var o in f.Overloads.OfType<PythonFunctionOverload>()) {
-                    if(parameters != null) {
-                        o.SetParameters(parameters);
-                    }
+                    o.SetParameters(parameters ?? new List<ParameterInfo>());
```
jakebailey: We typically use something like `Array.Empty` or `Enumerable.Empty` instead of creating an empty list logic.

Also, setting an empty list compared to not setting it at all seems like a difference; especially if we move to the idea we talked about (determining if the parameters were set at all, compared to actually empty).

CTrando: I felt that having the null check there was uglier, but if the plan is to separate the idea of having no parameters or having none set, then I'll switch it out.

src/Analysis/Ast/Impl/Modules/BuiltinsPythonModule.cs
```diff
@@ -158,7 +159,17 @@ protected override string[] GetScrapeArguments(IPythonInterpreter interpreter)
             Analysis.SpecializeFunction("max", BuiltinsSpecializations.Identity);
             Analysis.SpecializeFunction("min", BuiltinsSpecializations.Identity);
             Analysis.SpecializeFunction("next", BuiltinsSpecializations.Next);
-            Analysis.SpecializeFunction("open", BuiltinsSpecializations.Open, new[] { "io" });
+
+            Analysis.SpecializeFunction("open", BuiltinsSpecializations.Open, new[] {
```
AlexanderSher: Is it the same set of parameters for all supported python versions?

jakebailey: Nope, Python 3 has all of these, but Python 2 only has `open(name[, mode[, buffering]])`. Probably needs to be conditionally specialized.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
```diff
@@ -35,31 +35,48 @@ public GenericTypeParameter(string name, IPythonModule declaringModule, IReadOnl
...
+            var keyWordArgs = new List<string>();
+
+            foreach (var arg in args.Skip(1)) {
+                if (arg.ValueIsDefault)
+                    continue;
+
+                keyWordArgs.Add($"{arg.Name}={(arg.Value as IPythonConstant)?.Value}");
+            }
+
+            var docArgs = new[] { $"'{name}'" }.Concat(constraints.Select(c => c.IsUnknown() ? "?" : c.Name)).Concat(keyWordArgs);
```
AlexanderSher: `constraints.Select(c => c.IsUnknown() ? "?" : c.Name).Concat(keyWordArgs).Prepend($"'{name}'")`

AlexanderSher: Also, `name` can be null. Do we care about it?

CTrando: Thanks, that syntax looks much nicer. Name cannot be null when called in that function because I checked for it in FromTypeVar, I just forgot to remove the null check.

src/Analysis/Ast/Impl/Modules/BuiltinsPythonModule.cs
```diff
@@ -181,5 +173,25 @@ protected override string[] GetScrapeArguments(IPythonInterpreter interpreter)
...
+                    new ParameterInfo("buffering", Interpreter.GetBuiltinType(BuiltinTypeId.Int), ParameterKind.Normal, new PythonConstant(-1, Interpreter.GetBuiltinType(BuiltinTypeId.Int))),
+                    new ParameterInfo("encoding", Interpreter.GetBuiltinType(BuiltinTypeId.Str), ParameterKind.Normal, new PythonConstant(null, Interpreter.GetBuiltinType(BuiltinTypeId.NoneType))),
+                    new ParameterInfo("errors", Interpreter.GetBuiltinType(BuiltinTypeId.Str), ParameterKind.Normal, new PythonConstant(null, Interpreter.GetBuiltinType(BuiltinTypeId.NoneType))),
+                    new ParameterInfo("newline", Interpreter.GetBuiltinType(BuiltinTypeId.Str), ParameterKind.Normal, new PythonConstant(null, Interpreter.GetBuiltinType(BuiltinTypeId.NoneType))),
+                    new ParameterInfo("closefd", Interpreter.GetBuiltinType(BuiltinTypeId.Bool), ParameterKind.Normal, new PythonConstant(true, Interpreter.GetBuiltinType(BuiltinTypeId.Bool))),
+                    new ParameterInfo("opener", Interpreter.GetBuiltinType(BuiltinTypeId.Str), ParameterKind.Normal, new PythonConstant(null, Interpreter.GetBuiltinType(BuiltinTypeId.Str)))};
```
jakebailey: Nit: the `};` being on this line is inconsistent and should probably be on its own line below.

src/Analysis/Ast/Impl/Extensions/AnalysisExtensions.cs
```diff
@@ -39,10 +40,13 @@ public static IScope FindScope(this IDocumentAnalysis analysis, SourceLocation l
         /// <summary>
         /// Provides ability to dynamically calculate function return type.
         /// </summary>
-        public static void SpecializeFunction(this IDocumentAnalysis analysis, string name, ReturnValueProvider returnTypeCallback, string[] dependencies = null) {
+        internal static void SpecializeFunction(this IDocumentAnalysis analysis, string name, ReturnValueProvider returnTypeCallback, IReadOnlyList<ParameterInfo> parameters = null, string[] dependencies = null) {
             var f = analysis.GetOrCreateFunction(name);
             if (f != null) {
                 foreach (var o in f.Overloads.OfType<PythonFunctionOverload>()) {
+                    if(parameters != null) {
```
jakebailey: Nit: formatting

src/Analysis/Ast/Impl/Analyzer/Handlers/WithHandler.cs
```diff
@@ -35,7 +35,7 @@ internal sealed class WithHandler : StatementHandler {
                     // If fetching context from __enter__ failed, annotation in the stub may be using
                     // type from typing that we haven't specialized yet or there may be an issue in
                     // the stub itself, such as type or incorrect type. Try using context manager then.
-                    context = context ?? contextManager;
+                    context = context.IsUnknown() ? contextManager: context;
```
jakebailey: Nit: formatting (`a ? b : c`)

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -135,8 +135,11 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
                 def = args[2];
             }
 
+            // second argument to getattr was not a string, which is a runtime error
+            // getattr(a, 3.14)
             if (name == null) {
-                return def;
+                // TODO diagnostic error when second arg of getattr is not a string
+                return  module.Interpreter.UnknownType;
```
jakebailey: Nit: formatting (extra space)

I'm not sure I like this change; I know that if we believe the name argument wasn't a string then it's wrong, but returning unknown as a whole feels worse than inferring at least _something_ via the default.

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
```diff
@@ -49,6 +50,14 @@ internal sealed class TypingModule : SpecializedModule {
             // TypeVar
             var fn = PythonFunctionType.Specialize("TypeVar", this, GetMemberDocumentation("TypeVar"));
             var o = new PythonFunctionOverload(fn.Name, location);
+            o.SetParameters(new List<ParameterInfo> {
+                    new ParameterInfo("name", Interpreter.GetBuiltinType(BuiltinTypeId.Str), ParameterKind.Normal, null),
+                    new ParameterInfo("*constraints", Interpreter.GetBuiltinType(BuiltinTypeId.Str), ParameterKind.List, null),
```
jakebailey: Should the names include the `*`, or should that be dependent on `ParameterKind.List`? I'd write a function with `*args` and see what it's represented as first.

CTrando: Yeah the ```*``` shouldn't be there, I just tested it.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
CTrando: So the reason I made it Unknown was because a test checked for it to be UnknownType when I was returning NoneType. Would it preferable to return NoneType? 

Doing ```getattr()``` is a runtime error and our convention has been to return UnknownTypes for those.

jakebailey: What test is failing? Presumably, all tests passed before this PR, so the code I had written would already have been passing.

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
CTrando: You are right, I wrote a test and it should be constraints.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
jakebailey: The test actually was already saying that it should be unknown (ClassesTest.GetAttr). Resolving.

CTrando:
To test that the change works correctly, I've tested:

* Open function works fine after specifying its parameters

* TypeVar function works fine after specifying its parameters - can look into argument set after calling TypeVar and see all of its variables correctly assigned.

* Test with a function that does not have its arguments specified and verify that all its arguments became positional parameters.

src/Analysis/Ast/Test/ValuesTests.cs
```diff
@@ -189,7 +193,7 @@ class Y(object):
                 .And.HaveVariable("x").OfType("TextIOWrapper")
                 .And.HaveVariable("xb").OfType("BufferedWriter");
         }
-
+
```
jakebailey: Nit: weird whitespace added.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
```diff
@@ -35,31 +35,48 @@ public GenericTypeParameter(string name, IPythonModule declaringModule, IReadOnl
...
+            var documentation = GetDocumentation(args, constraints);
             return new GenericTypeParameter(name, declaringModule, constraints, documentation, location);
         }
+
+        private static string GetDocumentation(IReadOnlyList<IArgument> args, IReadOnlyList<IPythonType> constraints) {
+            var name = (args[0].Value as IPythonConstant).GetString();
+            var keyWordArgs = new List<string>();
+
+            foreach (var arg in args.Skip(1)) {
+                if (arg.ValueIsDefault)
```
jakebailey: Nit: we always use braces, even for single statement blocks.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
```diff
@@ -35,31 +35,48 @@ public GenericTypeParameter(string name, IPythonModule declaringModule, IReadOnl
...
 
-            var name = (args[0] as IPythonConstant)?.GetString();
+            var name = (args[0].Value as IPythonConstant)?.GetString();
             if (string.IsNullOrEmpty(name)) {
                 // TODO: report that type name is not a string.
                 return declaringModule.Interpreter.UnknownType;
             }
 
-            var constraints = args.Skip(1).Select(a => {
+            var constraints = constraintArgs?.Select(a => {
```
MikhailArkhipov: Not sure about this. `x?.Select` will yield null and `.ToArray` may then crash. Also, `constraints` is assumed to never be null below.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
```diff
@@ -35,31 +35,48 @@ public GenericTypeParameter(string name, IPythonModule declaringModule, IReadOnl
 
 
         public static IPythonType FromTypeVar(IArgumentSet argSet, IPythonModule declaringModule, IndexSpan location = default) {
-            var args = argSet.Values<IMember>();
+            var args = argSet.Arguments;
+            var constraintArgs = argSet.ListArgument?.Values;
```
MikhailArkhipov: `argSet.ListArgument?.Values ?? Array.Empty<>()

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
MikhailArkhipov: `var arg in args.Skip(1).Where(x => !x.ValueIsDefault)` will avoid comparison inside loop

MikhailArkhipov: In fact, `keyWordArgs.Add(args.Skip(1).Where(x => !x.ValueIsDefault).Select(x => $"{x.Name}={(x.Value as IPythonConstant)?.Value}")`

src/Analysis/Ast/Impl/Types/Definitions/IArgumentSet.cs
```diff
@@ -43,11 +43,16 @@ public interface IArgument {
...
         IPythonType Type { get; }
-        
+
         /// <summary>
         /// Parameter location in the AST.
         /// </summary>
         Node Location { get; }
+
+        /// <summary>
+        /// Returns true if this is a default argument.
```
MikhailArkhipov: `Value of the argument is default` I guess?

src/Analysis/Ast/Test/ArgumentSetTests.cs
```diff
@@ -166,6 +166,77 @@ public void TestInitialize()
...
+
+        [TestMethod, Priority(0)]
+        public async Task NewTypeSpecializedArgs() {
+            const string code = @"
+from typing import NewType
+
+NewType('T', int)
+";
+            var argSet = await GetArgSetAsync(code, funcName: "NewType");
+            argSet.Arguments.Count.Should().Be(2);
```
MikhailArkhipov: `Should().HaveCount(2)`

src/Analysis/Ast/Test/TypingTests.cs
```diff
@@ -64,11 +64,57 @@ import io
...
+        public async Task TypeVarCovariantDocCheck() {
+            const string code = @"
+from typing import TypeVar
+
+T = TypeVar('T', str, int, covariant=True)
+";
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Should().HaveVariable("T")
+                .Which.Value.Should().HaveDocumentation("TypeVar('T', str, int, covariant=True)");
+            analysis.Should().HaveVariable("T").OfType(typeof(IGenericTypeDefinition));
```
MikhailArkhipov: `analysis.Should().HaveVariable("T").Which.Should().BeAssignableTo<IGenericTypeDefinition>();` or `analysis.Should().HaveVariable("T").OfType<IGenericTypeDefinition>()`

MikhailArkhipov: Or `analysis.Should().HaveVariable("T").Which.IsGeneric().Should().BeTrue()` but it is a bit broader check.

MikhailArkhipov: I'd actually write fluent extension so we could write `analysis.Should().HaveGenericVariable("T")`. @AlexanderSher?

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
CTrando: that's sort of hard to read but I'll make the change.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
CTrando: ok

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
CTrando: ok

src/Analysis/Ast/Impl/Types/Definitions/IArgumentSet.cs
CTrando: ok

src/Analysis/Ast/Test/ArgumentSetTests.cs
CTrando: ok

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
MikhailArkhipov: We have lots of those :-) not too hard with some Linq experience. You can indent it as we do in other places.
```cs
keyWordArgs.Add(
    args
        .Skip(1)
        .Where(x => !x.ValueIsDefault)
        .Select(x => $"{x.Name}={(x.Value as IPythonConstant)?.Value}"
);
```

src/Analysis/Ast/Test/TypingTests.cs
CTrando: I'll try to do the fluent assertions

PR #1266, opened 2019-07-01
Switches indexing from using an object parameter to an argument set parameter

CTrando:


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
```diff
@@ -49,7 +49,7 @@ internal sealed partial class ExpressionEval {
                 }
                 var index = GetValueFromExpression(expr.Index);
                 if (index != null) {
-                    return type.Index(instance, index);
+                    return type.Index(instance, new ArgumentSet(new List<IMember>(){index}, expr, this));
```
jakebailey: `new[] { index }`

Generally, we'll make arrays like this (with type inference) rather than making Lists.

src/Analysis/Ast/Impl/Specializations/Typing/Values/TypingTuple.cs
```diff
@@ -17,6 +17,7 @@
 using Microsoft.Python.Analysis.Types;
 using Microsoft.Python.Analysis.Values;
 using Microsoft.Python.Analysis.Values.Collections;
+using System.Collections.Generic;
```
jakebailey: Make sure your usings are getting sorted; System usings should automatically be getting put above any others.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
CTrando: Got it.

src/Analysis/Ast/Impl/Specializations/Typing/Values/TypingTuple.cs
CTrando: Ok!

jakebailey:
I think I'd like to wait for feedback from @MikhailArkhipov on this, but I definitely like this change.

PR #1272, opened 2019-07-02
Check if #noqa is present in line before adding diagnostic

CTrando:


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.cs
```diff
@@ -16,6 +16,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Reflection;
```
jakebailey: Needed?

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsEntry.cs
```diff
@@ -51,6 +53,20 @@ public sealed class DiagnosticsEntry {
...
         public DiagnosticSource Source { get; }
 
+        public bool ShouldReport(IPythonModule module) {
+            // module should always be a user written python module
+            if (module.ModuleType != ModuleType.User || !(module is PythonModule pythonModule)) {
+                return false;
+            }
+
+            string line = pythonModule.GetLine(SourceSpan.Start.Line);
+            if (line.Contains("#noqa")) {
```
jakebailey: This is too simple; `# noqa` is also valid, and `# noqa: E123` should only disable whatever's associated with that code.

src/Analysis/Ast/Test/LintNoQATests.cs
```diff
@@ -0,0 +1,97 @@
...
+        [DataTestMethod, Priority(0)]
+        public async Task IgnoreMissingParameter(string code) {
+            var d = await LintAsync(code);
+            d.Should().HaveCount(0);
+        }
+
+        [TestMethod, Priority(0)]
+        public async Task IgnoreMissingImport() {
+            string code = @"
+from django.contrib.auth.models import User         #noqa
```
jakebailey: Make up a fake module name instead of using django; if these tests are run in an interpreter where django is installed, it could result in a false success (if the `noqa` code were to break in a future change).

src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -243,6 +243,24 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
...
+        /// Returns the string line corresponding to the given location
+        /// </summary>
+        /// <param name="line">The line number</param>
+        public string GetLine(int line) {
+            StringBuilder sb = new StringBuilder();
+            SourceLocation source = new SourceLocation(line, 1);
+            string content = Content;
+            var index = this.GetAst().LocationToIndex(source);
+
+            if (!content.IsNullOrEmpty()) {
```
jakebailey: This is probably not the most efficient way to do this; it's likely easier to loop and scan for the end of line index and then slice out the part you want with Substring.

src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -243,6 +243,24 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
             }
         }
 
+        /// <summary>
+        /// Returns the string line corresponding to the given location
+        /// </summary>
+        /// <param name="line">The line number</param>
+        public string GetLine(int line) {
```
jakebailey: I'd be more inclined to make this an extension method in PythonModuleExtensions, rather than making it actually on the type.

src/Analysis/Ast/Impl/Modules/PythonModule.cs
CTrando: Changed.

src/Analysis/Ast/Impl/Modules/PythonModule.cs
CTrando: Changed.

src/Analysis/Ast/Impl/Extensions/PythonModuleExtensions.cs
```diff
@@ -26,5 +28,23 @@ internal static PythonAst GetAst(this IPythonModule module)
...
+        /// Returns the string line corresponding to the given location
+        /// </summary>
+        /// <param name="line">The line number</param>
+        internal static string GetLine(this PythonModule module, int line) {
+            SourceLocation source = new SourceLocation(line, 1);
+            string content = module.Content;
+            var start = GetAst(module).LocationToIndex(source);
+            var length = 0;
+
+            if (!content.IsNullOrEmpty()) {
```
jakebailey: Two things:
- Use `string.IsNullOrEmpty(content)` instead; the extension method is on `IEnumerable` and is likely less efficient.
- I'd recommend writing `if (string.IsNullOrEmpty(content)) { return content }` instead, to avoid calling Substring on an empty (or worse, null) string.

src/Analysis/Ast/Impl/Extensions/PythonModuleExtensions.cs
```diff
@@ -26,5 +28,23 @@ internal static PythonAst GetAst(this IPythonModule module)
...
+                while(start + length < content.Length && content[start + length] != '\n' && content[start + length] != '\r') {
```
jakebailey: I'd take a look at GetLineIndent; the loop style used there is cleaner than having all of these `start + length` all over the place.

src/Analysis/Ast/Impl/Extensions/PythonModuleExtensions.cs
CTrando: Good points, completely forgot about the latter.

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsEntry.cs
```diff
@@ -51,6 +53,26 @@ public sealed class DiagnosticsEntry {
         /// </summary>
         public DiagnosticSource Source { get; }
 
+        public bool ShouldReport(IPythonModule module) {
+            // module should always be a user written python module
+            if (module.ModuleType != ModuleType.User || !(module is PythonModule pythonModule)) {
```
AlexanderSher: `!(module is PythonModule pythonModule)` shouldn't be needed. Do you have a case when `module.ModuleType == ModuleType.User` but `module` isn't a `PythonModule`?

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsEntry.cs
```diff
@@ -51,6 +53,26 @@ public sealed class DiagnosticsEntry {
...
+                return false;
+            }
+
+            string line = pythonModule.GetLine(SourceSpan.Start.Line);
+
+            // find if #noqa is in the string
+            int commentPos = line.IndexOf("#");
+            if (commentPos >= 0) {
```
AlexanderSher: `return commentPos >= 0 && line.IndexOfOrdinal("#noqa", commentPos) == commentPos`

src/Analysis/Ast/Impl/Extensions/PythonModuleExtensions.cs
```diff
@@ -26,5 +28,23 @@ internal static PythonAst GetAst(this IPythonModule module)
...
             => ((IAstNodeContainer)module).AddAstNode(o, n);
+
+        /// <summary>
+        /// Returns the string line corresponding to the given location
+        /// </summary>
+        /// <param name="line">The line number</param>
+        internal static string GetLine(this PythonModule module, int line) {
+            SourceLocation source = new SourceLocation(line, 1);
+            string content = module.Content;
+            var start = GetAst(module).LocationToIndex(source);
```
AlexanderSher: If you're getting AST, you can use `PythonAst.CommentLocations` and do check similar to `AstExtensions.IsInsideComment`.

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsEntry.cs
```diff
@@ -51,6 +53,26 @@ public sealed class DiagnosticsEntry {
         /// </summary>
         public DiagnosticSource Source { get; }
 
+        public bool ShouldReport(IPythonModule module) {
+            // module should always be a user written python module
+            if (module.ModuleType != ModuleType.User || !(module is PythonModule pythonModule)) {
+                return false;
+            }
+
+            string line = pythonModule.GetLine(SourceSpan.Start.Line);
```
AlexanderSher: You don't need to get a string line here. You can get the location of the comment using method similar to `AstExtensions.IsInsideComment`, something like `AstExtensions.HasComment(this PythonAst ast, string comment, int lineNumber)`.

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsEntry.cs
CTrando: I need to make the conversion between IPythonModule to PythonModule and I didn't feel comfortable just doing it without a matching check.

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsEntry.cs
CTrando: I think I see what you are saying, I've made some changes, can you tell me if I'm right?

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsEntry.cs
```diff
@@ -51,6 +53,19 @@ public sealed class DiagnosticsEntry {
         /// </summary>
         public DiagnosticSource Source { get; }
 
+        public bool ShouldReport(IPythonModule module) {
+            // module should always be a user written python module
+            if (module.ModuleType != ModuleType.User || !(module is PythonModule pythonModule)) {
+                return false;
+            }
+
+            if(!module.GetAst().HasComment(pythonModule, "noqa", SourceSpan.Start.Line)) {
```
jakebailey: Nit: formatting

src/Analysis/Ast/Impl/Extensions/AstExtensions.cs
```diff
@@ -60,9 +61,23 @@ public static Expression FindExpression(this PythonAst ast, SourceLocation locat
...
+
+            // line is not a comment
+            if (match < 0 || match >= ast.CommentLocations.Length) {
+                return false;
+            }
+
+            var commentLoc = ast.CommentLocations[match];
+            string line = module.GetLine(commentLoc.Line);
+
+            return line.Contains(comment);
```
jakebailey: I'm thinking this is too loose; if you're just doing `line.Contains("noqa")`, that's going to remove too much. Before I had said that you should really only be checking if a comment is exactly `"noqa"` (case insentitive), ignoring whitespace.

CTrando: You're right, I forgot about that again!

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsEntry.cs
```diff
@@ -51,6 +53,19 @@ public sealed class DiagnosticsEntry {
         /// </summary>
         public DiagnosticSource Source { get; }
 
+        public bool ShouldReport(IPythonModule module) {
+            // module should always be a user written python module
+            if (module.ModuleType != ModuleType.User || !(module is PythonModule pythonModule)) {
+                return false;
+            }
+
+            if (module.GetAst().HasComment(pythonModule, "noqa", SourceSpan.Start.Line)) {
```
MikhailArkhipov: `module.Analysis.Ast`. No need to cast to `PythonModule`

src/Analysis/Ast/Impl/Extensions/PythonModuleExtensions.cs
```diff
@@ -26,5 +28,23 @@ internal static PythonAst GetAst(this IPythonModule module)
             => ((IAstNodeContainer)module).GetAstNode<T>(o);
         internal static void AddAstNode(this IPythonModule module, object o, Node n)
             => ((IAstNodeContainer)module).AddAstNode(o, n);
+
+        /// <summary>
+        /// Returns the string line corresponding to the given location
+        /// </summary>
+        /// <param name="line">The line number</param>
+        internal static string GetLine(this PythonModule module, int line) {
```
MikhailArkhipov: During analysis AST attached to the module analysis is old. `module.Analysis` is actually previous. You need to explicitly use AST that is in the current evaluator.

src/Analysis/Ast/Impl/Extensions/AstExtensions.cs
```diff
@@ -60,9 +61,28 @@ public static Expression FindExpression(this PythonAst ast, SourceLocation locat
             return ast.CommentLocations[match].Column < location.Column;
         }
 
+        internal static bool HasComment(this PythonAst ast, PythonModule module, string comment, int lineNum) {
```
MikhailArkhipov: This does not seem right. If you are passing the AST, then it should be using to get locations and not `module.GetLine` since that appears to be using AST attached to the module which is most probably stale.

src/Analysis/Ast/Impl/Extensions/PythonModuleExtensions.cs
CTrando: I have observed doing GetAst(module) gets the most current AST.

src/Analysis/Ast/Impl/Extensions/AstExtensions.cs
CTrando: I went ahead and passed the AST to module.GetLine()

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsEntry.cs
CTrando: Removed the cast.

CTrando: I need the cast for PythonModule.GetLine. It doesn't make sense to put that in IPythonModule because some PythonModules are not user code so getting a line from them is impossible.

MikhailArkhipov: You can simply return an empty line for all others.

CTrando: I've gone ahead and put them in IPythonModule

src/Analysis/Ast/Impl/Extensions/AstExtensions.cs
```diff
@@ -60,9 +62,20 @@ public static Expression FindExpression(this PythonAst ast, SourceLocation locat
             return ast.CommentLocations[match].Column < location.Column;
         }
 
+        internal static bool HasComment(this PythonAst ast, IPythonModule module, string comment, int lineNum) {
```
MikhailArkhipov: This is unlikely to be used anywhere except linter since it is not particularly common to look for a specific comment. So this is basically a function somewhere inside linter reporting. `GetLineComment` might be useful though.

CTrando: I misinterpreted what you said above.

src/Analysis/Ast/Impl/Extensions/AstExtensions.cs
```diff
@@ -60,9 +62,20 @@ public static Expression FindExpression(this PythonAst ast, SourceLocation locat
...
+            string line = module.GetLine(ast, lineNum);
+
+            int commentPos = line.IndexOf('#');
+            if (commentPos < 0) {
+                return false; 
+            }
+
+            return line.Substring(commentPos + 1).Contains(comment);
```
jakebailey: This function is _still_ doing `Contains`. For `"noqa"`, we need to check the exact string.

I think this should be rewritten to be a function that returns the comment for a line (or `null`), then in `ReportDiagnostics`, you can do `ast.GetComment(SourceSpan.Start.Line)?.Trim().EqualsIgnoreCase("noqa")`.

MikhailArkhipov: I think it is rarely used case to be an extension. I'd rather have `GetCommentAt` on module which has AST rather than extension to AST that suddenly needs module. AST is a lower layer entity and extensions to it should not require higher level definitions. Consider that AST extensions can actually be placed in Python.Parsing assembly which does not have access to the module.

src/Analysis/Ast/Impl/Extensions/AstExtensions.cs
```diff
@@ -17,16 +17,18 @@
 using System.Diagnostics;
 using System.Linq;
 using Microsoft.Python.Analysis.Analyzer.Expressions;
+using Microsoft.Python.Analysis.Modules;
```
MikhailArkhipov: This is suspicious. AST extensions should not be needing analysis. We may actually move most of extensions to `Python.Parsing`

src/Analysis/Ast/Impl/Extensions/PythonModuleExtensions.cs
```diff
@@ -26,5 +28,24 @@ internal static PythonAst GetAst(this IPythonModule module)
             => ((IAstNodeContainer)module).GetAstNode<T>(o);
         internal static void AddAstNode(this IPythonModule module, object o, Node n)
             => ((IAstNodeContainer)module).AddAstNode(o, n);
+
+        /// <summary>
+        /// Returns the string line corresponding to the given location
+        /// </summary>
+        /// <param name="line">The line number</param>
+        internal static string GetLine(this IPythonModule module, PythonAst ast, int line) {
```
MikhailArkhipov: Can as well implement `GetComment` here as well.

src/Analysis/Ast/Impl/Extensions/AstExtensions.cs
jakebailey: The trouble here will be getting a matching pairing of AST and file content. It's a shame that we need the AST in order to convert a line number to an offset, only to then go back to the content to split lines.

MikhailArkhipov: Yeah, but it is called on `module.GetAst()` anyway

src/Analysis/Ast/Impl/Extensions/AstExtensions.cs
MikhailArkhipov: It is called on `module.GetAst()` anyway so module is sufficient

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsEntry.cs
```diff
@@ -51,6 +54,20 @@ public sealed class DiagnosticsEntry {
...
         public DiagnosticSource Source { get; }
 
+        public bool ShouldReport(IPythonModule module) {
+            // Only report for user written modules
+            if (module.ModuleType != ModuleType.User) {
+                return false;
+            }
+
+            // If user specifies #noqa, then do not report diagnostic
+            if (module.GetComment(SourceSpan.Start.Line).EqualsIgnoreCase("noqa")) {
```
jakebailey: This looks way better and cleaner.

src/Analysis/Ast/Test/LintNoQATests.cs
```diff
@@ -0,0 +1,103 @@
...
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Diagnostics.Should().BeEmpty();
+        }
+
+        private async Task<IReadOnlyList<DiagnosticsEntry>> LintAsync(string code, InterpreterConfiguration configuration = null) {
+            var analysis = await GetAnalysisAsync(code, configuration ?? PythonVersions.LatestAvailable3X);
+            var a = Services.GetService<IPythonAnalyzer>();
+            return a.LintModule(analysis.Document);
+        }
+    }
```
jakebailey: Can you add a test which does `# NOQA` (caps)? Just want to make sure we don't regress on that, since there are users who do all caps.

CTrando: Sure.

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsEntry.cs
CTrando: Reminder to myself to add a null check here.

src/Analysis/Ast/Impl/Extensions/PythonModuleExtensions.cs
```diff
@@ -26,5 +27,39 @@ internal static PythonAst GetAst(this IPythonModule module)
...
+        /// <summary>
+        /// Returns the comment corresponding to the given line or null if there is no comment
+        /// </summary>
+        /// <param name="lineNum">The line number</param>
+        internal static string GetComment(this IPythonModule module, int lineNum) {
+            string line = module.GetLine(lineNum);
+
+            int commentPos = line.IndexOf('#');
+            if (commentPos < 0) {
+                return null;
```
MikhailArkhipov: return `string.Empty` for safety

src/Analysis/Ast/Impl/Diagnostics/DiagnosticsEntry.cs
MikhailArkhipov: Don't have to if you return `string.Empty`.

MikhailArkhipov:
Add test for "NOQA" and return `string.Empty`. Otherwise LGTM.

src/Analysis/Ast/Impl/Extensions/PythonModuleExtensions.cs
CTrando: Was initially worried about distinguishing between the case where no comment exists and a comment exists but it has nothing after #. Do you think this is worth distinguishing?

jakebailey: Probably not, honestly. If we need to check that something is missing a comment altogether, then we can modify the function then, but it's much more likely that we're going to be checking for what's in a comment rather than if one doesn't exist.

CTrando: Sure then, I'll change this.

PR #1273, opened 2019-07-02
On redefinition of function, give a diagnostic message

CTrando:
Note, we skipped this one before because we were worried about collisions with imports, but the function definition code seems to only looks up functions in the local context - so it will only care about other functions declared in the same file as it.


jakebailey:
I think there are a few more test cases you're going to want to write, things like:

```python
def foo():
    def foo():
        return 123
    return foo()

foo()
```

Nested classes, etc.

CTrando:
Aha, thanks, I was wondering what kind of cases to test beyond this.

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
```diff
@@ -104,12 +105,16 @@ internal sealed class SymbolCollector : PythonWalker {
         }
 
         private void AddFunction(FunctionDefinition fd, IPythonType declaringType) {
-            if (!(_eval.LookupNameInScopes(fd.Name, LookupOptions.Local) is PythonFunctionType existing)) {
+            var existing = _eval.LookupNameInScopes(fd.Name, LookupOptions.Local) as PythonFunctionType;
+            if (existing is null) {
```
AlexanderSher: It is better to preserve type check instead of introducing `null` check:
```cs
if (_eval.LookupNameInScopes(fd.Name, LookupOptions.Local) is PythonFunctionType existing) {
    ReportRedefinedFunction(fd, existing);
} else {
    existing = new PythonFunctionType(fd, declaringType, _eval.GetLocationOfName(fd));
    // The variable is transient (non-user declared) hence it does not have location.
    // Function type is tracking locations for references and renaming.
    _eval.DeclareVariable(fd.Name, existing, VariableSource.Declaration);
}
```

CTrando: Yes this is much nicer syntax, thanks!

src/Analysis/Ast/Test/LintRedefinedFunctionTests.cs
```diff
@@ -0,0 +1,186 @@
...
+        public TestContext TestContext { get; set; }
+
+        [TestInitialize]
+        public void TestInitialize()
+            => TestEnvironmentImpl.TestInitialize($"{TestContext.FullyQualifiedTestClassName}.{TestContext.TestName}");
+
+        [TestCleanup]
+        public void Cleanup() => TestEnvironmentImpl.TestCleanup();
+
+        [TestMethod, Priority(0)]
```
jakebailey: Can you also test the behavior where one module contains:

```python
def hello():
    pass

def hello2():
    pass
```

And another imports it, like:

```python
from module1 import *

def hello():
    pass
```

Or:

```python
from module1 import hello

def hello():
    pass
```

Or:

```python
from module1 import hello2 as hello

def hello():
    pass
```

I don't think any of these should get diagnostics, but it should probably be tested.

CTrando: Sure!

jakebailey: You resolved this, but no change has been made.

CTrando: A change has been made.

CTrando:
For testing:

* Functions defined in the same scope

* Nested functions defined in the same scope

* Methods defined in the same class - different parameters etc.

* Functions defined in other modules and imported into global scope doesn't given diagnostics

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
```diff
@@ -104,12 +105,15 @@ internal sealed class SymbolCollector : PythonWalker {
         }
 
         private void AddFunction(FunctionDefinition fd, IPythonType declaringType) {
-            if (!(_eval.LookupNameInScopes(fd.Name, LookupOptions.Local) is PythonFunctionType existing)) {
+            if (_eval.LookupNameInScopes(fd.Name, LookupOptions.Local) is PythonFunctionType existing) {
```
MikhailArkhipov: `_eval.LookupNameInScopes(fd.Name, LookupOptions.Local)?.MemberType == MemberType.Function`

MikhailArkhipov: What about property redefinition?

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
```diff
@@ -202,6 +213,21 @@ internal sealed class SymbolCollector : PythonWalker {
             return member;
         }
 
+        private void ReportRedefinedFunction(FunctionDefinition redefined, PythonType existing) {
```
MikhailArkhipov: `ILocatedMember` should be sufficient

MikhailArkhipov:
Resolve conflicts and LGTM.

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
CTrando: yup!

PR #1277, opened 2019-07-03
When a class inherits from something that is not a class, give a diagnostic error

CTrando:
same as inherit-non-class for pylint

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -115,7 +115,6 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
             foreach (var a in _classDef.Bases.Where(a => string.IsNullOrEmpty(a.Name))) {
                 var expr = a.Expression;
 
-                // A base can either be the name of a class declared in the current module or a class declared in some other module
                 switch (expr) {
                     // class declared in the current module
                     case NameExpression nameExpression:
@@ -140,6 +139,7 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
         }
 
         private IPythonType TryAddBase(List<IPythonType> bases, Arg arg) {
+            // We cheat slightly and treat base classes as annotations.
             var b = Eval.GetTypeFromAnnotation(arg.Expression);
             if (b != null) {
                 var t = b.GetPythonType();
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -13,6 +13,7 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Python.Analysis.Analyzer.Evaluation;
@@ -110,7 +111,7 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
             UpdateClassMembers();
         }
 
-        private List<IPythonType> ProcessBases(Scope outerScope) {
+        private IEnumerable<IPythonType> ProcessBases(Scope outerScope) {
             var bases = new List<IPythonType>();
             foreach (var a in _classDef.Bases.Where(a => string.IsNullOrEmpty(a.Name))) {
                 var expr = a.Expression;
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -53,8 +53,7 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
 
                 _class = classInfo;
 
-                List<IPythonType> bases = new List<IPythonType>();
-                ProcessBases(bases, outerScope);
+                List<IPythonType> bases = ProcessBases(outerScope);
 
                 _class.SetBases(bases);
                 // Declare __class__ variable in the scope.
@@ -111,7 +110,8 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
             UpdateClassMembers();
         }
 
-        private void ProcessBases(List<IPythonType> bases, Scope outerScope) {
+        private List<IPythonType> ProcessBases(Scope outerScope) {
+            List<IPythonType> bases = new List<IPythonType>();
             foreach (var a in _classDef.Bases.Where(a => string.IsNullOrEmpty(a.Name))) {
                 var expr = a.Expression;
 
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -136,6 +136,7 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
                         break;
                 }
             }
+            return bases;
         }
 
         private void TryAddBase(List<IPythonType> bases, Arg arg) {
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -53,7 +53,7 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
 
                 _class = classInfo;
 
-                List<IPythonType> bases = ProcessBases(outerScope);
+                var bases = ProcessBases(outerScope);
 
                 _class.SetBases(bases);
                 // Declare __class__ variable in the scope.
@@ -111,7 +111,7 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
         }
 
         private List<IPythonType> ProcessBases(Scope outerScope) {
-            List<IPythonType> bases = new List<IPythonType>();
+            var bases = new List<IPythonType>();
             foreach (var a in _classDef.Bases.Where(a => string.IsNullOrEmpty(a.Name))) {
                 var expr = a.Expression;
 
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -138,23 +138,17 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
             }
         }
 
-        private IPythonType TryAddBase(List<IPythonType> bases, Arg arg) {
+        private void TryAddBase(List<IPythonType> bases, Arg arg) {
             // We cheat slightly and treat base classes as annotations.
             var b = Eval.GetTypeFromAnnotation(arg.Expression);
             if (b != null) {
                 var t = b.GetPythonType();
                 bases.Add(t);
                 t.AddReference(Eval.GetLocationOfName(arg.Expression));
-                return t;
             } else {
                 ReportInvalidBase(arg.ToCodeString(Eval.Ast, CodeFormattingOptions.Traditional));
-                return null;
             }
         }
-
-
-
-
 
         private void EvaluateConstructors(ClassDefinition cd) {
             // Do not use foreach since walker list is dynamically modified and walkers are removed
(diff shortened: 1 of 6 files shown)


src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -53,19 +55,50 @@ internal sealed class ClassEvaluator : MemberEvaluator {
...
-                    var b = Eval.GetTypeFromAnnotation(a.Expression);
-                    if (b != null) {
-                        var t = b.GetPythonType();
-                        bases.Add(t);
-                        t.AddReference(Eval.GetLocationOfName(a.Expression));
+                    // can either be the name of a class declared in the current module
+                    // or a class declared in some other module
+                    var cur = a.Expression as NameExpression;
+                    var mod = a.Expression as MemberExpression;
+                    if (cur == null && mod == null) {
```
AlexanderSher: Please use `switch` pattern matching and consider better names than `cur` and `mod` to represent either a type of the expression (name/member) or the semantics of the variable. 
```cs
switch (a.Expression) {
    case NameExpression nameExpression:
...
        break;
    case MemberExpression memberExpression:
...
        break;
    default:
...
        break;
}
```

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -53,19 +55,50 @@ internal sealed class ClassEvaluator : MemberEvaluator {
...
+                        var arg = Eval.GetInScope(cur.Name, outerScope);
+                        if (arg is PythonClassType classType) {
+                            bases.Add(classType);
+                            // trying to extend a constant
+                        } else if (arg is IPythonConstant m) {
+                            Eval.ReportDiagnostics(
+                                Eval.Module.Uri,
+                                new DiagnosticsEntry(
+                                    Resources.InheritNonClass.FormatInvariant(m.Value),
+                                    _classDef.NameExpression.GetLocation(Eval.Module)?.Span ?? default,
```
AlexanderSher: These 4 lines are similar to the previous case. You can extract it into a separate method.

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -53,19 +55,50 @@ internal sealed class ClassEvaluator : MemberEvaluator {
...
+                                    Diagnostics.ErrorCodes.InheritNonClass,
+                                    Severity.Error,
+                                    DiagnosticSource.Analysis));
+                        }
+                    } else if (mod != null) {
+                        // It's a module class, try and find type from annotation
+                        var b = Eval.GetTypeFromAnnotation(a.Expression);
```
AlexanderSher: you can use a variable here instead of `a.Expression`

src/Analysis/Ast/Test/InheritNonClassTests.cs
```diff
@@ -0,0 +1,161 @@
...
+
+class C(x):
+    def method(self):
+        return 'test'
+";
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Diagnostics.Should().HaveCount(1);
+
+            var diagnostic = analysis.Diagnostics.ElementAt(0);
+            diagnostic.SourceSpan.Should().Be(4, 7, 4, 8);
```
AlexanderSher: If you have time, you can add an assertion for the `DiagnosticsEntry` to have better messages in tests.

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
CTrando: Yup!

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
CTrando: Done!

src/Analysis/Ast/Test/InheritNonClassTests.cs
CTrando: I'm not sure what you mean by this, could you explain more?

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
CTrando: Done!

src/Analysis/Ast/Test/InheritNonClassTests.cs
AlexanderSher: Take a look, for example, at `CompletionResultAssertions`

CTrando: Oh I see what you mean, so create a framework like ```CompletionResultAssertions``` so that it reads more smoothly!

jakebailey: Talked with @CTrando, he'll do this in another PR that adds it and updates all of the other tests to use it.

src/Analysis/Ast/Test/InheritNonClassTests.cs
```diff
@@ -0,0 +1,172 @@
...
+        public async Task InheritFromRenamedBuiltin() {
+            const string code = @"
+tmp = str
+
+class C(tmp):
+    def method(self):
+        return 'test'
+";
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Diagnostics.Should().HaveCount(0);
```
jakebailey: All of these `HaveCount`s should be `BeEmpty()`.

jakebailey:
Failing analysis tests to do with inheritance, so you'll have to figure out what's going on there before we can merge.

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -50,22 +52,41 @@ internal sealed class ClassEvaluator : MemberEvaluator {
...
                 var bases = new List<IPythonType>();
                 foreach (var a in _classDef.Bases.Where(a => string.IsNullOrEmpty(a.Name))) {
-                    // We cheat slightly and treat base classes as annotations.
-                    var b = Eval.GetTypeFromAnnotation(a.Expression);
-                    if (b != null) {
-                        var t = b.GetPythonType();
-                        bases.Add(t);
-                        t.AddReference(Eval.GetLocationOfName(a.Expression));
+                    // A base can either be the name of a class declared in the current module or a class declared in some other module
+                    switch (a.Expression) {
```
jakebailey: The old logic didn't check between Name/Member expressions, and the new logic does different work; what does the old code not do that the new code needs to do?

CTrando: The old code would add to bases everything that was put into the class

For example 

```
class C(5, 6):
  def test(self):
     pass

``` was valid.

I wanted to make sure that the only thing added was python classes, so I distinguished between NameExpressions of other classes and classes declared in other modules.

Comment (jakebailey) [reply src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs]: Makes sense, though you may want to go look at those failing tests. I think they're to do with bases which come from specialized things (like generics, or specifically specialized types).

File: src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs [hunk 7]
```diff
@@ -118,6 +111,51 @@ internal sealed class ClassEvaluator : MemberEvaluator {
...
+                bases.Add(t);
+                t.AddReference(Eval.GetLocationOfName(arg.Expression));
+                return t;
+            } else {
+                ReportInvalidBase(arg.ToCodeString(Eval.Ast, CodeFormattingOptions.Traditional));
+                return null;
+            }
+        }
+
+
```
jakebailey: Nit: whitespace

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -118,6 +111,51 @@ internal sealed class ClassEvaluator : MemberEvaluator {
...
+                        }
+                        break;
+                    default:
+                        TryAddBase(bases, a);
+                        break;
+                }
+            }
+        }
+
+        private IPythonType TryAddBase(List<IPythonType> bases, Arg arg) {
```
jakebailey: The return value for the function isn't used anywhere; did you mean to?

CTrando: good catch that's my bad

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -50,22 +52,13 @@ internal sealed class ClassEvaluator : MemberEvaluator {
...
-                    var b = Eval.GetTypeFromAnnotation(a.Expression);
-                    if (b != null) {
-                        var t = b.GetPythonType();
-                        bases.Add(t);
-                        t.AddReference(Eval.GetLocationOfName(a.Expression));
-                    }
-                }
-                _class.SetBases(bases);
 
+                List<IPythonType> bases = new List<IPythonType>();
```
jakebailey: This seems weird; why doesn't `ProcessBases` just return an `IEnumerable`, instead of passing in an empty `List` to be modified?

src/Analysis/Ast/Test/InheritNonClassTests.cs
```diff
@@ -0,0 +1,172 @@
...
+            diagnostic.ErrorCode.Should().Be(ErrorCodes.InheritNonClass);
+
+            diagnostic = analysis.Diagnostics.ElementAt(1);
+            diagnostic.SourceSpan.Should().Be(10, 7, 10, 8);
+            diagnostic.Message.Should().Be(Resources.InheritNonClass.FormatInvariant("5"));
+            diagnostic.ErrorCode.Should().Be(ErrorCodes.InheritNonClass);
+        }
+
+        [TestMethod, Priority(0)]
+        public async Task InheritFromBinaryOp() {
```
jakebailey: This test is currently failing, it seems. Not showing a diagnostic is safer than showing a wrong one, though.

CTrando: Yeah, so this involved how all those tests were failing before. I tried to be too strict on what we can inherit and that caused a lot of behavior to fail. There's nothing I can really do here with binary operations, we can make the argument that a binary operation will never resolve to a class type, and under that assumption we can have a separate check for binary operations where we report them, but I'm more inclined to leave it alone and remove this test.

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
CTrando: I was trying to follow the ```Process...``` style, where they were void methods - but you're right I thought it was weird too.

src/Analysis/Ast/Test/InheritNonClassTests.cs
jakebailey: That, or stick an `[Ignore]` on it to signify that the test should pass, but doesn't for now.

CTrando: Ok sounds good.

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -50,22 +52,12 @@ internal sealed class ClassEvaluator : MemberEvaluator {
...
-                    var b = Eval.GetTypeFromAnnotation(a.Expression);
-                    if (b != null) {
-                        var t = b.GetPythonType();
-                        bases.Add(t);
-                        t.AddReference(Eval.GetLocationOfName(a.Expression));
-                    }
-                }
-                _class.SetBases(bases);
 
+                List<IPythonType> bases = ProcessBases(outerScope);
```
jakebailey: `var`

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -118,6 +110,47 @@ internal sealed class ClassEvaluator : MemberEvaluator {
             UpdateClassMembers();
         }
 
+        private List<IPythonType> ProcessBases(Scope outerScope) {
+            List<IPythonType> bases = new List<IPythonType>();
```
jakebailey: var

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
CTrando: Sorry, my Java is showing!

MikhailArkhipov:
Reverted. Broke ClassBase test in reference tests - no matter what class derives from `AddReference` must be called on the base. Also, it is possible to derive from functions, etc so more specific checks are needed.

PR #1279, opened 2019-07-03
Give a diagnostic when creating a function in a class and not specifying the first argument as self

CTrando:
Waiting on #1273 before I can merge this one.

e.g

```
class Test:
    def hello(x, y, z):
         pass
```


jakebailey:
I don't think we should merge this one, at least not for now. Although calling the first argument "self" is convention, it's not a runtime-requirement and code which doesn't say `self` will still work if the first variable is used correctly.

(Arguably, we should be trying to rely on the name `self` less and less, but that's another issue.)

MikhailArkhipov:
It is an error per http://pylint-messages.wikidot.com/messages:e0213. As in big VS, one can always suppress it via settings/pragmas. By default though, VS shows violations or common conventions although said violations may not always translate to crashes.

https://legacy.python.org/dev/peps/pep-0008/#function-and-method-arguments
Always use self for the first argument to instance methods.
Always use cls for the first argument to class methods.

src/Analysis/Ast/Test/LintNoSelfArgumentTests.cs
```diff
@@ -0,0 +1,173 @@
...
+        [TestMethod, Priority(0)]
+        public async Task FirstArgumentClassMethod() {
+            const string code = @"
+class Test:
+    @classmethod
+    def test(arg1, arg2):
+        pass
+";
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Diagnostics.Should().HaveCount(0);
```
jakebailey: `BeEmpty`

src/Analysis/Ast/Test/LintNoSelfArgumentTests.cs
```diff
@@ -0,0 +1,173 @@
...
+            diagnostic.ErrorCode.Should().Be(ErrorCodes.NoSelfArgument);
+            diagnostic.SourceSpan.Should().Be(6, 9, 6, 14);
+            diagnostic.Message.Should().Be(Resources.NoSelfArgument);
+        }
+
+        [TestMethod, Priority(0)]
+        public async Task FirstArgumentClassMethod() {
+            const string code = @"
+class Test:
+    @classmethod
```
jakebailey: What happens when the method has other decorators than `classmethod`, potentially in different orders?

src/Analysis/Ast/Test/LintNoSelfArgumentTests.cs
jakebailey: And everywhere else.

src/Analysis/Ast/Test/LintNoSelfArgumentTests.cs
CTrando: We process only the builtin decorators, see `PythonFunctionType.ProcessDecorators`.

src/Analysis/Ast/Test/LintNoSelfArgumentTests.cs
CTrando: Ok!

CTrando:
WIP until I can add no class argument and no method argument.

brettcannon:
Not using `self` as the name of the first parameter of an instance method is not an error and potentially done on purpose (e.g. if you are using `**kwargs` and you don't want `self` to be captured as a positional or keyword argument). It also won't work if you use `staticmethod` (where it doesn't make sense), or `classmethod` (which is traditionally `cls`). This also doesn't cover other clever decorators, metaclasses, etc.

If this being done as a stylistic thing then I would leave to a style linter like Pylint. If the worry is someone mis-typed `self` then the preexisting check for undefined variables should catch that. And if it's due to someone simply forgetting to include `self` then tests should capture that, but on its own I would say that since this isn't an obvious run-time error that this is going to be difficult to get right like the redefinition check.

MikhailArkhipov:
Per links above it is considered an error in pylint which then refers to https://legacy.python.org/dev/peps/pep-0008/#function-and-method-arguments. Tests may be useful, but in C# I's rather see that I mistyped `this` to `ths` before I even get to compilation. VS editor does not require even to build to see issues. 

Besides, some people never write tests and that is their choice. We can't really insist on 'write tests and that will catch the error'.

src/Analysis/Ast/Impl/Analyzer/Symbols/FunctionEvaluator.cs
```diff
@@ -98,6 +100,95 @@ public FunctionEvaluator(ExpressionEval eval, PythonFunctionOverload overload)
...
+        /// Returns if the function is part of a metaclass definition 
+        /// e.g 
+        /// class A(type):
+        ///  def f(cls): ...
+        /// f is a metaclass function
+        /// </summary>
+        /// <returns></returns>
+        private bool SelfIsMetaclass() {
+            // Just allow all specialized types in Mro to avoid false positives
+            return _self.Mro.Any(b => b.IsSpecialized);
```
CTrando: Need to check for Type in the Mro, don't know an easy way besides this

src/Analysis/Ast/Impl/Extensions/PythonFunctionExtensions.cs
```diff
@@ -34,5 +38,44 @@ public static bool HasClassFirstArgument(this IPythonClassMember m)
...
         }
+
+        /// <summary>
+        /// Reports any decorator errors on function.
+        /// Returns true if the decorator combinations for the property are valid
+        /// </summary>
+        public static bool HasValidDecorators(this IPythonFunctionType f, IExpressionEvaluator eval) {
+            bool valid = true;
+            // If function is abstract, allow all decorators because it will be overridden
+            if (f.IsAbstract) {
```
CTrando: I looked at [this](https://stackoverflow.com/questions/45248243/most-pythonic-way-to-declare-an-abstract-class-property) for the abstract check

src/Analysis/Ast/Impl/Extensions/PythonPropertyExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+                    case @"classmethod":
+                        ReportInvalidDecorator(dec, Resources.InvalidDecoratorForProperty.FormatInvariant("Classmethods"), eval);
+                        valid = false;
+                        break;
+                }
+            }
+            return valid;
+        }
+
+        private static void ReportInvalidDecorator(NameExpression name, string errorMsg, IExpressionEvaluator eval) {
```
CTrando: Very similar to the one in `PythonFunctionExtensions`, any idea how to remove the duplication?

CTrando:
Features:

Gives diagnostics on:

* Not using`self` in a regular function 
```
class A: 
  def tmp(x, y, z): ...
```

* Not using `cls` in a class method or an abstract class method
```
class A: 
  @classmethod / @abstractclassmethod
  def tmp(x, y, z): ...
```

* Not using `cls` in an automatic class method, I have `__new__`, `__init_subclass__`, `__class_getitem__`.
```
class A: 
  def __new__(x, y, z): ...
```

No diagnostics on:
* Methods declared with `@staticmethod`
* Methods declared in a class that is in a metaclass chain, (has type in its Mro) 

What I don't handle:
Decorators with somewhat unorthodox behavior such as `singledispatch`.

I could add something that checks if there are any unknown decorators on the function, don't give any diagnostics on the parameters.

src/Analysis/Ast/Impl/Types/PythonFunctionType.cs
```diff
@@ -25,6 +25,7 @@
 namespace Microsoft.Python.Analysis.Types {
     [DebuggerDisplay("Function {Name} ({TypeId})")]
     internal sealed class PythonFunctionType : PythonType, IPythonFunctionType {
+        private readonly IReadOnlyList<string> DefaultClassMethods = new[] { "__new__", "__init_subclass__", "__class_getitem__" };
```
MikhailArkhipov: Hashset is probably better for `Contains`

jakebailey: Yeah, with 3 items, an array is almost definitely going to be faster.

jakebailey: It should however be a static member; it doesn't need to be allocated for every `PythonFunctionType`.

src/Analysis/Ast/Impl/Extensions/PythonPropertyExtensions.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// Returns true if the decorator combiantions for the property are valid
+        /// </summary>
+        public static bool HasValidDecorators(this IPythonPropertyType p, IExpressionEvaluator eval) {
+            bool valid = true;
+            // If property is abstract, allow all decorators because it will be overridden
+            if(p.IsAbstract) {
+                return valid;
+            }
+
+            foreach (var dec in (p.FunctionDefinition?.Decorators?.Decorators).MaybeEnumerate().OfType<NameExpression>()) {
```
MikhailArkhipov: Since we already walk decorators and set flags in `ProcessDecorators`, do we need to do it again? Or can we just check odd combinations of the flags set? I.e. `IsStatic && IsClassMethod` then report error?

src/Analysis/Ast/Impl/Extensions/PythonPropertyExtensions.cs
MikhailArkhipov: Instead of passing type, pass actual flags. I.e. `HasValidDecorators(bool isAbstract, bool isClassMethod, ...)`

src/Analysis/Ast/Impl/Types/PythonFunctionType.cs
CTrando: OK

src/Analysis/Ast/Impl/Extensions/PythonPropertyExtensions.cs
CTrando: I loop over decorators so I can highlight their locations. If it was fine to just generically highlight the function definition instead, then I can use the flag method, but if I want to highlight the decorators themselves I need their locations.

src/Analysis/Ast/Impl/Analyzer/Symbols/FunctionEvaluator.cs
```diff
@@ -119,10 +225,10 @@ public FunctionEvaluator(ExpressionEval eval, PythonFunctionOverload overload)
             var value = Eval.GetValueFromExpression(node.Expression);
             if (value != null) {
                 // although technically legal, __init__ in a constructor should not have a not-none return value
-                if (FunctionDefinition.Name.EqualsOrdinal("__init__") && _function.DeclaringType.MemberType == PythonMemberType.Class 
-                    && !value.IsOfType(BuiltinTypeId.NoneType)) { 
+                if (FunctionDefinition.Name.EqualsOrdinal("__init__") && _function.DeclaringType.MemberType == PythonMemberType.Class
```
MikhailArkhipov: See https://github.com/microsoft/python-language-server/issues/1401. You may want to make an extension to check it is `__init__` constructor.

CTrando: Ok

src/Analysis/Ast/Impl/Extensions/PythonTypeExtensions.cs
```diff
@@ -37,8 +37,5 @@ public static bool IsGeneric(this IPythonType value)
                 dst.Location = src.Location;
             }
         }
-
```
CTrando: Removing because was unused and does not check for declaration type and null check for name.

src/Analysis/Ast/Impl/Analyzer/Symbols/FunctionEvaluator.cs
```diff
@@ -19,6 +19,7 @@
 using System.Linq;
 using Microsoft.Python.Analysis.Analyzer.Evaluation;
 using Microsoft.Python.Analysis.Diagnostics;
+using Microsoft.Python.Analysis.Extensions;
```
MikhailArkhipov: Namespace for extensions is `Microsoft.Python.Analysis`. We don't add separate namespace for them.

CTrando: Ok

PR #1290, opened 2019-07-08
[WIP]: Handle inheriting from the ABC module to create abstract classes

CTrando:
Does not handle metadata=ABCMetadata, will probably try that later. 

IsAbstract is now set correctly for abstract classes. 

Basically a class is abstract if it derives from an abstract class and has abstract methods.

A class derives from an abstract class if any of its bases derive from abstract classes or are abstract classes.

A class is abstract if it extends ABC from package abc.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -155,7 +155,7 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
             // A class is derived from an abstract class if one of its bases is abstract or one of its bases is also derived from an abstract class
             var derivedFromAbstract = _class.Bases.OfType<PythonClassType>().Any(b => b.IsAbstract || b.IsDerivedFromAbstractClass);
             // A class is abstract if it is derived from an abstract class and has at least one abstract function
-            var isAbstract = _class.IsDerivedFromAbstractClass && _class.GetMembers<PythonFunctionType>().Any(f => f.IsAbstract);
+            var isAbstract = derivedFromAbstract && _class.GetMembers<PythonFunctionType>().Any(f => f.IsAbstract);
 
             _class.IsDerivedFromAbstractClass = derivedFromAbstract;
             _class.IsAbstract = isAbstract;
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -68,7 +68,6 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
 
                 ProcessClassBody();
 
-                // TODO do abstract check here 
                 _class.DecideAbstract();
             }
         }
@@ -119,8 +118,6 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
             // Process remaining methods.
             SymbolTable.EvaluateScope(_classDef);
             UpdateClassMembers();
-
-            _class.DecideAbstract();
         }
 
         private void EvaluateConstructors(ClassDefinition cd) {
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -69,7 +69,7 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
                 ProcessClassBody();
 
                 // TODO do abstract check here 
-                DecideAbstract();
+                _class.DecideAbstract();
             }
         }
 
@@ -119,10 +119,8 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
             // Process remaining methods.
             SymbolTable.EvaluateScope(_classDef);
             UpdateClassMembers();
-        }
 
-        private void DecideAbstract() {
-            _class.SetAbstract(true);
+            _class.DecideAbstract();
         }
 
         private void EvaluateConstructors(ClassDefinition cd) {
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -67,6 +67,9 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
                 Eval.DeclareVariable("__class__", _class, VariableSource.Declaration);
 
                 ProcessClassBody();
+
+                // TODO do abstract check here 
+                DecideAbstract();
             }
         }
 
@@ -116,6 +119,10 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
             // Process remaining methods.
             SymbolTable.EvaluateScope(_classDef);
             UpdateClassMembers();
+        }
+
+        private void DecideAbstract() {
+            _class.SetAbstract(true);
         }
 
         private void EvaluateConstructors(ClassDefinition cd) {
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -68,7 +68,7 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
 
                 ProcessClassBody();
 
-                _class.DecideAbstract();
+                DecideAbstract();
             }
         }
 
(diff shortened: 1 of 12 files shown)


src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -146,6 +148,19 @@ internal sealed class ClassEvaluator : MemberEvaluator {
             }
         }
 
+        /// <summary>
+        /// Decides if this class is an abstract class or not
+        /// </summary>
+        private void DecideAbstract() {
+            // A class is derived from an abstract class if one of its bases is abstract or one of its bases is also derived from an abstract class
+            var derivedFromAbstract = _class.Bases.OfType<PythonClassType>().Any(b => b.IsAbstract || b.IsDerivedFromAbstractClass);
```
MikhailArkhipov: `OfType<IPythonClassType>` or, better still, `MemberType == PythonMemberType.Class`

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -146,6 +148,19 @@ internal sealed class ClassEvaluator : MemberEvaluator {
...
+            // A class is abstract if it is derived from an abstract class and has at least one abstract function
+            var isAbstract = derivedFromAbstract && _class.GetMembers<PythonFunctionType>().Any(f => f.IsAbstract);
```
MikhailArkhipov: `IPythonFunctionType`. Are there abstract properties? https://docs.python.org/3/library/abc.html#abc.abstractproperty

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
MikhailArkhipov: This might be a useful extension to `IPythonClassType` i.s. `IsAbstract(this ...)`

src/Analysis/Ast/Impl/Specializations/Abc/AbcModule.cs
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        #region IMemberContainer
+        public override IMember GetMember(string name) => _members.TryGetValue(name, out var m) ? m : null;
+        public override IEnumerable<string> GetMemberNames() => _members.Keys;
+        #endregion
+
+        private void SpecializeMembers() {
+            // ABC
+            var cls = PythonClassType.Specialize("ABC", this, GetMemberDocumentation("ABC"), isAbstract: true);
```
MikhailArkhipov: If we specialize module, we no longer look into the original one. Are there any members which we may lose as we only specialize one?

src/Analysis/Ast/Impl/Types/Definitions/IPythonClassType.cs
```diff
@@ -36,6 +36,11 @@ public interface IPythonClassType : IPythonType, ILocatedMember {
         /// </summary>
         IReadOnlyList<IPythonType> Bases { get; }
 
+        /// <summary>
+        /// If the class contains an abstract class on its inheritance chain
+        /// </summary>
+        bool IsDerivedFromAbstractClass { get; set;  }
```
MikhailArkhipov: Probably should be an extension as mentioned above.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -38,12 +38,26 @@ internal class PythonClassType : PythonType, IPythonClassType, IPythonTemplateTy
         private Dictionary<string, IPythonType> _genericParameters;
         private string _documentation;
 
+        /// <summary>
+        /// Creates function for specializations
```
MikhailArkhipov: Creates _class_ for specialization.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -38,12 +38,26 @@ internal class PythonClassType : PythonType, IPythonClassType, IPythonTemplateTy
...
+        /// </summary>
+        public static PythonClassType Specialize(string name, IPythonModule declaringModule, string documentation, bool isAbstract = false) {
+            return new PythonClassType(name, new Location(declaringModule, default), documentation, isAbstract: isAbstract);
```
MikhailArkhipov: I think `new Location(declaringModule)` will do.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -38,12 +38,26 @@ internal class PythonClassType : PythonType, IPythonClassType, IPythonTemplateTy
         private Dictionary<string, IPythonType> _genericParameters;
         private string _documentation;
 
+        /// <summary>
+        /// Creates function for specializations
+        /// </summary>
+        public static PythonClassType Specialize(string name, IPythonModule declaringModule, string documentation, bool isAbstract = false) {
```
MikhailArkhipov: Does class specialization actually needs `isAbstract`? I think we determine it from methods/bases, right?

src/Analysis/Ast/Impl/Types/PythonType.cs
```diff
@@ -59,7 +59,10 @@ internal class PythonType : LocatedMember, IPythonType {//, IEquatable<IPythonTy
         public virtual string Documentation { get; private set; }
         public virtual BuiltinTypeId TypeId => _typeId;
         public bool IsBuiltin => DeclaringModule == null || DeclaringModule is IBuiltinsPythonModule;
-        public virtual bool IsAbstract => false;
+        public virtual bool IsAbstract {
```
MikhailArkhipov: Should not allow setters on public interface

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
CTrando: Sounds good.

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
CTrando: The link you gave says abstractproperties are deprecated. But for Python 2.7 I believe we should handle this.

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
CTrando: I set isAbstract on the class type in order to prevent doing more computations each time someone references isAbstract. I wouldn't think it would be an extension but rather a property. Do you think that would be better than computing it once and setting it on the class type? It would technically always be correct even if someone added new bases or changed methods.

src/Analysis/Ast/Impl/Specializations/Abc/AbcModule.cs
CTrando: We lose ABCMeta, which I plan to handle once I figure out how to handle metaclasses. But I believe that is it. Relevant link is: https://github.com/python/cpython/blob/master/Lib/abc.py

src/Analysis/Ast/Impl/Types/Definitions/IPythonClassType.cs
CTrando: so take for example:

```
class A:
  @abstractmethod
  def tmp(self): 
     pass

class B(A):
  def tmp(self):
     pass

class C(B):
  @abstractmethod 
  def tmp2(self):
    pass
```

Class C is still considered abstract even though it inherits a class that is not abstract. Therefore I say that C is derived from an abstract class, which basically means C (and any descendants) can become abstract, even if they do not inherit directly from abstract classes.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
CTrando: Ok

src/Analysis/Ast/Impl/Types/PythonClassType.cs
CTrando: This is to specialize abc.ABC as an abstract class (sort of the root abstract class)

src/Analysis/Ast/Impl/Types/PythonClassType.cs
CTrando: ok

src/Analysis/Ast/Impl/Specializations/Abc/AbcModule.cs
MikhailArkhipov: So I guess this PR should be in `WIP` unless we are sure users won't complain that `ABCMeta` is no longer in completions?

src/Analysis/Ast/Impl/Types/PythonType.cs
CTrando: Can you give me your reasoning for this?

MikhailArkhipov: So users of the analysis, such as extensions like IntelliCode would not be able to modify completed analysis.

src/Analysis/Ast/Impl/Specializations/Abc/AbcModule.cs
CTrando: Yup, I'll do that.

jakebailey:
Closing for now.

Issue #1174, opened 2019-06-06
All references are not listed when using ms language server


DonJayamanne:
@jahan01 commented on [Wed Jun 05 2019](https://github.com/microsoft/vscode-python/issues/5917)

## Environment data

- VS Code version: latest
- Extension version (available under the Extensions sidebar): 2019.5.18426
- OS and version: Mac Os
- Python version (& distribution if applicable, e.g. Anaconda): 3.6 (Anaconda)
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): conda
- Language Server (more info #3977): Language Server

`constants.py`
```python
VARIABLE1 = 'afad'
VARIABLE2 = 'dcef'
```

`use.py`
```python
from vscode_test.constants import *
from vscode_test import constants

print(VARIABLE1)
print(constants.VARIABLE2)

print(VARIABLE2)
print(constants.VARIABLE1)

```

## Expected behaviour

All usages of the symbol to be listed under references when `Right click -> Find All References` on `VARIABLE1` in `constants.py`

## Actual behaviour

Doesn't list all usages when ms language server is used.
Usage of `print(constants.VARIABLE1)` in `use.py`  isnt listed.

Looks like if symbol is imported directly, the reference is listed. If the symbol referenced via modules it isn't listed.

This issue seems to be recurring - why not have some tests for this ?

When Jedi is enabled it works as expected.

![image](https://user-images.githubusercontent.com/566264/58970523-30119100-87ec-11e9-83ce-4fbf904bfc31.png)

MikhailArkhipov:
Slightly simplified
```python
from .constants import *
import constants

print(VARIABLE1)
print(constants.VARIABLE2)

print(VARIABLE2)
print(constants.VARIABLE1)
```

MikhailArkhipov:
So the problem is that since expression is not assigned (i.s. as opposed to `x = func(a)`, none of the parts participate in analysis downstream, the arguments are not evaluated and hence reference is not added. Reference to `VARIABLE*` is added by the linter when it is looking up defined/undefined variables.

MikhailArkhipov:
Adding cost to eval arguments in all cases.

PR #1297, opened 2019-07-09
Multiple analysis fixes

### LINKED ISSUE #1151: Reliable way to determine final analysis of a module
Continuation of #1143.

This will
- improve performance since we can delay some activities, such as linting run until everything is settled for the module. 
- Prevent unnecessary reloads and reparses when AST was dropped too early per https://github.com/microsoft/python-language-server/pull/1134 for #1088 .
- Better determine when module analysis can be persisted into a database.

One possible way is to simulate analysis run, i.e.

Run all sessions but with only parsing in order determine final session and then
Run analysis without parsing repeating the above.
Technically end point should be the same.

### LINKED ISSUE #1201: NRE in SymbolCollector.AddProperty
```
Microsoft.Python.Analysis.Analyzer.Symbols.SymbolCollector.AddProperty(FunctionDefinition, IPythonType, Boolean) in Analysis\Ast\Impl\Analyzer\Symbols\SymbolCollector.cs:line 170
Microsoft.Python.Analysis.Analyzer.Symbols.SymbolCollector.TryAddProperty(FunctionDefinition, IPythonType) in Analysis\Ast\Impl\Analyzer\Symbols\SymbolCollector.cs:line 158
Microsoft.Python.Analysis.Analyzer.Symbols.SymbolCollector.AddFunctionOrProperty(FunctionDefinition) in Analysis\Ast\Impl\Analyzer\Symbols\SymbolCollector.cs:line 101
Microsoft.Python.Analysis.Analyzer.Symbols.SymbolCollector.Walk(FunctionDefinition) in Analysis\Ast\Impl\Analyzer\Symbols\SymbolCollector.cs:line 77
Microsoft.Python.Parsing.Ast.FunctionDefinition.Walk(PythonWalker) in Parsing\Impl\Ast\FunctionDefinition.cs:line 197
```

Not sure what could be null at line 170, though. Maybe it's getting optimized and `fd` is null and is breaking at `fd.Name`.

### LINKED ISSUE #1228: NRE in TryFindMissingDependencies
```
Microsoft.Python.Analysis.Dependencies.DependencyResolver`2.TryFindMissingDependencies(ImmutableArray`1&, ImmutableArray`1&, Int32) in Analysis\Ast\Impl\Dependencies\DependencyResolver.cs:line 563
Microsoft.Python.Analysis.Dependencies.DependencyResolver`2.TryCreateWalker(Int32, Int32, IDependencyChainWalker`2&) in Analysis\Ast\Impl\Dependencies\DependencyResolver.cs:line 214
Microsoft.Python.Analysis.Analyzer.PythonAnalyzer.TryCreateSession(Int32, PythonAnalyzerEntry, PythonAnalyzerSession&) in Analysis\Ast\Impl\Analyzer\PythonAnalyzer.cs:line 246
Microsoft.Python.Analysis.Analyzer.PythonAnalyzer.AnalyzeDocument(AnalysisModuleKey, PythonAnalyzerEntry, ImmutableArray`1) in Analysis\Ast\Impl\Analyzer\PythonAnalyzer.cs:line 224
```

Line 563 is just `version != _version`, neither of which can ever be null (they're `int`!!!). Happening just as often as the GetBodyStartIndex NRE.

cc @AlexanderSher

### LINKED ISSUE #1294: Handle valid case for reloading AST
1. Create module with code `import datetime`
2. Wait for analysis to complete. `LibraryAnalysis` will be created for `datetime` module
3. F12 on `datetime`

Opened module should be correctly handled.

### LINKED ISSUE #1295: FindReferences doesn't work on package init members
`package/__init__.py`:
```
x = 1
```
`module.py`:
```
import package
y = package.x
```

FindReferences doesn't show reference in `module.py`

### LINKED ISSUE #1296: DependencyResolver graph misses intermediate imports
If `module` is imported as `import package.module` and `package` has `__init__.py`, `package` won't be a dependency in DependencyResolver. As a result, `package` is analyzed later than it should and additional iteration of analysis is triggered.

AlexanderSher:
- Fixes #1151: Reliable way to determine final analysis of a module
- Fixes #1201: NRE in SymbolCollector.AddProperty
- Fixes #1228: NRE in TryFindMissingDependencies
- Fixes #1294: Handle valid case for reloading AST
- Fixes #1295: FindReferences doesn't work on package init members
- Fixes #1296: DependencyResolver graph misses intermediate imports
- Part of the fix for #1174: All references are not listed when using ms language server (bug isn't fully fixed).

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/ActivityTracker.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/ActivityTracker.cs
@@ -19,15 +19,10 @@ using System.Linq;
 
 namespace Microsoft.Python.Analysis.Analyzer {
     internal static class ActivityTracker {
-        private static readonly Dictionary<string, AnalysisState> _modules = new Dictionary<string, AnalysisState>();
+        private static readonly HashSet<string> _modules = new HashSet<string>();
         private static readonly object _lock = new object();
         private static bool _tracking;
         private static Stopwatch _sw;
-
-        private struct AnalysisState {
-            public int Count;
-            public bool IsComplete;
-        }
 
         public static void OnEnqueueModule(string path) {
             if (string.IsNullOrEmpty(path)) {
@@ -35,31 +30,9 @@ namespace Microsoft.Python.Analysis.Analyzer {
             }
 
             lock (_lock) {
-                if (!_modules.TryGetValue(path, out var st)) {
-                    _modules[path] = default;
-                } else {
-                    st.IsComplete = false;
-                }
+                _modules.Add(path);
             }
         }
-
-        public static void OnModuleAnalysisComplete(string path) {
-            lock (_lock) {
-                if (_modules.TryGetValue(path, out var st)) {
-                    st.Count++;
-                    st.IsComplete = true;
-                }
-            }
-        }
-
-        public static bool IsAnalysisComplete {
-            get {
-                lock (_lock) {
-                    return _modules.All(m => m.Value.IsComplete);
-                }
-            }
-        }
-
 
         public static void StartTracking() {
             lock (_lock) {
@@ -71,22 +44,15 @@ namespace Microsoft.Python.Analysis.Analyzer {
             }
         }
 
-        public static void EndTracking() {
+        public static (int modulesCount, double totalMilliseconds) EndTracking() {
             lock (_lock) {
                 if (_tracking) {
                     _sw?.Stop();
                     _tracking = false;
                 }
-            }
-        }
 
-        public static int ModuleCount {
-            get {
-                lock (_lock) {
-                    return _modules.Count;
-                }
+                return (_modules.Count, _sw?.Elapsed.TotalMilliseconds ?? 0);
             }
         }
-        public static double MillisecondsElapsed => _sw?.Elapsed.TotalMilliseconds ?? 0;
     }
 }
--- a/src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
@@ -15,31 +15,48 @@
 
 using System;
 using System.Diagnostics;
+using Microsoft.Python.Analysis.Documents;
 using Microsoft.Python.Analysis.Modules;
 using Microsoft.Python.Analysis.Types;
 using Microsoft.Python.Core;
 
 namespace Microsoft.Python.Analysis.Analyzer {
     [DebuggerDisplay("{Name} : {FilePath}")]
-    internal struct AnalysisModuleKey : IEquatable<AnalysisModuleKey> {
+    internal readonly struct AnalysisModuleKey : IEquatable<AnalysisModuleKey> {
+        private enum KeyType { Default, Typeshed, LibraryAsDocument }
+
+        private readonly KeyType _type;
         public string Name { get; }
         public string FilePath { get; }
-        public bool IsTypeshed { get; }
+        public bool IsTypeshed => _type == KeyType.Typeshed;
+        public bool IsLibraryAsDocument => _type == KeyType.LibraryAsDocument;
 
         public AnalysisModuleKey(IPythonModule module) {
             Name = module.Name;
             FilePath = module.ModuleType == ModuleType.CompiledBuiltin ? null : module.FilePath;
-            IsTypeshed = module is StubPythonModule stub && stub.IsTypeshed;
+            _type = module is StubPythonModule stub && stub.IsTypeshed
+                ? KeyType.Typeshed
+                : module.ModuleType == ModuleType.Library && module is IDocument document && document.IsOpen
+                    ? KeyType.LibraryAsDocument
+                    : KeyType.Default;
         }
 
         public AnalysisModuleKey(string name, string filePath, bool isTypeshed) {
             Name = name;
             FilePath = filePath;
-            IsTypeshed = isTypeshed;
+            _type = isTypeshed ? KeyType.Typeshed : KeyType.Default;
         }
 
+        private AnalysisModuleKey(string name, string filePath, KeyType type) {
+            Name = name;
+            FilePath = filePath;
+            _type = type;
+        }
+
+        public AnalysisModuleKey GetLibraryAsDocumentKey() => new AnalysisModuleKey(Name, FilePath, KeyType.LibraryAsDocument);
+
         public bool Equals(AnalysisModuleKey other)
-            => Name.EqualsOrdinal(other.Name) && FilePath.PathEquals(other.FilePath) && IsTypeshed == other.IsTypeshed;
+            => Name.EqualsOrdinal(other.Name) && FilePath.PathEquals(other.FilePath) && _type == other._type;
 
         public override bool Equals(object obj) => obj is AnalysisModuleKey other && Equals(other);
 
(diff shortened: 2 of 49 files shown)


src/Analysis/Ast/Impl/Analyzer/ActivityTracker.cs
```diff
@@ -19,48 +19,21 @@
 
 namespace Microsoft.Python.Analysis.Analyzer {
     internal static class ActivityTracker {
-        private static readonly Dictionary<string, AnalysisState> _modules = new Dictionary<string, AnalysisState>();
+        private static readonly HashSet<string> _modules = new HashSet<string>();
```
AlexanderSher: `AnalysisState` has never been used

src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
```diff
@@ -15,39 +15,56 @@
...
-    internal struct AnalysisModuleKey : IEquatable<AnalysisModuleKey> {
+    internal readonly struct AnalysisModuleKey : IEquatable<AnalysisModuleKey> {
+        private enum KeyType { Default, Typeshed, LibraryAsDocument }
+
+        private readonly KeyType _type;
         public string Name { get; }
         public string FilePath { get; }
-        public bool IsTypeshed { get; }
+        public bool IsTypeshed => _type == KeyType.Typeshed;
+        public bool IsLibraryAsDocument => _type == KeyType.LibraryAsDocument;
```
AlexanderSher: This is a change for the #1294. Opened library module may be part of the loop, so with `LibraryAnalysis` it will trigger reparsing of all the modules in the loop. As a workaroung, we create new instance of the `PythonAnalyzerEntry` that will contain `DocumentAnalysis`.

src/Analysis/Ast/Impl/Analyzer/Definitions/IPythonAnalyzer.cs
```diff
@@ -20,14 +20,15 @@
...
+using Microsoft.Python.Parsing.Ast;
 
 namespace Microsoft.Python.Analysis.Analyzer {
     public interface IPythonAnalyzer {
         Task WaitForCompleteAnalysisAsync(CancellationToken cancellationToken = default);
         /// <summary>
         /// Schedules module for analysis. Module will be scheduled if version of AST is greater than the one used to get previous analysis
         /// </summary>
-        void EnqueueDocumentForAnalysis(IPythonModule module, int version);
+        void EnqueueDocumentForAnalysis(IPythonModule module, PythonAst ast, int version);
```
AlexanderSher: There were several cases when `ast` was out of sync, so we need to pass to `PythonAnalyzer` explicitly.

src/Analysis/Ast/Impl/Analyzer/Handlers/ImportHandler.cs
```diff
@@ -56,19 +56,22 @@ internal sealed partial class ImportHandler : StatementHandler {
             // import_module('fob.oar.baz')
             var importNames = ImmutableArray<string>.Empty;
             var variableModule = default(PythonVariableModule);
+            var importedNamesCount = 0;
             foreach (var nameExpression in moduleImportExpression.Names) {
                 importNames = importNames.Add(nameExpression.Name);
                 var imports = ModuleResolution.CurrentPathResolver.GetImportsFromAbsoluteName(Module.FilePath, importNames, forceAbsolute);
                 if (!HandleImportSearchResult(imports, variableModule, asNameExpression, moduleImportExpression, out variableModule)) {
-                    return;
+                    break;
```
AlexanderSher: This handles a case when `import package.subpackage.module` fails on `subpackage` part, but `package` has `__init__.py` and successfully imported.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -120,11 +122,13 @@ public Task WaitForCompleteAnalysisAsync(CancellationToken cancellationToken = d
         }
 
         public void RemoveAnalysis(IPythonModule module) {
+            AnalysisModuleKey key;
             lock (_syncObj) {
-                _analysisEntries.Remove(new AnalysisModuleKey(module));
+                key = new AnalysisModuleKey(module);
```
AlexanderSher: Since `key` reads `IsOpen` property of the `IDocument`, we need to access it under lock.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -144,24 +148,36 @@ public Task WaitForCompleteAnalysisAsync(CancellationToken cancellationToken = d
...
+
             lock (_syncObj) {
+                key = new AnalysisModuleKey(module);
                 version = _version + 1;
                 if (_analysisEntries.TryGetValue(key, out entry)) {
                     if (entry.BufferVersion >= bufferVersion) {
                         return;
                     }
+
+                    // It is possible that parsing request for the library has been started when document is open,
```
AlexanderSher: This happens when `Go to definition` is done by `Ctrl+Click`. We got three events in a row: `didOpen`, `didClose`, `didOpen`.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -205,12 +221,12 @@ public Task WaitForCompleteAnalysisAsync(CancellationToken cancellationToken = d
...
             => AnalysisComplete?.Invoke(this, new AnalysisCompleteEventArgs(moduleCount, msElapsed));
 
-        private void AnalyzeDocument(AnalysisModuleKey key, PythonAnalyzerEntry entry, ImmutableArray<AnalysisModuleKey> dependencies) {
+        private void AnalyzeDocument(in AnalysisModuleKey key, in PythonAnalyzerEntry entry, in ImmutableArray<AnalysisModuleKey> dependencies) {
             _analysisCompleteEvent.Reset();
             ActivityTracker.StartTracking();
             _log?.Log(TraceEventType.Verbose, $"Analysis of {entry.Module.Name}({entry.Module.ModuleType}) queued");
 
-            var graphVersion = _dependencyResolver.ChangeValue(key, entry, entry.IsUserModule, dependencies);
+            var graphVersion = _dependencyResolver.ChangeValue(key, entry, entry.IsUserOrBuiltin || key.IsLibraryAsDocument, dependencies);
```
AlexanderSher: Since library module is added to the graph as a separate entry, it doesn't have any references. So we need to add it as root to trigger analysis.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
```diff
@@ -285,8 +297,14 @@ private class DependencyWalker : PythonWalker {
             }
 
             public override bool Walk(ImportStatement import) {
+                var forceAbsolute = import.ForceAbsolute;
                 foreach (var moduleName in import.Names) {
-                    HandleSearchResults(_pathResolver.FindImports(_module.FilePath, moduleName, import.ForceAbsolute));
+                    var importNames = ImmutableArray<string>.Empty;
```
AlexanderSher: This fixes #1296

src/Analysis/Ast/Impl/Analyzer/Definitions/IPythonAnalyzer.cs
jakebailey: Wouldn't this indicate some other race condition, if the module's Ast is changing during an analysis? When would this occur? When the user modifies their code?

AlexanderSher: Any change to the file, including opening/closing it. I'm not sure that having separate Parse/Analyze queue is still valuable for us, cause we don't have analysis on demand anymore - we use a graph, but that would be a separate change that I don't want to make until database support is in master.

PR #1302, opened 2019-07-09
Adding diagnostic message when using duplicate arguments in function …

CTrando:
…definition.

```
def test(a, a=1, *a, **a):
  pass
```

The above code gives a syntax error. Same as pylint duplicate-argument-name (E0108).

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Callables.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Callables.cs
@@ -16,13 +16,17 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using Microsoft.Python.Analysis.Diagnostics;
 using Microsoft.Python.Analysis.Documents;
 using Microsoft.Python.Analysis.Extensions;
 using Microsoft.Python.Analysis.Modules;
 using Microsoft.Python.Analysis.Types;
 using Microsoft.Python.Analysis.Values;
+using Microsoft.Python.Core;
 using Microsoft.Python.Core.Collections;
+using Microsoft.Python.Parsing;
 using Microsoft.Python.Parsing.Ast;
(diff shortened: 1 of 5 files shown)


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Callables.cs
```diff
@@ -367,6 +374,20 @@ private bool EvaluateFunctionBody(IPythonFunctionType fn)
             return parameters;
         }
 
+        private void ValidateParameters(FunctionDefinition fd) {
+            var duplicates = fd.Parameters.Where(p => !string.IsNullOrEmpty(p.Name)).GroupBy(p => p.Name).Where(g => g.Count() > 1).Select(y => y.Key);
```
CTrando: Could anyone give me a nicer syntax for this? 
I'm trying to find all duplicate names in ```fd.Parameters```

jakebailey: I don't think there's a nicer way off the top of my head. Maybe `fd.Parameters.GroupBy(p => p.Name).SelectMany(g => g.Skip(1)).Distinct()`, but I'm not convinced that's better.

CTrando: yeah and we have to add the string null empty check on top of that too, so maybe this is the best.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Callables.cs
```diff
@@ -367,6 +374,20 @@ private bool EvaluateFunctionBody(IPythonFunctionType fn)
...
+
+            foreach(var paramName in duplicates) {
```
jakebailey: Nit: formatting (`foreach (var ...`)

jakebailey:
Fix the merge conflicts and this is fine.

PR #1305, opened 2019-07-10
No diagnostic message when Generic is called with no args

CTrando:
This is a fix for the issue I talked about in our meeting. The issue was that when in an index expression, if you do:
```
A = [1,2,3]

A[]
```

We would get an error expression for the inside part, but we would fall into the default case in evaluating it and return Unknown type for the arguments when there are actually no arguments and we should (in my opinion) return null.


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
```diff
@@ -125,7 +125,10 @@ internal sealed partial class ExpressionEval {
                 }
             } else {
                 var index = GetValueFromExpression(expr.Index);
-                indices.Add(index ?? UnknownType);
+                // Don't count null indexes as arguments
+                if(index != null) {
```
jakebailey: Nit: formatting.

Issue #1076, opened 2019-05-14
Better package installation watcher


MikhailArkhipov:
Current watcher is fairly unreliable. We need better detection when package is installed so we can either reload modules properly or analyze new module at the time of installation so analysis is ready when user actually imports it.

MikhailArkhipov:
Also, #984

MikhailArkhipov:
https://github.com/acken/FSWatcher

PR #1306, opened 2019-07-10
Reenable file path watcher, watching interpreter paths

### LINKED ISSUE #562: IntelliSense/completions not updating after installing package into current virtual environment
Using Microsoft Python Language Server version 0.1.78.0 with 2019.1.0 of the extension, on Windows 10.

Actual:
- Have virtual environment without requests installed
- Type `import requests`, see warning that module is not found (correct)
- Open terminal, `pip install requests` into virtual environment
- Go back to editor and keep typing, module is still not found
- Reload VS Code, intellisense for requests starts working

Expected: after installing requests, the language server picks up completions for newly installed packages without having to restart VS Code.

I believe this is a regression, as this used to work.

### LINKED ISSUE #1216: Fix path watcher to watch only library paths
A few months ago, I made some changes to clean up the search path logic in the initialization for the LS and in the extension. This had the effect of the extension no longer calling out to python to get the interpreter paths (duplicate work), but the code for the paths watcher relied on that info being there:

```csharp
_pathsWatcher?.Dispose();
_pathsWatcher = new PathsWatcher(
    _initParams.initializationOptions.searchPaths,  // <--
    () =>_server.NotifyPackagesChanged(cancellationToken),
        _services.GetService<ILogger>()
    );

_watchSearchPaths = true;
_searchPaths = _initParams.initializationOptions.searchPaths;
```

So, all the paths watcher does at the moment is watch the user-provided search paths, which can be a bad idea now that `extraPaths` is the recommended way to configure directories in an editable-installed package.

This code should instead be watching the interpreter paths only. However, that is only available later, since that logic is completed in `ReloadModules` (which this code also calls).

Likely needs #1213 to be completed first, but this is a motivating example to move the logic for search path analysis earlier (and just have an event which can be used to re-run it and reload everything).

jakebailey:
Fixes #562.
Fixes #1216.
Updates #1076.

This moves the file path watching code down from `LanguageServer` into `Server`. It then has access to the interpreter's `MainResolution.InterpreterPaths`, and can use that list to watch instead of the (wrong) `searchPath` list from the initialization message.

Two conditions where a paths watcher can be created:

- The paths watcher is created at the first `didConfigurationChange` event. Because the interpreter paths are populated after `ReloadModule` finishes, the interpreter paths are present (as a configuration change can only occur after the initialization is complete).
- If the paths watcher detects a change, it calls `ReloadModule`, then afterwards will check the paths again and recreate itself if the paths have changed.

In both cases, the paths are accessed after `ReloadAsync`, so are in the correct state.

I need to do a bit more testing, but it should be equivalent to the way it was before I changed the `searchPath` generating code in the extension to not send everything at once.

### DIFF
--- a/src/LanguageServer/Impl/Implementation/Server.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.cs
@@ -288,7 +288,11 @@ namespace Microsoft.Python.LanguageServer.Implementation {
             var analyzer = Services.GetService<IPythonAnalyzer>();
             analyzer.ResetAnalyzer();
             foreach (var doc in _rdt.GetDocuments()) {
-                doc.Reset(null);
+                if (!doc.IsOpen) {
+                    _rdt.RemoveDocument(doc.Uri);
+                } else {
+                    doc.Reset(null);
+                }
             }
         }
     }
--- a/src/LanguageServer/Impl/Implementation/Server.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.cs
@@ -287,13 +287,7 @@ namespace Microsoft.Python.LanguageServer.Implementation {
         private void RestartAnalysis() {
             var analyzer = Services.GetService<IPythonAnalyzer>();
             analyzer.ResetAnalyzer();
-            foreach (var doc in _rdt.GetDocuments()) {
-                if (!doc.IsOpen) {
-                    _rdt.RemoveDocument(doc.Uri);
-                } else {
-                    doc.Reset(null);
-                }
-            }
+            _rdt.ReloadAll();
         }
     }
 }
--- a/src/LanguageServer/Impl/Implementation/Server.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.cs
@@ -256,10 +256,6 @@ namespace Microsoft.Python.LanguageServer.Implementation {
         }
 
         private void ResetPathWatcher() {
-            if (!_watchSearchPaths) {
-                return;
-            }
-
             var paths = _interpreter.ModuleResolution.InterpreterPaths.ToArray();
 
             if (_searchPaths == null || !_searchPaths.SequenceEqual(paths)) {
@@ -278,8 +274,12 @@ namespace Microsoft.Python.LanguageServer.Implementation {
             interpreter.ModuleResolution.ReloadAsync(cancellationToken).ContinueWith(t => {
                 _log?.Log(TraceEventType.Information, Resources.Done);
                 _log?.Log(TraceEventType.Information, Resources.AnalysisRestarted);
+
                 RestartAnalysis();
-                ResetPathWatcher();
+                
+                if (_watchSearchPaths) {
+                    ResetPathWatcher();
+                }
             }, cancellationToken).DoNotWait();
 
         }
--- a/src/LanguageServer/Impl/Implementation/Server.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.cs
@@ -115,8 +115,7 @@ namespace Microsoft.Python.LanguageServer.Implementation {
 
             _rootDir = @params.rootUri != null ? @params.rootUri.ToAbsolutePath() : @params.rootPath;
             if (_rootDir != null) {
-                _rootDir = PathUtils.NormalizePath(_rootDir);
-                _rootDir = PathUtils.TrimEndSeparator(_rootDir);
+                _rootDir = PathUtils.NormalizePathAndTrim(_rootDir);
             }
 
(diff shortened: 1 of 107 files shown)


src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -233,7 +238,36 @@ public sealed partial class Server : IDisposable {
         }
         #endregion
 
-        public void NotifyPackagesChanged(CancellationToken cancellationToken) {
+        public void HandlePathWatchChanges(bool watchSearchPaths) {
+            if (!watchSearchPaths) {
```
CTrando: This logic is sort of confusing. Essentially the value of _watchSearchPaths is always the same as the parameter watchSearchPaths?

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -233,7 +238,36 @@ public sealed partial class Server : IDisposable {
...
+            }
+
+            if (!_watchSearchPaths) {
+                _watchSearchPaths = true;
+                ResetPathWatcher();
+            }
+        }
+
+        private void ResetPathWatcher() {
+            if (!_watchSearchPaths) {
```
CTrando: Maybe consider taking this if statement out and put it where ResetPathWatcher is called

src/LanguageServer/Impl/Implementation/Server.cs
CTrando: so you detect when previously we were not watching then we start watching, and when that happens reset the path watcher. Perhaps add a comment to make that more clear, it's hard because the names of the parameter and the field are so similar.

jakebailey: I rewrote it to be more obvious.

jakebailey:
I think this is working. I need to go and remove a bunch of files from the table, as package uninstalls can mean that they no longer exist; keeping them there causes them to be reanalyzed, which causes weird things to happen in the dependency resolution.

@MikhailArkhipov 	I added a removal function to the document table. I can do this another way if it makes sense to be different.

@AlexanderSher 	One downside of removing documents is that if a user has opened a library file, then I won't be able to remove it (otherwise future calls can break, as another didOpen won't be issued), which means that the dependency resolver tries to deal with it and eventually prints something about missing keys. Is this an issue, or is it alright to ignore these? If you uninstall, then reinstall, things still work as expected, it's just slightly noisy.

AlexanderSher:
Have you tried to put breakpoint to the `PythonAnalyzer.RemoveAnalysis` to check if it is called?

jakebailey:
If the file has been opened by the user, then I must assume that it has been modified (I have no way of tracking that it was dirty, unless I'm sorely mistaken), so I do not remove it. This includes goto def to some library where the user leaves it open then removes the package (rare).

jakebailey:
Tested working in Linux after fixing #1324. Moving on to macOS now.

jakebailey:
File watching also works in macOS (though I did find #1326, which explains some of the weird paths I've seen in the past).

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -242,16 +274,25 @@ public sealed partial class Server : IDisposable {
...
         }
 
         private void RestartAnalysis() {
-            var analyzer = Services.GetService<IPythonAnalyzer>();;
+            var analyzer = Services.GetService<IPythonAnalyzer>();
             analyzer.ResetAnalyzer();
             foreach (var doc in _rdt.GetDocuments()) {
-                doc.Reset(null);
+                if (!doc.IsOpen) {
+                    _rdt.RemoveDocument(doc.Uri);
```
MikhailArkhipov: I think you can `UnlockDocument` until count reaches 0 and then close it.

jakebailey: If something has a lock on the document, shouldn't that mean that the other thing is what needs to do the unlocking? I do see `UnlockDocument` called at the beginning of ReloadAsync, though.

jakebailey: Unlocking then closing wouldn't work with the current implementation of CloseDocument, actually. It requires that the document have at least one lock before preceding, and I had hit cases where closing documents put me into the situation where I'd hit the debug assert. Maybe I just allow it to close documents which have not been locked, but I feel like that assert was there for a reason.

jakebailey:
@MikhailArkhipov I moved the reloading code into the RDT itself. This actually avoids any locking/unlocking, since it's purely removing stuff from the dictionary.

PR #1307, opened 2019-07-10
Adding framework for wrong number of arguments, adding check for NewT…

CTrando:
Adding framework for wrong number of arguments, adding check for NewType requiring two arguments.

So when you do:

```
T = NewType('T', int, float)

X = NewType()
```

You will get a diagnostic message.


src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
```diff
@@ -29,5 +29,6 @@ public static class ErrorCodes {
         public const string ReturnInInit = "return-in-init";
         public const string TypingNewTypeArguments = "typing-newtype-arguments";
         public const string TypingGenericArguments = "typing-generic-arguments";
+        public const string WrongNumberArguments = "wrong-number-arguments";
```
jakebailey: I don't really like this naming. I'd feel better if it were `wrong-argument-count` or similar.

CTrando: Sounds good.

src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
```diff
@@ -29,5 +29,6 @@ public static class ErrorCodes {
         public const string ReturnInInit = "return-in-init";
         public const string TypingNewTypeArguments = "typing-newtype-arguments";
         public const string TypingGenericArguments = "typing-generic-arguments";
+        public const string WrongArgumentCount = "wrong-argument-count";
```
jakebailey: On second look, how is this differing from the other two already defined messages, `too-many-function-arguments` and `parameter-missing`?

CTrando: You have a good point. Those two messages are used when building argument sets. What I could do instead of explicitly checking for the number of arguments is once again do the whole 

```
o.setParameters(...)
```

And I believe that if there were not enough arguments we would get them ourselves. What I do not like about those error messages is that they do not say how many arguments needed but I might be able to add that functionality.

CTrando: After looking into more, those errorcodes are used in ArgumentSet DiagnosticEntries, but those entries get put in the Errors list in ArgumentSet, which is not actually shown to the user. 

For that reason I think it's better to make something separate. 

For example:

```
def hello(a, b):
  pass

hello(1,2,1)
```

Will not show an error to the user. We record it internally.

However, for something like:

```
NewType(1)
```
We know for sure this is an error so we can report this.

We have two options 

* Make the Errors field in ArgumentSet accessible to the user (I believe it was not because of low confidence in the argument parsing) and we can use the same errors codes in NewType.

* Make a new error code for functions that we know the definitions of and we are confident in error checking.

Also we should consider what the Errors field in ArgumentSet is even used for. At the moment I don't see where the diagnostic entries in Errors are actually used, only their count.

CTrando:
Leaving this for now until we can resolve what errors in ArgumentSet do.

src/Analysis/Ast/Impl/Types/ArgumentSet.cs
```diff
@@ -178,7 +179,7 @@ internal sealed class ArgumentSet : IArgumentSet {
                     if (formalParamIndex >= overload.Parameters.Count) {
                         // We ran out of formal parameters and yet haven't seen
                         // any sequence or dictionary ones. This looks like an error.
-                        _errors.Add(new DiagnosticsEntry(Resources.Analysis_TooManyFunctionArguments, arg.GetLocation(eval).Span,
+                        _errors.Add(new DiagnosticsEntry(Resources.Analysis_TooManyFunctionArguments, callLocation.Span,
```
CTrando: Added this because currently on too many function arguments we highlight one of the extra args then bail. We should just highlight the whole function instead.

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
```diff
@@ -150,10 +144,24 @@ internal sealed class TypingModule : SpecializedModule {
             _members["Generic"] = new GenericType("Generic", CreateGenericClassParameter, this);
         }
 
-
         private string GetMemberDocumentation(string name)
         => base.GetMember(name)?.GetPythonType()?.Documentation;
 
+        private IPythonType CreateNewType(Location location) {
```
MikhailArkhipov: `SpecializeNewType`

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
```diff
@@ -227,27 +235,32 @@ private IPythonType CreateTupleType(IReadOnlyList<IPythonType> typeArgs)
...
+
+            var args = argSet.Values<IMember>();
+            var name = (args[0] as IPythonConstant)?.GetString();
 
-                var firstArgType = (typeArgs[0] as PythonInstance)?.Type.Name;
-                var eval = args.Eval;
-                var expression = args.Expression;
+            if (!string.IsNullOrEmpty(name)) {
+                return new TypeAlias(name, args[1].GetPythonType() ?? Interpreter.UnknownType);
+            }
```
MikhailArkhipov: need to check `args .Count > 1`

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
```diff
@@ -227,27 +235,32 @@ private IPythonType CreateTupleType(IReadOnlyList<IPythonType> typeArgs)
...
-                    return new TypeAlias(typeName, typeArgs[1].GetPythonType() ?? Interpreter.UnknownType);
-                }
+        private IPythonType CreateTypeAlias(IArgumentSet argSet) {
+            if (argSet.Errors.Count > 0) {
+                argSet.ReportErrors();
+                return Interpreter.UnknownType;
+            }
+
+            var args = argSet.Values<IMember>();
+            var name = (args[0] as IPythonConstant)?.GetString();
```
MikhailArkhipov: need to `check args.Count > 0`? Or never happens?

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
```diff
@@ -227,27 +235,32 @@ private IPythonType CreateTupleType(IReadOnlyList<IPythonType> typeArgs)
...
 
-                var firstArgType = (typeArgs[0] as PythonInstance)?.Type.Name;
-                var eval = args.Eval;
-                var expression = args.Expression;
+            if (!string.IsNullOrEmpty(name)) {
+                return new TypeAlias(name, args[1].GetPythonType() ?? Interpreter.UnknownType);
+            }
 
+            // If first arg is not a string and not unknown, provide diagnostic
+            if (!args[0].IsUnknown() && args[0].GetPythonType() != Interpreter.GetBuiltinType(BuiltinTypeId.Str)) {
```
MikhailArkhipov: `name` will be null (via `(args[0] as IPythonConstant)?.GetString();` and you already checked for non-null above, so it looks like this `if` is redundant.

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
CTrando: So with the argument set change, I specified that the arguments will be name and a list of types, which means the arguments can never be null here.

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
CTrando: It will never happen, see below.

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
CTrando: But the string could be empty, so I include that check specifically for that. `IsNullOrEmpty` also checks for empty string. 

Perhaps now I could change it to a regular null check.

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
MikhailArkhipov: Yeah but what about _count_ of arguments.

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
MikhailArkhipov: Yes, I mean if argument is not a string, then name is null. If name is empty, then it is a string.

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
CTrando: Number of arguments will always be 2. I specified the arguments for the first one to be `name` and the next one to be `type`, which means even if they are not provided by the user, there will still be 2 arguments. Moreover, if the user fails to provide arguments or provides too many, the earlier check with argument errors will return early. But even if we fail to detect weird cases, the number of arguments should be 2.

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
CTrando: ok

src/Analysis/Ast/Impl/Specializations/Typing/TypingModule.cs
MikhailArkhipov: Then you should add argument check on this. Like `Check.Arguments...` since the call here come from public `o.SetReturnValueProvider((interpreter, overload, args) => CreateTypeAlias(args));` and it is hard to tell that some other code somewhere will guarantee 2 args.

CTrando: OK

Issue #1270, opened 2019-07-02
Adding diagnostic error on binary operations with incompatible types


CTrando:

MikhailArkhipov:
https://github.com/microsoft/python-language-server/pull/1254

PR #1308, opened 2019-07-10
Adding diagnostic when doing comparisons with callables

CTrando:
Same as pylint comparison-with-callables. It's a warning that shows up when you do:

```
def test():
  pass

x = 5 < test
```

Doing ```test``` instead of ```test()```

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
@@ -123,9 +123,9 @@ namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
                 ReportDiagnostics(
                     Module.Uri,
                     new DiagnosticsEntry(
-                        Resources.ComparisonWithCallable,
+                        Resources.FunctionComparison,
                         GetLocation(expr).Span,
-                        ErrorCodes.ComparisonWithCallable,
+                        ErrorCodes.FunctionComparison,
                         Severity.Warning,
                         DiagnosticSource.Analysis
                 ));
--- a/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
@@ -119,7 +119,7 @@ namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
                 return Interpreter.GetBuiltinType(leftTypeId);
             }
 
-            if ((left is PythonFunctionType || right is PythonFunctionType) && op.IsComparison()) {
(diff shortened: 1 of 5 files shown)


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Operators.cs
```diff
@@ -119,6 +119,18 @@ internal sealed partial class ExpressionEval {
                 return Interpreter.GetBuiltinType(leftTypeId);
             }
 
+            if ((left is PythonFunctionType || right is PythonFunctionType) && op.IsComparison()) {
```
MikhailArkhipov: Preferably compare to interface rather than class. ie `left is IPythonFuntionType` or you may actually use `MemberType` field instead. Also, what about properties. They are `IPythonPropertyType`.

CTrando: So the reason I put down the concrete class is because I thought there was a distinction between methods and functions when specifying IPythonFunctionType or PythonFunctionType. 

But yes, I added a test for handling class methods and IPythonFunctionType is the way to go, it handles unbound methods just fine.

I tried pylint and it does not handle properties.

src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
```diff
@@ -29,5 +29,5 @@ public static class ErrorCodes {
         public const string ReturnInInit = "return-in-init";
         public const string TypingNewTypeArguments = "typing-newtype-arguments";
         public const string TypingGenericArguments = "typing-generic-arguments";
-    }
-}
+        public const string ComparisonWithCallable = "comparison-with-callable";
+    } }
```
jakebailey: Nit: formatting.

src/Analysis/Ast/Impl/Resources.resx
```diff
@@ -204,4 +204,7 @@
   <data name="GenericNotAllUnique" xml:space="preserve">
     <value>Arguments to Generic must all be unique.</value>
   </data>
-</root>
+  <data name="ComparisonWithCallable" xml:space="preserve">
+    <value>Comparing against a callable, did you omit the parenthesis?</value>
```
jakebailey: This is weird. No other message are phrased like this. Perhaps "Functions are not comparable.".

Note that I would also change this to `FunctionComparison` and `function-comparison`, because something can be "Callable" as well as comparable. "Callable" just means it has `__call__`.

CTrando: Ok, sure sounds good.

src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
CTrando: yup!

CTrando:
Due to the same issues that plagued #1270, going to close this. Same issues are that people can do operator overloading on some functions so we can erroneously give errors.

Issue #1291, opened 2019-07-09
Stuck at analyzing in background and causing high cpu load


kaushall:
### Environment data

- VS Code version: 1.36.0 (Stable Channel)
- Extension version (ms-python.python): 2019.6.22090
- OS and version: Ubuntu 18.04 LTS
- Python version (& distribution if applicable, e.g. Anaconda): 2.7.15
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): virtualenv

============================================================
### Logs

User belongs to experiment group 'AlwaysDisplayTestExplorer - control'
Starting Microsoft Python language server.
[Info  - 10:33:32 AM] Analysis cache path: /home/kaushal/.cache/Microsoft/Python Language Server
[Info  - 10:33:32 AM] GetCurrentSearchPaths /home/kaushal/microfinance/venv/bin/python 
[Info  - 10:33:33 AM] Interpreter search paths:
[Info  - 10:33:33 AM]     /home/kaushal/microfinance/venv/lib/python2.7
[Info  - 10:33:33 AM]     /home/kaushal/microfinance/venv/lib/python2.7/lib-dynload
[Info  - 10:33:33 AM]     /usr/lib/python2.7
[Info  - 10:33:33 AM]     /usr/lib/python2.7/plat-x86_64-linux-gnu
[Info  - 10:33:33 AM]     /usr/lib/python2.7/lib-tk
[Info  - 10:33:33 AM]     /home/kaushal/microfinance/venv/local/lib/python2.7/site-packages
[Info  - 10:33:33 AM]     /home/kaushal/microfinance/venv/lib/python2.7/site-packages
[Info  - 10:33:33 AM] User search paths:
[Info  - 10:33:35 AM] Microsoft Python Language Server version 0.3.21.0
[Info  - 10:33:35 AM] Initializing for /home/kaushal/microfinance/venv/bin/python
.
.
.
Analysis of pandas.core.dtypes.dtypes(Library) completed in 5.9831 ms.
Analysis of pandas.core.dtypes.api(Library) completed in 0.7383 ms.
Analysis of pandas.io.json.table_schema(Library) completed in 1.6775 ms.
Analysis of pandas.core.indexing(Library) completed in 32.862 ms.
Analysis of pandas.util.testing(Library) completed in 49.5013 ms.
Analysis of pandas.core.sorting(Library) completed in 12.6855 ms.
Analysis of pandas.plotting._misc(Library) completed in 12.8782 ms.
[...]

PR #1309, opened 2019-07-11
Optimize GetMemberNames and SetClassMembers

jakebailey:
For #1291.

Not sure if this is safe, but makes the example given in #1291 complete in a reasonable amount of time. Running tests now.


jakebailey:
Not safe as-is. Failing the GenericClass tests.

jakebailey:
These optimizations are probably good in general (with code to collect lists before outputting them), but aren't a fix to #1291. Closing.

PR #1328, opened 2019-07-12
Check content for null in ScanClosedFiles

### LINKED ISSUE #1317: lsp-request: Object reference not set to an instance of an object with lsp-ui-peek-find-references
# Issue
When running ```lsp-ui-peek-find-reference``` on any token in a source file I get above error.

# Environment
- Spacemacs: 0.300.0@26.2
- emacs:  GNU Emacs 26.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.8) of 2019-04-12
- This repo:  commit 4a261428ac3f6eadea3b208c91ae8e1b612a838c  
- python runtime: 3.6.7 64-bit

jakebailey:
Fixes #1317.

### DIFF
--- a/src/LanguageServer/Impl/Sources/ReferenceSource.cs
+++ b/src/LanguageServer/Impl/Sources/ReferenceSource.cs
@@ -121,6 +121,10 @@ namespace Microsoft.Python.LanguageServer.Sources {
                     }
 
                     var content = fs.ReadTextWithRetry(filePath);
+                    if (content == null) {
+                        continue;
+                    }
+
                     if (content.Contains(name)) {
                         files.Add(uri);
                     }


yyoncho:
This solution is problematic - imagine that there are 100 files that trigger this issue - it will take 100 seconds to return the references. I will come back with the root cause of this issue, I am having a hard time to reproduce the issue because I deleted the original set of backup files.

MikhailArkhipov:
100 files scan is more like couple of seconds. Only candidates that possibly contain symbol are parsed and parsing is ~20ms per fairly large file.

yyoncho:
> 100 files scan is more like couple of seconds. Only candidates that possibly contain symbol are parsed and parsing is ~20ms per fairly large file.

No, take a look at `ReadTextWithRetry`. It returns null after retrying for a second.

MikhailArkhipov:
Only when it fails. Retries are rare, they are mostly caused by something else reading or writing the file. Such as AV software, another instance of VSC or VS, that king of thing. Consistent failure is a different issue.

yyoncho:
> Only when it fails. Retries are rare, they are mostly caused by something else reading or writing the file. Such as AV software, another instance of VSC or VS, that kind of thing. Consistent failure is a different issue.

When the issue was first reported I asked the OP to test the same fix as the one in this PR and he got a timeout. I have identified 2(probably there are more) cases in which the language server will wait for 1 second every time.

1. FOO.py symlink that points to non-existing file.
2. If the user does not have permission to access FOO.py 

IMO ReadTextWithRetry should not retry on any IO exception, what do you think?

jakebailey:
There can be temporary errors that throw IO exceptions, as mentioned before. That's what the retry loop is there to try and stem. Those two cases may be individually identifiable, however.

Your original issue mentioned temp files made by emacs and suggested not trying to open them at all; If you have the names of the files that shouldn't be opened, it may be best to start by not trying to open them in the first place.

MikhailArkhipov:
@yyoncho _'IMO ReadTextWithRetry should not retry on any IO exception, what do you think?'_

Yes it should. The code is there for a reason, it actually came from earlier versions and from VS and was added as a result of issues reported. There are many cases when there may be temporary IO issue, like AV scan, other app, user opening file in another app, cache writing in one instance on VSC or VS while reading cache in another instance, momentary network issues with mapped drives and UNC locations, and so on and so forth. Ex we do share generated code stubs across running instances.

yyoncho:
@jakebailey 
> Your original issue mentioned temp files made by emacs and suggested not trying to open them at all; 

FTR I am not the original author of the issue, I am the maintainer of the Emacs Language Server Protocol client (lsp-mode).

> If you have the names of the files that shouldn't be opened, it may be best to start by not trying to open them in the first place.

They are like that `.#<original-file-name>.py`.

@MikhailArkhipov 

> There are many cases when there may be temporary IO issue, like AV scan, other app, user opening file in another app, cache writing in one instance on VSC or VS while reading cache in another instance, momentary network issues with mapped drives and UNC locations, and so on and so forth. 

This doesn't mean that you should retry on **any** IOException, right? (e. g. there is no point to retry on System.IO.FileNotFoundException exception although it is IOException).

vv111y:
Apologies from being away from this. 
When I try to find references for token under cursor The delay is 10.4 seconds, not 1s.   
The response is:  
`lsp-request: Timeout while waiting for response. Method: textDocument/references.`

I rebuilt mspyls from latest 10 minutes ago. I am unfamiliar with how mspyls works. What is needed to debug this? Thanks.

MikhailArkhipov:
Depends on the OS you are using. Debugging steps are in https://github.com/microsoft/python-language-server/blob/master/CONTRIBUTING.md

PR #1334, opened 2019-07-15
Treat persistent module as regular specialized

MikhailArkhipov:
- Eliminates errors like `missing key sys()` in the output.

PR also includes merge of master and re-enable storing/retrieving of persistent data.


src/LanguageServer/Impl/Program.cs
```diff
@@ -13,7 +13,7 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
-// #define WAIT_FOR_DEBUGGER
+#define WAIT_FOR_DEBUGGER
```
AlexanderSher: Comment

MikhailArkhipov: Removed

PR #1337, opened 2019-07-15
Give diagnostic when inheriting from something that is not a class

### LINKED ISSUE #1278: When class inherits from something that is not a class, give a diagnostic error

CTrando:
Fixes #1278 

Try number 2. 

Test Plan:

**Diagnostic**

* Inheriting a variable that is a constant 

* Inheriting a constant

* Inheriting a function

* Inheriting a method

* Inheriting an object instance

* Inheriting a property that returns a constant

* Inheriting a property that returns a function

* Inheriting a property that returns a method

**No Diagnostic**

* Builltin types

* Renamed builin types

* User Class

* A property that returns nothing


src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -118,6 +111,40 @@ internal sealed class ClassEvaluator : MemberEvaluator {
...
+                var expr = a.Expression;
+                var m = Eval.GetValueFromExpression(expr);
+
+                switch (m?.MemberType) {
+                    case PythonMemberType.Method:
+                    case PythonMemberType.Function:
+                    case PythonMemberType.Property:
+                    case PythonMemberType.Instance:
+                    case PythonMemberType.Variable when m is IPythonConstant:
+                        // all invalid types to inherit from
```
MikhailArkhipov: I think you can derive from instance if the instance is a type info. I.e.
```python
class A: ...
a = A
class B(a): ...
```
appears to work
image

so we need to test that

CTrando: Will check.

CTrando: Works fine. The reason is that I resolve the name expression 'a' to a member whose type is Class.

jakebailey:
A question, what would the diagnostics be for something like this?

```python
X = 123

class Foo(X):
   pass

class Bar(Foo):
   pass
```

Is it going to complain at both `Foo` and `Bar`, or just `Foo`? I'm thinking about the case where we get this wrong in one place, and it infects an inheritance tree. (Particularly for classes that come from mis-analyzed libraries or similar.)

CTrando:
> A question, what would the diagnostics be for something like this?
> 
> ```python
> X = 123
> 
> class Foo(X):
>    pass
> 
> class Bar(Foo):
>    pass
> ```
> 
> Is it going to complain at both `Foo` and `Bar`, or just `Foo`? I'm thinking about the case where we get this wrong in one place, and it infects an inheritance tree. (Particularly for classes that come from mis-analyzed libraries or similar.)

Just tested it. It only complains at Foo. It will only complain about a direct inheritance, not a chain.

jakebailey:
Perfect, but would you mind adding a test for it so we don't regress later on? (And fix the merge conflicts). Then this should be good.

CTrando:
Sorry, just haven't pushed test yet.

PR #1342, opened 2019-07-16
Check analysis dependencies for null

### LINKED ISSUE #1341: Argument null exception in PythonAnalyzerEntry.Invalidate
`UpdateAnalysisTcs` can set `_analysisDependencies` to null. In one `Invalidate` the null check is like
```cs
    UpdateAnalysisTcs(analysisVersion);
    dependencies = _analysisDependencies != null ? ...
```
but in another overload it is
```cs
    UpdateAnalysisTcs(analysisVersion);
    dependencies = _parserDependencies != null ? ...
```
which causes argument exception in `_parserDependencies.Union(_analysisDependencies)`

MikhailArkhipov:
Fixes #1341

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
@@ -221,7 +221,6 @@ namespace Microsoft.Python.Analysis.Analyzer {
                 dependencies = _parserDependencies != null
                     ? ImmutableArray<AnalysisModuleKey>.Create(_parserDependencies.Union(_analysisDependencies).ToArray())
                     : ImmutableArray<AnalysisModuleKey>.Create(_analysisDependencies);
-
                 return true;
             }
         }
--- a/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
@@ -77,7 +77,7 @@ namespace Microsoft.Python.Analysis.Analyzer {
                 }
             }
         }
-        
+
         public int BufferVersion => _bufferVersion;
 
         public int AnalysisVersion {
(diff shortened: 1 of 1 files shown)


src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
```diff
@@ -219,9 +218,14 @@ public Task<IDocumentAnalysis> GetAnalysisAsync(CancellationToken cancellationTo
...
 
                 UpdateAnalysisTcs(analysisVersion);
-                dependencies = _parserDependencies != null
-                    ? ImmutableArray<AnalysisModuleKey>.Create(_parserDependencies.Union(_analysisDependencies).ToArray())
-                    : ImmutableArray<AnalysisModuleKey>.Create(_analysisDependencies);
+                if (_analysisDependencies != null) {
+                    dependencies = _parserDependencies != null
+                        ? ImmutableArray<AnalysisModuleKey>.Create(_parserDependencies.Union(_analysisDependencies).ToArray())
+                        : ImmutableArray<AnalysisModuleKey>.Create(_analysisDependencies);
+                } else if(_parserDependencies != null) {
```
jakebailey: Nit: formatting

MikhailArkhipov: whats wrong with it?

jakebailey: `if(` vs `if (`. Minor (but VS normally fixes it).

AlexanderSher:
I think it would be better to remove `_analysisDependencies = null` from `UpdateAnalysisTcs`, cause it is already set to `null` in `TrySetAnalysis`.

MikhailArkhipov:
Should we then assume non-null in 
```cs
public bool Invalidate(IPythonModule module, PythonAst ast, int bufferVersion, int analysisVersion, out ImmutableArray<AnalysisModuleKey> dependencies)
```
and just do
```
dependencies = ImmutableArray<AnalysisModuleKey>.Create(_parserDependencies.Union(_analysisDependencies).ToArray())
```
?

AlexanderSher:
No, this is true only for library analysis. User code analysis erases all dependencies cause changed code may not contain both parsing-created dependencies and analysis-created dependencies.

PR #1343, opened 2019-07-16
Implement goto declaration

### LINKED ISSUE #494: Add support for go-to declaration
https://code.visualstudio.com/updates/v1_30#_go-to-declaration

### LINKED ISSUE #1138: Go to definition behavior change request
Currently, using "Go to definition" with the MS Language Server will jump you to your import statement, likely at the top of your file, rather than jumping you into the class. I understand from a separate issue https://github.com/microsoft/python-language-server/issues/850 that this is the intended behavior. Below is a gif of it.

58559204-dd9c0780-8222-11e9-9bed-cebbfa5b5504

In my view, "Go to definition" means that you go to where the class is defined. Taking me to the top of the file and showing me the import statement doesn't seem nearly as widely used or necessary as jumping directly to where the class is implemented. 

Imagine you were in the full version of Visual Studio coding in C#, "Go to definition" would take you to the definition of your class. 

If we insist that we need a separate feature to navigate us to the import statement, it seems that we should make a separate button and keyboard shortcut called "Go to declaration", and leave the "Go to definition" function to be the same behavior as Jedi.

MikhailArkhipov:
Closes #494
Closes #1138
And change behavior of goto definition to go to the root.

### DIFF
--- a/src/Analysis/Ast/Impl/Extensions/MemberExtensions.cs
+++ b/src/Analysis/Ast/Impl/Extensions/MemberExtensions.cs
@@ -86,5 +86,21 @@ namespace Microsoft.Python.Analysis {
             }
             return null;
         }
+
+        public static ILocatedMember GetRootDefinition(this ILocatedMember lm) {
+            if (!(lm is IImportedMember im) || im.Parent == null) {
+                return lm;
+            }
+
+            var parent = im.Parent;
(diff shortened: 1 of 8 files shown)


amittleider:
Thank you! This is big!

Jma353:
+100

PR #1347, opened 2019-07-18
Fixing issue with decorators with same prefix seen as redefinition

CTrando:
Followed pylint's guide more closely.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
@@ -241,21 +241,19 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
         private bool IsRedefinedByDecorator(FunctionDefinition fd) {
             var decorators = fd.Decorators?.Decorators ?? Array.Empty<Expression>();
             foreach (var m in decorators.OfType<MemberExpression>()) {
-                // If the decorator prefix has the same name as the function, then the function is redefined by decorator
+                // If a decorator member expression's target has the same name as the function, then the function is redefined via decorator
                 // e.g
-                // @property
-                // def x(self): ...
-                // 
                 // @x.setter
                 // def x(self): ...
+                // x.setter has the target x and the function is named x, so x is redefined via decorator
                 //
                 // But pylint only compares one level member expressions for decorators
                 // e.g 
                 // @foo.a
-                // def foo(self): ...
-                // is valid but not
+                // def foo(self): ... is a redefinition but 
+                //
                 // @foo.a.b
-                // def foo(self): ... 
+                // def foo(self): ... is not.
                 var name = (m.Target as NameExpression)?.Name ?? string.Empty;
                 if (name.Equals(fd.Name)) {
                     return true;
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
@@ -101,17 +101,8 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
             var declaringType = fd.Parent != null && _typeMap.TryGetValue(fd.Parent, out var t) ? t : null;
             var existing = _eval.LookupNameInScopes(fd.Name, LookupOptions.Local);
 
-            // If function redefined, only output diagnostic if function name
-            // is not redefined via decorator
-            // e.g
-            // 
-            // @property
-            // def x(self): ...
-            // 
-            // @x.setter
-            // def x(self): ...
-            // Should not give any diagnostics because x is redefined by a valid decorator 
-            if (!IsRedefinedByDecorator(fd)) {
+            // lambdas have the same function name, so don't count them as redefinitions
+            if (!IsRedefinedByDecorator(fd) && !string.IsNullOrEmpty(fd.Name) && !fd.IsLambda) {
                 switch (existing?.MemberType) {
                     case PythonMemberType.Method:
                     case PythonMemberType.Function:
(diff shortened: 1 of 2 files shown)


src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
```diff
@@ -229,6 +242,39 @@ internal sealed class SymbolCollector : PythonWalker {
                     DiagnosticSource.Analysis));
         }
 
+        /// <summary>
+        ///
```
MikhailArkhipov: Empty doc?

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
```diff
@@ -229,6 +242,39 @@ internal sealed class SymbolCollector : PythonWalker {
...
+        /// 
+        /// </summary>
+        /// <param name="fd"></param>
+        /// <returns></returns>
+        private bool IsRedefinedByDecorator(FunctionDefinition fd) {
+            if (fd.Decorators != null) {
+                foreach (var d in fd.Decorators.Decorators) {
+                    switch (d) {
+                        case NameExpression n:
+                            if (fd.Name.Equals(n.Name)) {
```
MikhailArkhipov: I think `fd.Name` can be null in `def (a, b, c)`

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
```diff
@@ -229,6 +242,39 @@ internal sealed class SymbolCollector : PythonWalker {
...
+                            // pylint only compares one level member expressions for decorators
+                            // e.g 
+                            // @foo.a
+                            // def foo(self): ...
+                            // is valid but not
+                            // @foo.a.b
+                            // def foo(self): ... 
+                            var ne = m.Target as NameExpression;
+                            if (fd.Name.Equals(ne?.Name ?? string.Empty)) {
+                                return true;
```
MikhailArkhipov: You do can compare to null, `Equals` won't crash. But `fd.Name` can be null, see above

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
```diff
@@ -243,34 +242,28 @@ internal sealed class SymbolCollector : PythonWalker {
...
+                // If the decorator prefix has the same name as the function, then the function is redefined by decorator
+                // But pylint only compares one level member expressions for decorators
+                // e.g 
+                // @foo.a
+                // def foo(self): ...
+                // is valid but not
+                // @foo.a.b
+                // def foo(self): ... 
+                var name = (m.Target as NameExpression)?.Name ?? string.Empty;
+                if (fd.Name.Equals(name)) {
```
MikhailArkhipov: fd.Name can be null

MikhailArkhipov: Use `fd.Name == name` or, since name is not null, do `name.Equals(fd.Name)`

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
CTrando: I checked this, doesn't seem to be the case.

CTrando: I'll leave a check in there however.

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
```diff
@@ -101,17 +101,8 @@ internal sealed class SymbolCollector : PythonWalker {
...
-            // 
-            // @property
-            // def x(self): ...
-            // 
-            // @x.setter
-            // def x(self): ...
-            // Should not give any diagnostics because x is redefined by a valid decorator 
-            if (!IsRedefinedByDecorator(fd)) {
+            // lambdas have the same function name, so don't count them as redefinitions
+            if (!IsRedefinedByDecorator(fd) && !string.IsNullOrEmpty(fd.Name) && !fd.IsLambda) {
```
jakebailey: Check IsLambda and empty name first; IsRedefinedByDecorator is expensive.

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
```diff
@@ -101,12 +101,15 @@ internal sealed class SymbolCollector : PythonWalker {
...
             var existing = _eval.LookupNameInScopes(fd.Name, LookupOptions.Local);
 
-            switch (existing?.MemberType) {
-                case PythonMemberType.Method:
-                case PythonMemberType.Function:
-                case PythonMemberType.Property:
-                    ReportRedefinedFunction(fd, existing as PythonType);
-                    break;
+            // lambdas have the same function name, so don't count them as redefinitions
+            if (!string.IsNullOrEmpty(fd.Name) && !fd.IsLambda && !IsRedefinedByDecorator(fd)) {
```
jakebailey: On a second reading, I think it'd be best to do this:

```
if (existing != null
    && !fd.IsLambda
    && !string.IsNullOrEmpty(fd.Name)
    && !IsRedefinedByDecorator(fd)) {
    // ...
}
```

As for the most part, there aren't going to be any redefinitions, and `existing` will be null. As written now with the switch case's `existing?` check, the null check happens last, even though it's the most likely.

CTrando: Good insight, I'll add it.

jakebailey:
Closing; see #1359.

PR #1359, opened 2019-07-22
Revert redefined function diagnostic

### LINKED ISSUE #1346: False positive redefinition warning
The new redefinition logic from (I'm guessing) #1273 looks to have false positives for property attributes in classes. For example, from the [Python docs](https://docs.python.org/3.6/library/functions.html#property):

```python
class C:
    def __init__(self):
        self._x = None

    @property
    def x(self):
        return self._x

    @x.setter
    def x(self, value):
        self._x = value
```

This has a `function-redefined` warning on the second `def x`, but should not. It is valid code.

I assume it's the same issue, but this also impacts SQLAlchemy `hybrid_property`s:
```python
from sqlalchemy.ext.hybrid import hybrid_property

class Interval(object):
    @hybrid_property
    def radius(self):
        return 'foo'

    @radius.expression
    def radius(cls):
        return 'bar'
```

### LINKED ISSUE #1358: function-redefined raised on conditionally-defined code
E.g.
```python
if __debug__:
    def spam():
        return 42
else:
    def spam():  # Flagged as a redefinition.
        return -13
```
This is also a very common pattern when you use different code based on OS, whether a module is available, etc.

jakebailey:
Closes #1346.
Closes #1358.

This is too risky for now; without dataflow analysis and handling of odd cases like in https://github.com/microsoft/python-language-server/issues/1358#issuecomment-513922071 this is probably too much to be there by default. Manual revert of #1273 (auto-revert had conflicts).


src/Analysis/Ast/Impl/Diagnostics/ErrorCodes.cs
```diff
@@ -25,8 +25,6 @@ public static class ErrorCodes {
         public const string UndefinedVariable = "undefined-variable";
         public const string VariableNotDefinedGlobally= "variable-not-defined-globally";
         public const string VariableNotDefinedNonLocal = "variable-not-defined-nonlocal";
-        public const string FunctionRedefined = "function-redefined";
-        public const string UnsupportedOperandType = "unsupported-operand-type";
```
jakebailey: This was in #1273 but was a merge error there; removing it now too.

PR #1368, opened 2019-07-23
Not giving diagnostic message on unknown types being bases

### LINKED ISSUE #1361: False positives for "inheriting from something that is not a class"
The Diagnostics for Inheriting from something that is not a class reports false positives when a library like [PonyORM](https://github.com/ponyorm/pony) creates classes dynamically.

MWE:
```python
from pony.orm import *
db = Database()
# You're supposed to have your entities inherit from db.Entity
# For reference, this gets created in pony/orm/core.py:746
class MyEntity(db.Entity): # Diagnostic warning because the python language server thinks we're not inheriting from a class
    mystr = Optional(str)

db.bind(provider='sqlite', filename='database.sqlite', create_db=True, create_tables=True)
```

CTrando:
Fixes #1361

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -120,6 +120,8 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
                 var m = Eval.GetValueFromExpression(expr);
 
                 switch (m) {
+                    // Allow unknown members 
+                    case IMember member when member.GetPythonType() == Eval.UnknownType:
                     // Allow any members from typing module
                     // TODO handle typing module specialization better: https://github.com/microsoft/python-language-server/issues/1367
                     case ILocatedMember l when l.DeclaringModule is TypingModule:
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -16,6 +16,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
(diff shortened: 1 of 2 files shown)


src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -120,6 +120,8 @@ internal sealed class ClassEvaluator : MemberEvaluator {
                 var m = Eval.GetValueFromExpression(expr);
 
                 switch (m) {
+                    // Allow unknown members 
+                    case IMember member when member.GetPythonType() == Eval.UnknownType:
```
jakebailey: Why `when member.GetPythonType() == Eval.UnknownType` over `when member.IsUnknown()`?

MikhailArkhipov:
```cs
if(member.IsUnknown()) {
    return ...
}
```

jakebailey: In fact, `m` is already an `IMember`, so you can just do `if (m.IsUnknown())` before the `switch`.

CTrando: Ah man, I forgot about that extension method.

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -16,6 +16,7 @@
 using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.CompilerServices;
```
jakebailey: Remove?

jakebailey: (A few usings are unused, the document could use an autoformat)

CTrando: Ok.

PR #1372, opened 2019-07-23
Multiple fixes and additional tests for persistence

### LINKED ISSUE #1217: Additional tests for module restoration from db
- Tests with generics
- Tests with nested classes
- Tests with functions returning functions
- Deeper comparison of real and restore analysis (drill into members)?
- ...

### LINKED ISSUE #1222: Handle types from submodules when restoring from db
I.e. cases like
```json
    {
      "Value": "requests.models:Request",
      "Name": "Request"
    },
```

### LINKED ISSUE #1340: Provide single source for interpreter configuration information
Currently `InterpreterConfiguration` is fully populated in tests while being almost empty in live product since paths are calculated. Thus some code in `PythonLibraryPath` only works in tests.

Persistent storage needs to know module location in order to calculate unique names with versions.  We should have something like `IPythonConfigurationService` which then can be used to obtain actual data. 

We can either completely eliminate `InterpreterConfiguration` and have everything rely on the interface and mock it in tests or remove some unused fields from the struct and have new service return `InterpreterConfiguration` properly populated.

The service can also handle changes to the configuration after `pip install` or user updates to the client app settings.

MikhailArkhipov:
Closes #1217
Closes #1222 
Closes #1340
Includes https://github.com/microsoft/python-language-server/pull/1339

- Improve qualified name construction
- Implement separate qualified name for typing classes
- Fix issues with merging stubs (do not replace classes, do not replace `__class__`, only replace matching types).
- Fix case when standalone function could become an unbound method
- Handle case when there is no module, but there is stub (`_importlib_modulespec`)
- Make `typing:AnyStr` to behave like a string and not a generic type so it is simpler to persist/restore.
- Handle cases when type came from stub and primary module is not loaded (case of `mmap` in `io`).

To be continued in https://github.com/microsoft/python-language-server/issues/1373


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.cs
```diff
@@ -238,29 +238,31 @@ private IMember GetValueFromFString(FString fString)
...
-
+            var type = m.GetPythonType();
             var value = type?.GetMember(expr.Name);
             type?.AddMemberReference(expr.Name, this, GetLocationOfName(expr));
 
             if (type is IPythonModule) {
                 return value;
             }
 
+            IPythonInstance instance = null;
```
MikhailArkhipov: Change to fix the issue when standalone function could become an unbound method with declaring type of null. Does not really affect completions, but present problems when restoring function from the database.

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -242,57 +242,103 @@ public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst
...
-                            continue; // Leave builtins alone.
-                        }
-                        if (!IsStubBetterType(memberType, stubMemberType)) {
-                            continue;
+                // If type does not exist in module, but exists in stub, declare it unless it is an import.
+                // If types are the classes, merge members. Otherwise, replace type from one from the stub.
+                switch (sourceType) {
+                    case null:
+                        if (v.Source == VariableSource.Declaration) {
+                            Eval.DeclareVariable(v.Name, v.Value, v.Source);
```
MikhailArkhipov: Bug was that we did not transfer variables from stub to the primary if primary did not have same variable declared. This caused problems restoring types since said variable could declare a class that the type uses in say, annotations.

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -242,57 +242,103 @@ public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst
...
-                    // should remain library with its own stub attached.
-                    if (!(stubType is IPythonModule) && !builtins.Equals(stubType.DeclaringModule)) {
-                        sourceType.TransferDocumentationAndLocation(stubType);
-                        // TODO: choose best type between the scrape and the stub. Stub probably should always win.
-                        var source = Eval.CurrentScope.Variables[v.Name]?.Source ?? VariableSource.Declaration;
-                        Eval.DeclareVariable(v.Name, v.Value, source);
-                    }
+                        break;
+
+                    case PythonClassType cls when Module.Equals(cls.DeclaringModule):
```
MikhailArkhipov: Just changing if to a switch

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -242,57 +242,103 @@ public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst
...
         private static bool IsStubBetterType(IPythonType sourceType, IPythonType stubType) {
-            // If stub says 'Any' but we have better type, keep the current type.
             if (stubType.IsUnknown()) {
+                // Do not use worse types than what is in the module.
+                return false;
+            }
+            if (sourceType.IsUnknown()) {
+                return true; // Anything is better than unknowns.
+            }
+            if (sourceType.MemberType != stubType.MemberType) {
```
MikhailArkhipov: Occasionally stub can have type of the same name but completely different type (like class vs function). We don't want to stop on the primary definitions.

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -242,57 +242,103 @@ public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst
...
+                // Types should match, we are not replacing unrelated types.
                 return false;
             }
-            return sourceType.IsUnknown() || !(stubType.DeclaringModule is TypingModule) || stubType.Name != "Any";
+            // If stub says 'Any' but we have better type, keep the current type.
+            return !(stubType.DeclaringModule is TypingModule) || stubType.Name != "Any";
+        }
+
+        private static void TransferDocumentationAndLocation(IPythonType s, IPythonType d) {
+            if (s.IsUnknown()) {
```
MikhailArkhipov: Moved code - it was an extension method that was only used here.

src/Analysis/Ast/Impl/Extensions/ArgumentSetExtensions.cs
```diff
@@ -58,7 +58,7 @@ public static IReadOnlyList<T> Values<T>(this IArgumentSet args)
             }
 
             if (args.ListArgument != null && !string.IsNullOrEmpty(args.ListArgument.Name)) {
-                var type = new PythonCollectionType(null, BuiltinTypeId.List, eval.BuiltinsModule, false);
+                var type = new PythonCollectionType(BuiltinTypeId.List, eval.BuiltinsModule, false);
```
MikhailArkhipov: Removing parameter that was passed as null everywhere

src/Analysis/Ast/Impl/Extensions/PythonFunctionExtensions.cs
```diff
@@ -37,6 +38,10 @@ public static bool HasClassFirstArgument(this IPythonClassMember m)
         }
 
         public static string GetQualifiedName(this IPythonClassMember cm) {
+            if(cm.DeclaringModule.ModuleType == ModuleType.Builtins) {
```
MikhailArkhipov: Per `TypeNames` builtins are not prefixed with the module name.

src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -164,8 +164,17 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
                     if (valueType is PythonModule) {
                         return false; // Do not re-export modules.
                     }
-                    // Do not re-export types from typing
-                    return !(valueType?.DeclaringModule is TypingModule) || this is TypingModule;
+                    if (this is TypingModule) {
+                        return true; // Let typing module behave normally.
+                    }
+                    // Do not re-export types from typing. However, do export variables
```
MikhailArkhipov: Fixes the issue when we did not export variables that were not types from `typing` but rather assignments like `x = Union[int, str]`, thinking that `x` is a type from `typing`.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericType.cs
```diff
@@ -119,5 +120,8 @@ public IPythonType CreateSpecificType(IArgumentSet typeArguments)
 
         public override int GetHashCode()
             => TypeId != BuiltinTypeId.Unknown ? TypeId.GetHashCode() : base.GetHashCode();
+
+        private IReadOnlyList<IPythonType> GetTypesFromValues(IEnumerable<IArgument> args)
```
MikhailArkhipov: Arguments can be types of instances, so direct cast to types was not correct.

src/Analysis/Ast/Impl/Specializations/Typing/Types/NamedTupleType.cs
```diff
@@ -27,20 +27,25 @@ internal sealed class NamedTupleType : TypingTupleType, ITypingNamedTupleType {
         /// <summary>
         /// Creates type info for a strongly-typed tuple, such as Tuple[T1, T2, ...].
         /// </summary>
-        public NamedTupleType(string tupleName, IReadOnlyList<string> itemNames, IReadOnlyList<IPythonType> itemTypes, IPythonInterpreter interpreter)
-            : base(itemTypes, interpreter) {
+        public NamedTupleType(string tupleName, IReadOnlyList<string> itemNames, IReadOnlyList<IPythonType> itemTypes, IPythonModule declaringModule, IPythonInterpreter interpreter)
```
MikhailArkhipov: When type is declared via `NamedTuple`, if should belong to the module it is declared in and not to `typing`, similar to how other generics behave when specific type is created.

src/Analysis/Ast/Impl/Specializations/Typing/Types/NamedTupleType.cs
```diff
@@ -27,20 +27,25 @@ internal sealed class NamedTupleType : TypingTupleType, ITypingNamedTupleType {
...
             ItemNames = itemNames;
 
             var typeNames = itemTypes.Select(t => t.IsUnknown() ? string.Empty : t.Name);
             var pairs = itemNames.Zip(typeNames, (name, typeName) => string.IsNullOrEmpty(typeName) ? name : $"{name}: {typeName}");
             Name = CodeFormatter.FormatSequence(tupleName, '(', pairs);
+
+            typeNames = itemTypes.Select(t => t.IsUnknown() ? string.Empty : t.QualifiedName);
+            pairs = itemNames.Zip(typeNames, (name, typeName) => string.IsNullOrEmpty(typeName) ? name : $"{name}: {typeName}");
+            QualifiedName = CodeFormatter.FormatSequence($"{declaringModule}:{tupleName}", '(', pairs);
         }
```
MikhailArkhipov: Adding special calculation of qualified name to all specialized types from `typing`

src/Analysis/Ast/Impl/Specializations/Typing/Types/TypingTupleType.cs
```diff
@@ -28,16 +29,20 @@ internal class TypingTupleType : PythonCollectionType, ITypingTupleType {
...
         /// </summary>
         /// <param name="itemTypes">Tuple item types.</param>
+        /// <param name="declaringModule">Declaring module. If null, then 'typing' is assumed.</param>
         /// <param name="interpreter">Python interpreter.</param>
-        public TypingTupleType(IReadOnlyList<IPythonType> itemTypes, IPythonInterpreter interpreter)
-            : base(null, BuiltinTypeId.Tuple, interpreter.ModuleResolution.GetSpecializedModule("typing"), false) {
-            ItemTypes = itemTypes;
+        public TypingTupleType(IReadOnlyList<IPythonType> itemTypes, IPythonModule declaringModule, IPythonInterpreter interpreter)
+            : base(BuiltinTypeId.Tuple, declaringModule ?? interpreter.ModuleResolution.GetSpecializedModule("typing"), false) {
+            ItemTypes = itemTypes.Count > 0 ? itemTypes : new[] { interpreter.UnknownType };
```
MikhailArkhipov: Occasionally during restore we may get `Tuple[]`. This code works around this. I will continue looking into why thing might be happening.

src/Analysis/Ast/Impl/Types/PythonTypeWrapper.cs
```diff
@@ -32,7 +34,11 @@ protected IPythonType InnerType
         /// Creates delegate type wrapper over an existing type.
         /// Use dedicated constructor for wrapping builtin types.
         /// </summary>
-        public PythonTypeWrapper(IPythonType type) : this(type, type.DeclaringModule) {
+        public PythonTypeWrapper(IPythonType type) : this(type, type.DeclaringModule) { }
+
+        public PythonTypeWrapper(string typeName, string documentation, IPythonModule declaringModule, IPythonType baseType) : this(baseType, declaringModule) {
```
MikhailArkhipov: Allow override name and documentation in the wrapper. Comes handy in some type restoration cases, such as with named tuples.

src/Analysis/Ast/Test/FluentAssertions/MemberAssertions.cs
```diff
@@ -88,9 +89,70 @@ internal class MemberAssertions : ReferenceTypeAssertions<IMember, MemberAsserti
...
 
-        public AndConstraint<MemberAssertions> HaveSameMembersAs(IMember m) {
-            m.Should().BeAssignableTo<IMemberContainer>();
-            return HaveMembers(((IMemberContainer)m).GetMemberNames(), string.Empty);
+        public AndConstraint<MemberAssertions> HaveSameMemberNamesAs(IMember member) {
+            member.Should().BeAssignableTo<IMemberContainer>();
+            return HaveMembers(((IMemberContainer)member).GetMemberNames(), string.Empty);
+        }
+
+        public void HaveSameMembersAs(IMember other) {
```
MikhailArkhipov: Provides way to deeply compare type members

src/Caching/Impl/Factories/ClassFactory.cs
```diff
@@ -27,13 +27,14 @@ public ClassFactory(IEnumerable<ClassModel> classes, ModuleFactory mf)
...
 
         protected override PythonClassType CreateMember(ClassModel cm, IPythonType declaringType) 
-            => new PythonClassType(cm.Name, ModuleFactory.DefaultLocation);
+            => new PythonClassType(cm.Name, new Location(ModuleFactory.Module, cm.IndexSpan.ToSpan()));
 
         protected override void CreateMemberParts(ClassModel cm, PythonClassType cls) {
             // In Python 3 exclude object since type creation will add it automatically.
             var is3x = ModuleFactory.Module.Interpreter.LanguageVersion.Is3x();
             var bases = cm.Bases.Select(b => is3x && b == "object" ? null : TryCreate(b)).ExcludeDefault().ToArray();
             cls.SetBases(bases);
```
MikhailArkhipov: We now properly restore documentation and fields

src/Analysis/Ast/Impl/Analyzer/Handlers/FromImportHandler.cs
```diff
@@ -40,10 +40,16 @@ internal sealed partial class ImportHandler {
                 }
             }
 
-            var imports = ModuleResolution.CurrentPathResolver.FindImports(Module.FilePath, node);
-            if (HandleImportSearchResult(imports, null, null, node.Root, out var variableModule)) {
-                AssignVariables(node, imports, variableModule);
+            if(node.Root is RelativeModuleName) {
```
MikhailArkhipov: https://github.com/microsoft/python-language-server/pull/1380

src/Analysis/Ast/Impl/Analyzer/Handlers/ImportHandler.cs
```diff
@@ -50,16 +50,30 @@ internal sealed partial class ImportHandler : StatementHandler {
         }
 
         private void HandleImport(ModuleName moduleImportExpression, NameExpression asNameExpression, bool forceAbsolute) {
```
MikhailArkhipov: https://github.com/microsoft/python-language-server/pull/1380

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -227,8 +228,9 @@ public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst
             if (_stubAnalysis == null) {
                 return;
             }
-
-            var builtins = Module.Interpreter.ModuleResolution.BuiltinsModule;
+            // TODO: figure out why module is getting analyzed before stub.
+            // https://github.com/microsoft/python-language-server/issues/907
+            // Debug.Assert(!(_stubAnalysis is EmptyAnalysis));
```
MikhailArkhipov: https://github.com/microsoft/python-language-server/issues/907

src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -242,57 +244,152 @@ public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst
...
+            }
+            // Documentation and location are always get transferred from module type
+            // to the stub type and never the other way around. This makes sure that
+            // we show documentation from the original module and goto definition
+            // navigates to the module source and not to the stub.
+            if (s != d && s is PythonType src && d is PythonType dst) {
+                // If type is a class, then doc can either come from class definition node of from __init__.
+                // If class has doc from the class definition, don't stomp on it.
+                var transferDoc = true;
+                if (src is PythonClassType srcClass && dst is PythonClassType dstClass) {
```
MikhailArkhipov: We need to be able to tell if documentation came from class definition, from `__init__` or from the class base since we don't want lower priority sources stomping on the higher ones (i.e. class definition wins over `__init__`)

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
```diff
@@ -90,9 +94,16 @@ internal sealed class SymbolCollector : PythonWalker {
         }
 
         private PythonClassType CreateClass(ClassDefinition cd) {
-            var cls = new PythonClassType(cd, _eval.GetLocationOfName(cd),
+            PythonType declaringType = null;
+            if(!(cd.Parent is PythonAst)) {
+                Debug.Assert(_typeMap.ContainsKey(cd.Parent));
+                _typeMap.TryGetValue(cd.Parent, out declaringType);
+            }
```
MikhailArkhipov: Better handle inner classes

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
```diff
@@ -104,14 +115,16 @@ internal sealed class SymbolCollector : PythonWalker {
...
-                existing = new PythonFunctionType(fd, declaringType, _eval.GetLocationOfName(fd));
+        private void AddFunction(FunctionDefinition fd, PythonType declaringType) {
+            if (!(_eval.LookupNameInScopes(fd.Name, LookupOptions.Local) is PythonFunctionType f)) {
+                f = new PythonFunctionType(fd, declaringType, _eval.GetLocationOfName(fd));
                 // The variable is transient (non-user declared) hence it does not have location.
                 // Function type is tracking locations for references and renaming.
-                _eval.DeclareVariable(fd.Name, existing, VariableSource.Declaration);
+                _eval.DeclareVariable(fd.Name, f, VariableSource.Declaration);
+                _typeMap[fd] = f;
+                declaringType?.AddMember(f.Name, f, overwrite: true);
```
MikhailArkhipov: Better handle functions with members (i.e. classes and functions inside functions)

src/Analysis/Ast/Impl/Extensions/IfStatementExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+using Microsoft.Python.Parsing.Ast;
+
+namespace Microsoft.Python.Analysis {
+    public enum ConditionTestResult {
+        Unrecognized,
+        DontWalkBody,
+        WalkBody
+    }
+
+    public static class IfStatementExtensions {
```
MikhailArkhipov: https://github.com/microsoft/python-language-server/pull/1380

src/Analysis/Ast/Impl/Extensions/PythonFunctionExtensions.cs
```diff
@@ -15,18 +15,21 @@
...
     public static class PythonFunctionExtensions {
-        public static bool IsUnbound(this IPythonFunctionType f) 
+        public static bool IsUnbound(this IPythonFunctionType f)
             => f.DeclaringType != null && f.MemberType == PythonMemberType.Function;
 
-        public static bool IsBound(this IPythonFunctionType f) 
+        public static bool IsBound(this IPythonFunctionType f)
             => f.DeclaringType != null && f.MemberType == PythonMemberType.Method;
 
+        public static bool IsLambda(this IPythonFunctionType f) => f.Name == "<lambda>";
```
MikhailArkhipov: Need to be able to tell if function is lambda so we don't expose it in members and do not persist.

src/Analysis/Ast/Impl/Extensions/PythonWalkerExtensions.cs
```diff
@@ -0,0 +1,61 @@
...
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using Microsoft.Python.Parsing;
+using Microsoft.Python.Parsing.Ast;
+
+namespace Microsoft.Python.Analysis {
+    public static class PythonWalkerExtensions {
+        public static bool WalkIfWithSystemConditions(this IfStatement node, PythonWalker walker, PythonLanguageVersion languageVersion, bool isWindows) {
+            // System version, platform and os.path specializations
```
MikhailArkhipov: https://github.com/microsoft/python-language-server/pull/1380

PR #1378, opened 2019-07-24
Allowing classes to inherit from all specialized functions

### LINKED ISSUE #1377: metaclasses/inheriting from `type`
**Language server** 0.3.39.0 and also tried 0.3.40.0 from daily
**VSCode** 1.36.1
**Python** 3.7.3
**OS** Ubuntu 19.04

Using this simple example in a single file, no imports, etc:
```
class Foo(type):
    pass

class Bar(metaclass=Foo):
    pass
```

`Foo` is marked as inheriting from `type` which is not a class, similar to #1361 and #1364 
image

And the signature/docstring info for `type` is completely missing:
image

For a more concrete example, I discovered this problem with the following:

```
import weakref

class PackageMeta(type):
    def __new__(cls, name, bases, attrs):
        pkg_class = super(PackageMeta, cls).__new__(cls, name, bases, attrs)
        pkg_class.__instances = weakref.WeakValueDictionary()
        return pkg_class

    def __call__(cls, name, version, arch):
        return cls.__instances.setdefault(
            (name, version, arch,),
            super(PackageMeta, cls).__call__(name, version, arch))
```
Same problem here wherein `PackageMeta` is flagged for inheriting from `type`
image

but in this case, it's a bit more strange as the signature/docstring for `type` is now available but is wrong:
image

and if you peek at the definition...
image
Its coming from `http.client.HTTPMessage` which inherits from `email.message.Message` -- neither of which are imported here.

CTrando:
I think it would be good to have a PR in the future that would add a field to `IPythonFunctionType` that allowed linting to tell if something was specialized as a function for convenience but should be treated as a class. 

Fixes #1377

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
@@ -52,7 +52,7 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
                 // Evaluate inner classes, if any
                 EvaluateInnerClasses(_classDef);
                 _class = classInfo;
-                
+
                 var bases = ProcessBases();
                 _class.SetBases(bases);
                 // Declare __class__ variable in the scope.
@@ -131,9 +131,10 @@ namespace Microsoft.Python.Analysis.Analyzer.Symbols {
                 return true;
             }
 
-            // Allow any members from typing module
-            // TODO handle typing module specialization better: https://github.com/microsoft/python-language-server/issues/1367
-            if (m is ILocatedMember l && l.DeclaringModule is TypingModule) {
+            // Allow extensions from specialized functions 
+            // We specialized type to be a function even though it is a class, so this allows extension of type
+            // TODO handle module specialization better: https://github.com/microsoft/python-language-server/issues/1367
+            if (m is IPythonType t && t.IsSpecialized) {
                 return true;
             }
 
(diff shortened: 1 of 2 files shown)


src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -131,9 +131,10 @@ internal sealed class ClassEvaluator : MemberEvaluator {
                 return true;
             }
 
-            // Allow any members from typing module
-            // TODO handle typing module specialization better: https://github.com/microsoft/python-language-server/issues/1367
-            if (m is ILocatedMember l && l.DeclaringModule is TypingModule) {
+            // Allow extensions from specialized functions 
+            // We specialized type to be a function even though it is a class, so this allows extension of type
+            // TODO handle module specialization better: https://github.com/microsoft/python-language-server/issues/1367
+            if (m is IPythonType t && t.IsSpecialized) {
```
jakebailey: Does changing this check from `typing` to the more general `IsSpecialized` negatively affect anything that was previously allowed from `typing`? I'm not sure how it works, but I think that `typing.ByteString` is a class and should work, but I don't think that it's "specialized". Maybe it's handled at a different point.

CTrando: `typing.ByteString` is treated the same as int, so it is a class, not specialized to a function. I've added another test with some of the typing module functions just to verify however.

jakebailey: ByteString should be the same as `bytes`, but thanks for checking either way.

CTrando: Ah yes that's right.

PR #1380, opened 2019-07-24
Handle else in platform and version clauses + handle it in symbol collector

### LINKED ISSUE #1376: Functions collector does not handle language version
So in
```python
if sys.version_info >= (3, 3):
    def access(path: _FdOrPathType, mode: int, *, dir_fd: Optional[int] = ..., effective_ids: bool = ..., follow_symlinks: bool = ...) -> bool: ...
else:
    def access(path: _PathType, mode: int) -> bool: ...
```
Function gets both overloads.

MikhailArkhipov:
Closes #1376 
- Add handling of 'else' to platform and language version
- Make methods into extensions so they can be reused
- Make symbol collector to properly walk conditions using the above


MikhailArkhipov:
Reduced to conditional handling

PR #1396, opened 2019-07-31
Add RDT count to the analysis_complete event

### LINKED ISSUE #1379: Add RDT count to the analysis_complete event
The event currently sends the module count, time elapsed, and memory usage at the completion. However, the module count and time elapsed are on the analysis session, while memory usage is just overall. This means it's impossible to figure out a number like "this much memory per module on average", as a small session with only a few modules analyzed will still report the total usage.

Adding the RDT count will give a cumulative measure of what all is supposed to be there so we can do a generic count.

jakebailey:
Fixes #1379.

Can't use `GetDocuments().Count` as it collects the entire list under lock.

### DIFF
--- a/src/Analysis/Ast/Impl/Documents/Definitions/IRunningDocumentTable.cs
+++ b/src/Analysis/Ast/Impl/Documents/Definitions/IRunningDocumentTable.cs
@@ -94,6 +94,6 @@ namespace Microsoft.Python.Analysis.Documents {
         /// <summary>
         /// Gets the number of documents in the table.
         /// </summary>
-        int Count();
+        int DocumentCount { get; }
     }
 }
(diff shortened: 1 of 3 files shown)


src/Analysis/Ast/Impl/Documents/Definitions/IRunningDocumentTable.cs
```diff
@@ -90,5 +90,10 @@ public interface IRunningDocumentTable {
         /// Fires when document is removed.
         /// </summary>
         event EventHandler<DocumentEventArgs> Removed;
+
+        /// <summary>
+        /// Gets the number of documents in the table.
+        /// </summary>
+        int Count();
```
MikhailArkhipov: Should be property I think.

Issue #1298, opened 2019-07-09
High memory usage in 0.3.40(memory leak?)


CaselIT:
The fixed implemented in https://github.com/microsoft/python-language-server/issues/832 worked for some releases, but I'm having again high memory usage problems (10gb+) with version 0.3.20

These seems to be due to a memory leak: if finishes analyzing the project without problems, using less than 1gb in my case. After working on the project for some time it starts using more memory, even more than 10gb+.

I've not kept note of the details regarding after how long or if there is an action that triggers it. I usually notice a slow down, check the task manager and usually the language service is using many gb of memory. I have not kept track to see it the increase in memory usage is sudden or more gradual.

Below are the packages of the project used and some system information. I haven't checked if I can reproduce this issue in other projects

Is there some logging or telemetry I can enable to help with this issue?

Extension version: 2019.6.22090
Microsoft Python Language Server version 0.3.20.0
Python version: 3.6.8
VS Code version: Code 1.36.0 (0f3794b38477eea13fb47fbe15a42798e6129338, 2019-07-03T13:25:46.372Z)
OS version: Windows_NT x64 10.0.18362

<details>
<summary>requirements of the project</summary

`argon2_cffi==18.3.0`  
`python-dateutil==2.7.5`  
`decorator==4.3.0`  
`falcon>=2,<3`  
`falcon-auth==1.1.0`  
`falcon-cors==1.1.7`  
`graphene==2.1.3`  
`graphene_sqlalchemy==2.0.0`  
`jsonschema==2.6.0`  
`keras>=2.1.2`  
`numpy>=1.13.3`  
`ortools<7.1 `  
`pandas>=0.22.0,!=0.24.0`  
`psycopg2-binary>=2.7.3.2`  
`psycopg2<2.8`  
`pyjwt>=1.6.4`  
`scikit-learn>=0.19.1`  
`scipy>=1.0.0`  
`SQLAlchemy<1.3.0`  
`SQLAlchemy-Utils>=0.32.21`  
`sqlalchemy-postgres-copy>=0.5.0`  
`simplejson>=3.13.2`  
`tensorflow==1.12.0`  
`pyDOE>=0.3.8`  
`geomdl>=4.1.0`  
`pyomo==5.5.0`  
`pyutilib>=5.6.3`  
`joblib==0.11`  
`pytest>=4.4.0`  
`pytest-cov>=2.6.1`  
`yapf>=0.20.0,!=0.27`  
`flake8>=3.7.0`  
[...]

PR #1402, opened 2019-08-02
Remove all references to dropped modules and force GC on reload

jakebailey:
For #1298.

- Pulls ModuleWalker changes from #1372, as stub merging was adding things to places it shouldn't.
- Drops more ASTs (compiled/stubs).
- Prevents adding references to the builtin module (i.e. references to object).
- Resets both the typeshed and main module resolution.
- Clears the builtin module's AST map for a future reanalysis.
  - This needs to be reexamined; our method of modifying the scope of functions during calls to them is not right and needs to be fixed separately.
- Forces a GC/compact after a reload (to actually lower the memory usage after reloading)
- Drops `[Flag]` on `ModuleType`.
- Removes some dead code that would have caused reloads (but didn't need to).

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Definitions/IPythonAnalyzer.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Definitions/IPythonAnalyzer.cs
@@ -56,10 +56,12 @@ namespace Microsoft.Python.Analysis.Analyzer {
         /// </summary>
         IReadOnlyList<DiagnosticsEntry> LintModule(IPythonModule module);
 
+
         /// <summary>
-        /// Removes all the modules from the analysis, except Typeshed and builtin
+        /// Removes all the modules from the analysis and restarts it.
         /// </summary>
-        void ResetAnalyzer();
+        /// <param name="full">True if everything should be dropped, including closed files and stubs.</param>
+        Task ResetAnalyzer(bool full);
 
         /// <summary>
         /// Returns list of currently loaded modules.
--- a/src/Analysis/Ast/Impl/Analyzer/Definitions/IPythonAnalyzer.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Definitions/IPythonAnalyzer.cs
@@ -58,10 +58,9 @@ namespace Microsoft.Python.Analysis.Analyzer {
 
 
         /// <summary>
-        /// Removes all the modules from the analysis and restarts it.
+        /// Removes all the modules from the analysis and restarts it, including stubs.
         /// </summary>
-        /// <param name="full">True if everything should be dropped, including closed files and stubs.</param>
-        Task ResetAnalyzer(bool full);
+        Task ResetAnalyzer();
 
         /// <summary>
         /// Returns list of currently loaded modules.
--- a/src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
@@ -271,7 +271,7 @@ namespace Microsoft.Python.Analysis.Analyzer {
                             var memberType = member?.GetPythonType();
                             var stubMemberType = stubMember.GetPythonType();
 
-                            if (sourceType.IsBuiltin || stubType.IsBuiltin) {
+                            if (builtins.Equals(memberType?.DeclaringModule) || builtins.Equals(stubMemberType?.DeclaringModule)) {
                                 continue; // Leave builtins alone.
                             }
 
--- a/src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
@@ -255,7 +255,6 @@ namespace Microsoft.Python.Analysis.Analyzer {
                         if (v.Source == VariableSource.Declaration) {
                             Eval.DeclareVariable(v.Name, v.Value, v.Source);
                         }
-
                         break;
 
                     case PythonClassType cls when Module.Equals(cls.DeclaringModule):
@@ -286,6 +285,13 @@ namespace Microsoft.Python.Analysis.Analyzer {
                         }
                         break;
 
+                    case IPythonClassType _:
+                        // We do not re-declare classes, we only transfer members, see above.
+                        break;
+
+                    case IPythonModule _:
+                        // We do not re-declare modules.
+                        break;
 
                     default:
                         // Re-declare variable with the data from the stub unless member is a module.
@@ -301,10 +307,6 @@ namespace Microsoft.Python.Analysis.Analyzer {
                         break;
                 }
             }
-
-
-            var o = Eval.Interpreter.GetBuiltinType(BuiltinTypeId.Object);
-            Debug.Assert(o.DeclaringModule.ModuleType == ModuleType.Builtins);
         }
 
         private static bool IsStubBetterType(IPythonType sourceType, IPythonType stubType) {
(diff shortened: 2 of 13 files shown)


src/Analysis/Ast/Impl/Analyzer/ModuleWalker.cs
```diff
@@ -240,53 +242,101 @@ public ModuleWalker(IServiceContainer services, IPythonModule module, PythonAst
...
+                        // Do not replace __class__ since it has to match class type and we are not
+                        // replacing class type, we are only merging members.
+                        foreach (var name in stubType.GetMemberNames().Except(new[] { "__class__", "__base__", "__bases__", "__mro__", "mro" })) {
+                            var stubMember = stubType.GetMember(name);
+                            var member = cls.GetMember(name);
+
+                            var memberType = member?.GetPythonType();
+                            var stubMemberType = stubMember.GetPythonType();
+
+                            if (builtins.Equals(memberType?.DeclaringModule) || builtins.Equals(stubMemberType?.DeclaringModule)) {
```
MikhailArkhipov: Most recent looks like 
```cs
                            if (sourceType.IsBuiltin || stubType.IsBuiltin) {
                                // If stub type does not have an immediate member such as __init__() and
                                // rather have it inherited from object, we do not want to use the inherited
                                // since current class may either have its own of inherits it from the object.
                                continue;
                            }
```
which also prevents pulling inherited `__init__` and others from stub

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerSession.cs
```diff
@@ -368,10 +370,22 @@ private Task StartAnalysis(IDependencyChainNode<PythonAnalyzerEntry> node, Async
         }
 
         private IDocumentAnalysis CreateAnalysis(IDependencyChainNode<PythonAnalyzerEntry> node, IDocument document, PythonAst ast, int version, ModuleWalker walker, bool isCanceled) {
+            var mtIsDroppable = false;
```
MikhailArkhipov: Probably `canHaveLibraryAnalysis`

src/Analysis/Ast/Impl/Types/Definitions/IBuiltinPythonModule.cs
```diff
@@ -33,5 +33,7 @@ namespace Microsoft.Python.Analysis.Types {
     /// </summary>
     public interface IBuiltinsPythonModule : IPythonModule {
         IMember GetAnyMember(string name);
+
+        void ResetAst();
```
MikhailArkhipov: Dangerous to have this on a public interface. Safer to do internal method.

src/Analysis/Ast/Impl/Types/LocatedMember.cs
```diff
@@ -53,6 +53,10 @@ internal abstract class LocatedMember : ILocatedMember {
         }
 
         public virtual void AddReference(Location location) {
+            if (DeclaringModule == null || DeclaringModule.ModuleType == ModuleType.Builtins) {
```
MikhailArkhipov: Might be cleaner to override in PythonType and do `this.IsBuiltin` instead.

src/Analysis/Ast/Impl/Types/Definitions/IBuiltinPythonModule.cs
jakebailey: Internal meaning I try casting to the real thing and go from there?

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -269,19 +268,22 @@ public sealed partial class Server : IDisposable {
...
-            interpreter.ModuleResolution.ReloadAsync(cancellationToken).ContinueWith(t => {
-                _log?.Log(TraceEventType.Information, Resources.Done);
-                _log?.Log(TraceEventType.Information, Resources.AnalysisRestarted);
+
+            interpreter.ModuleResolution.BuiltinsModule.ResetAst();
+
+            interpreter.TypeshedResolution.ReloadAsync().ContinueWith(async t => {
+                await interpreter.ModuleResolution.ReloadAsync();
+
+                _services.GetService<PythonAnalyzer>().GCNextSession();
```
MikhailArkhipov: `<IPythonAnalyzer>`

jakebailey: I'll be moving things down into PythonAnalyzer instead, that way I don't have to modify IBuiltinModule and so on.

AlexanderSher: This doesn't have to be public API, there is `PythonAnalyzer.ResetAnalyzer` method that is called later.

src/Analysis/Ast/Impl/Analyzer/Definitions/IPythonAnalyzer.cs
```diff
@@ -56,10 +56,12 @@ public interface IPythonAnalyzer {
...
         IReadOnlyList<DiagnosticsEntry> LintModule(IPythonModule module);
 
+
         /// <summary>
-        /// Removes all the modules from the analysis, except Typeshed and builtin
+        /// Removes all the modules from the analysis and restarts it.
         /// </summary>
-        void ResetAnalyzer();
+        /// <param name="full">True if everything should be dropped, including closed files and stubs.</param>
+        Task ResetAnalyzer(bool full);
```
jakebailey: I had to modify this interface method. Should I duplicate it with another signature to ensure the old one is retained, or is this alright?

AlexanderSher: If we need to reload builtins on `NotifyPackagesChanged`, we probably should do the same on `DidChangeConfiguration`, so there is no reason for the `full == false` case.

jakebailey: In its current form, HandleConfigurationChanges only returns true if the diagnostics settings change. I can unconditionally do a full reload in that case, but we weren't before so I didn't want to change it. I just know that we don't want to do a reload on _any_ config change, as only some impact the analysis.

jakebailey: Actually, I'm surprised that the diagnostics settings cause any reloads at all. My impression was that changing them only affected the layer that exposes them to the editor. We could probably notify that of a change to have it re-filter and re-send things rather than a full reanalysis (nothing changed except the filters).

MikhailArkhipov: Yeah diag changes should not be reload

jakebailey: My thoughts were correct, the non-full branch is basically useless and doesn't need to be done at all, so I can clean this up and eliminate the boolean arg.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -196,18 +196,16 @@ public Task WaitForCompleteAnalysisAsync(CancellationToken cancellationToken = d
...
+            var interpreter = _services.GetService<IPythonInterpreter>();
+            var builtins = interpreter.ModuleResolution.BuiltinsModule;
+            builtins.SetAst(builtins.Analysis.Ast);
+
+            await interpreter.TypeshedResolution.ReloadAsync();
+            await interpreter.ModuleResolution.ReloadAsync();
 
             lock (_syncObj) {
-                _forceGCOnNextSession = _forceGCOnNextSession || full;
+                _forceGCOnNextSession = true;
```
AlexanderSher: If we reload `TypeshedResolution` and `BuiltinsModule`, there is no reason for a split bleow. _analysisEntries should be completely cleared, `_dependencyResolver` can be thrown away.

jakebailey: Sort of; the way we create the builtins module is really contrived and currently can only be created once when the interpreter object is created. What I had to do here is only a partial fix; we don't have a way to do a complete reload including the builtin module, so I kept the existing code and allowed it to remove more.

I'd like for this to be fixed in the future so we can have a "real" reload that drops everything but file contents and does a complete restart, but I think that'd be a bigger change.

PR #1404, opened 2019-08-05
Improving handling generics

### LINKED ISSUE #1390: False positive: inherit-non-class on generic subclass
I get an `inherit-non-class` error on the following.

```py
import typing as t

T = t.TypeVar("T")

class GenericThing(t.Generic[T]):
    pass

# Inheriting 'GenericThing[str]', which is not a class.Python(inherit-non-class)
class SpecificThing(GenericThing[str]):
    pass
```

This usage is specified in [pep-560](https://www.python.org/dev/peps/pep-0560/) and executes at runtime in Python 3.6 with no problems

CTrando:
Fixes #1390


src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericClassParameter.cs
```diff
@@ -13,22 +13,31 @@
...
+        internal GenericClassParameter(IReadOnlyList<IGenericTypeParameter> typeArgs, IPythonModule declaringModule)
+        : base("Generic", new Location(declaringModule)) {
+            TypeParameters = typeArgs;
         }
 
-        public IReadOnlyList<IGenericTypeDefinition> TypeDefinitions { get; }
+        public override bool IsGeneric => true;
+
+        public override IReadOnlyDictionary<IGenericTypeParameter, IPythonType> GenericParameters
+            => TypeParameters?.ToDictionary(tp => tp, tp => tp as IPythonType) ?? EmptyDictionary<IGenericTypeParameter, IPythonType>.Instance;
```
MikhailArkhipov: Do you want explicitly null as value or should it be `Unknown` for safety?

src/Analysis/Ast/Impl/Types/PythonType.cs
```diff
@@ -24,7 +24,7 @@ namespace Microsoft.Python.Analysis.Types {
     [DebuggerDisplay("{Name}")]
     internal class PythonType : LocatedMember, IPythonType {//, IEquatable<IPythonType> {
         private readonly object _lock = new object();
-        private readonly string _name;
+        protected readonly string _name;
         private Dictionary<string, IMember> _members;
```
MikhailArkhipov: If it is protected, make it a read-only property. Just .NET style thing.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -57,6 +53,16 @@ internal PythonClassType(string name, Location location)
             location.Module.AddAstNode(this, classDefinition);
         }
 
+        public override string Name {
+            get {
+                if (GenericParameters.Values.Count() > 0) {
```
MikhailArkhipov: Although analysis of a particular module is guaranteed to run on the same task, reading data such as class name can come from multiple threads. Consider module A imported by B and C. Analysis of A is a single task, but analysis of B and C will run in parallel reading data from A. So you need some lock here. In fact, I added lock for `CreateSpecificType` since I did hit the case when two threads attempted to modify collection concurrently.

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
```diff
@@ -0,0 +1,358 @@
...
+            }
+        }
+
+        #endregion
+
+
+        /// <summary>
+        /// Gets a list of distinct type parameters from bases and the class itself
+        /// </summary>
+        private IGenericTypeParameter[] GetTypeParameters() {
```
MikhailArkhipov: See threading comment above.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericClassParameter.cs
CTrando: Unknown is better good point.

CTrando: Actually upon further inspection, there should not be a way of getting null into GenericClassParameter.

src/Analysis/Ast/Impl/Types/PythonType.cs
CTrando: Ok

src/Analysis/Ast/Impl/Types/PythonClassType.cs
CTrando: I locked the caching part.

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
CTrando: This method only reads so it should be fine.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericClassParameter.cs
```diff
@@ -13,22 +13,31 @@
...
+        internal GenericClassParameter(IReadOnlyList<IGenericTypeParameter> typeArgs, IPythonModule declaringModule)
+        : base("Generic", new Location(declaringModule)) {
+            TypeParameters = typeArgs ?? new List<IGenericTypeParameter>();
         }
 
-        public IReadOnlyList<IGenericTypeDefinition> TypeDefinitions { get; }
+        public override bool IsGeneric => true;
+
+        public override IReadOnlyDictionary<IGenericTypeParameter, IPythonType> GenericParameters
+            => TypeParameters.ToDictionary(tp => tp, tp => tp as IPythonType) ?? EmptyDictionary<IGenericTypeParameter, IPythonType>.Instance;
```
MikhailArkhipov: I thought you wanted `tp => tp as IPythonType ?? UnknownType`

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
```diff
@@ -0,0 +1,360 @@
...
+            StoreGenericParameters(classType, genericTypeParameters, genericTypeToSpecificType);
+
+            // Locking so only one thread access cache
+            lock (_genericParameterLock) {
+                // Store generic parameters first so name updates correctly, then check if class type has been cached
+                _specificTypeCache = _specificTypeCache ?? new Dictionary<string, PythonClassType>();
+                if (_specificTypeCache.TryGetValue(classType.Name, out var cachedType)) {
+                    return cachedType;
+                }
+                _specificTypeCache[classType.Name] = classType;
```
MikhailArkhipov: So here is the problem with lock scope: thread 1 enters this and adds `classType` to the dictionary. Thread 2 enters and gets it from the dictionary. But class type hasn't been constructed fully yet - class bases and members are only set below, at line 234+. The lock has to fully protect type construction.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericClassParameter.cs
CTrando: Yes, sorry I forgot to push the change.

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
CTrando: I understand now, I will lock the initialization of the class then.

MikhailArkhipov: Actually problem is below, when members/bases are set, not when the class is new'ed. Basically, all code from current `lock() {` to the end of the function has to be locked.

CTrando: Yes that is what I did, by initialization I meant setting the members and bases as well.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -78,36 +77,38 @@ internal PythonClassType(string name, Location location)
         }
 
         public override IMember GetMember(string name) {
-            // Push/Pop should be lock protected.
-            if (Members.TryGetValue(name, out var member)) {
-                return member;
-            }
+            lock (_lock) {
```
jakebailey: This seems heavy handed. What condition is this fixing?

@MikhailArkhipov Is this alright?

CTrando: Without it, multiple threads can pop the stack which gave the stack empty exception from before. I believe the reason we did not previously have to lock it is because the `Push` `Pop` mechanism from before was much simpler and just set _processing to null.

src/Analysis/Ast/Impl/Types/PythonType.cs
```diff
@@ -23,8 +23,7 @@
 namespace Microsoft.Python.Analysis.Types {
     [DebuggerDisplay("{Name}")]
     internal class PythonType : LocatedMember, IPythonType {//, IEquatable<IPythonType> {
-        private readonly object _lock = new object();
-        private readonly string _name;
+        protected readonly object _lock = new object();
```
MikhailArkhipov: No protected variables please. If something is protected it should be readonly property. Why do we need protected lock?

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
```diff
@@ -106,5 +106,9 @@ public GenericTypeParameter(string name, IPythonModule declaringModule, IReadOnl
             var documentation = CodeFormatter.FormatSequence("TypeVar", '(', docArgs);
             return documentation;
         }
+
+        public bool Equals(IGenericTypeParameter other) {
```
MikhailArkhipov: use `=>` construct please

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -57,6 +53,17 @@ internal PythonClassType(string name, Location location)
             location.Module.AddAstNode(this, classDefinition);
         }
 
+        public override string Name {
+            get {
+                var genericParams = GenericParameters.Values.ToList();
```
MikhailArkhipov: This should be `{get; private set;}` assigned right when parameters are filled, no need for delayed construction - which also may end up having threading issues.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
MikhailArkhipov: Is this the case when class is generic and creation of specific type is being requested from multiple threads and the creation process is asking for the class members? In this case it should be sufficient to protect specific type creation.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
CTrando: Sure

src/Analysis/Ast/Impl/Types/PythonClassType.cs
CTrando: Ok I agree.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
MikhailArkhipov: Also, push is down below, total lock is not really necessary, it may only be needed for the MRO access. Can we add a trace which thread re-enters and Debug.Fail on it? Such as simple variable with owning thread id (`Thread.CurrentThread.ManagedThreadId`) that checks if another thread reenters and asserts so you can break into debugger and see it?

CTrando: I don't think so. I think this is just because I switched out the Push and Pop from before to the new reentrant guard. I can check this directly off master and see if that's the case as well.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -108,7 +117,7 @@ internal PythonClassType(string name, Location location)
                     return _documentation;
                 }
                 // Make sure we do not cycle through bases back here.
-                if (!Push(this)) {
+                if (!_memberGuard.Push(this)) {
```
MikhailArkhipov: Technically this can also be called during analysis (like user hovers over a method or variable) to get the doc and this will be another thread.

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
```diff
@@ -0,0 +1,368 @@
...
+        /// Where T is a generic type parameter that needs to be filled in by the class
+        /// </summary>
+        public virtual IReadOnlyList<IGenericTypeParameter> Parameters => _parameters ?? Array.Empty<IGenericTypeParameter>();
+
+        public virtual bool IsGeneric => _isGeneric;
+
+        /// <summary>
+        /// A Class is generic if it has one or more unfilled generic type parameters or one of its bases is generic
+        /// </summary>
+        public void DecideGeneric() {
```
MikhailArkhipov: Method does not seem to be part of `IGeneric`. Does it need to be public? You may want to move it out of the `#region`

src/Analysis/Ast/Impl/Types/PythonClassType.cs
MikhailArkhipov: I am trying to understand what is the case how you end up with empty stack. Even if there are 2 threads doing push1-push2-pop1-pop2 you should not get any exceptions. However, protecting stack inside the reentrancy guard on push and pop may be a good idea since `Stack<T>` is not a thread safe collection.

src/Analysis/Ast/Impl/Types/PythonType.cs
CTrando: This was because I had to use the same lock in PythonClassType.

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
CTrando: Yes I did some refactors and did that, just haven't pushed yet.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
CTrando: I will work on debugging this and get back to you.

src/Analysis/Ast/Impl/Types/PythonType.cs
MikhailArkhipov: Does it have to be the same lock?

src/Analysis/Ast/Impl/Types/PythonClassType.cs
CTrando: I tried replacing Stack with a ConcurrentStack and removing the locks and there are no issues anymore. I believe what could be happening is that the stack is being pushed to at the same time by two threads and then popped twice? Whatever the case it has to do with how Stack is not thread safe. I'll make the scope of the locks tighter.

MikhailArkhipov: Yeah that's what I thought. So you won't need global lock on `GetMember`. `ConcurrentStack` is probably an overkill, like `ConcurrentDictionary` it is generally for more advanced cases such as large stacks with heavy read/write. just add simple `lock()` around the stack operations. `lock(this)` is probably fine, its a small object with a short lock.

CTrando: Ok, sure I can lock just this.

src/Analysis/Ast/Impl/Types/PythonType.cs
CTrando: I was concerned about reading and writing members so I wanted to use the same lock, but after looking at it more it's really about pushing and popping with two different threads, not writing and reading.

PR #1406, opened 2019-08-05
Fix 'type name as string' conversion in scraper.

### LINKED ISSUE #1403: Scraper incorrectly declares IOError as instance
Instead of `IOError = OSError()` it should generate `IOError = OSError`. This bug has implications since classes in `Requests` derive from `IOError`.

MikhailArkhipov:
Closes #1403 

`type name as string` should not be adding `()` since it turns type info into an instance.
Changes version of the stub cache so it gets re-generated on user machine.

### DIFF
--- a/src/Analysis/Ast/Test/ClassesTests.cs
+++ b/src/Analysis/Ast/Test/ClassesTests.cs
@@ -634,5 +634,14 @@ class Test():
             // Test run time: typically ~ 20 sec.
             sw.ElapsedMilliseconds.Should().BeLessThan(60000); 
         }
+
+        [TestMethod, Priority(0)]
+        public async Task IOErrorBase() {
+            const string code = @"
+class A(IOError): ...
+";
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Should().HaveClass("A").Which.Should().HaveBase("OSError");
+        }
     }
 }
--- a/src/Analysis/Ast/Impl/Caching/StubCache.cs
+++ b/src/Analysis/Ast/Impl/Caching/StubCache.cs
@@ -23,7 +23,7 @@ using Microsoft.Python.Core.Logging;
 
 namespace Microsoft.Python.Analysis.Caching {
     internal sealed class StubCache : IStubCache {
-        private const int _stubCacheFormatVersion = 1;
+        private const int _stubCacheFormatVersion = 2;
 
         private readonly IFileSystem _fs;
         private readonly ILogger _log;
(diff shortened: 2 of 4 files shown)


src/Analysis/Ast/Test/ClassesTests.cs
```diff
@@ -634,5 +634,14 @@ class Test():
             // Test run time: typically ~ 20 sec.
             sw.ElapsedMilliseconds.Should().BeLessThan(60000); 
         }
+
+        [TestMethod, Priority(0)]
+        public async Task IOErrorBase() {
+            const string code = @"
+class A(IOError): ...
```
zooba: This example should be hitting the `_lines_with_members` function above I think.

The `str_from_typename` path is for an instance of the type, e.g.:

```python
class A: pass

GLOBAL_A = A(some, arguments, that, do, not, matter)
```

Becomes:

```
class A:
    """Documentation"""
    pass

GLOBAL_A = A()
```

MikhailArkhipov: The problem was in `IOError = OSError()`  line in builtins which then prevented `RequestException` from having `IOError` as a base.

zooba: Then there are two cases that both ought to be handled here. I guess when looking at the value we need to do an extra check to see if it's a direct alias or an instance of the type and set the scraped value appropriately.

zooba: So the `IOError` case should be going through here:

https://github.com/microsoft/python-language-server/blob/a018586c98d2fceee23e8c6fea3502d56241df0b/src/Analysis/Ast/Impl/scrape_module.py#L714-L721

And other cases that end up in the same spot would go through here:

https://github.com/microsoft/python-language-server/blob/a018586c98d2fceee23e8c6fea3502d56241df0b/src/Analysis/Ast/Impl/scrape_module.py#L734-L736

Maybe add an extra flag to one or the other to control whether the parens appear (yes for the second case; no for the first)?

MikhailArkhipov:
So how's about in `def __init__(`
```python
        if issubclass(value_type, type):
            self.instance = False
```
```python
    def _str_from_typename(self, type_name):
        mod_name, sep, name = type_name.rpartition('.')
        s = self.name + ' = ' + safe_module_name(mod_name) + sep + name
        if self.instance:
            s = s + '()'
        return s
```

src/Analysis/Ast/Test/ClassesTests.cs
MikhailArkhipov: So how's about in def __init__(

        if issubclass(value_type, type):
            self.instance = False
    def _str_from_typename(self, type_name):
        mod_name, sep, name = type_name.rpartition('.')
        s = self.name + ' = ' + safe_module_name(mod_name) + sep + name
        if self.instance:
            s = s + '()'
        return s

PR #1421, opened 2019-08-08
Properly add members to class rather than declaring variables

### LINKED ISSUE #1231: `self.` is not required to access class variables inside the class
image

`self.` should be required to access the instance variable. This is likely happening because the variable is "in" the scope.

Notably, this isn't just a completion thing, and affects the analysis:

image

But the tooltip displayed is different in the `self.` case:

image

Also, for class variables:

```python
val = 1234

class Foo:
    val = "str"

    def foo(self):
        print(val)


Foo().foo()
```

`val` in `foo` is `1234`, not "str", but:

image

This also affects the completion list:

image

`self.val` is the right way to access the one in the class.

MikhailArkhipov:
Fixes #1231 

Root cause is that class evaluator was declaring members as variables and then collected variables into members. Changes:
- introduce callback action that can decide if it wants to declare variable or add a class member.
- class variables type is added so completion can tell what to show depending on context
- class variables in class scope can be accesses without `self` while in method scopes and other places require `self`. Thus they are declared as both variables and members.

### DIFF
--- a/src/LanguageServer/Impl/Completion/TopLevelCompletion.cs
+++ b/src/LanguageServer/Impl/Completion/TopLevelCompletion.cs
@@ -19,6 +19,7 @@ using System.Linq;
 using Microsoft.Python.Analysis;
 using Microsoft.Python.Analysis.Analyzer.Expressions;
 using Microsoft.Python.Analysis.Types;
+using Microsoft.Python.Analysis.Values;
 using Microsoft.Python.Core;
 using Microsoft.Python.Core.Text;
 using Microsoft.Python.LanguageServer.Protocol;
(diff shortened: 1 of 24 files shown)


src/LanguageServer/Impl/Completion/TopLevelCompletion.cs
```diff
@@ -80,6 +81,14 @@ internal static class TopLevelCompletion {
             return new CompletionResult(items, applicableSpan);
         }
 
+        private static IReadOnlyList<IVariable> CollectVisibleVariables(IScope currentScope) {
+            var variables = new List<IVariable>();
+            foreach (var s in currentScope.EnumerateTowardsGlobal) {
+                variables.AddRange(s.Variables.Where(v => v.Source != VariableSource.ClassMember || s == currentScope));
```
jakebailey: Is there a check like this in the analysis? Filters in the completion code tend to hide behaviors of the analysis (like allowing access to something that doesn't exist), but if there's equivalent code there then it's probably not a big deal...

PR #1422, opened 2019-08-08
Ignoring assignment statements that occur before class and function declaration

### LINKED ISSUE #1419: False positive: inherit-non-class on Enum
This may be related to issue #1390 .

I have a false positive `inherit-non-class` error on the following code:

```py
import enum
from enum import Enum


# Inheriting 'enum.Enum', which is not a class.Python(inherit-non-class)
class FooEnum(enum.Enum):
    FOO: int = 1
    BAR: int = 2


# Inheriting 'Enum', which is not a class.Python(inherit-non-class)
class BarEnum(Enum):
    FOO: int = 1
    BAR: int = 2
```

Related Python documentation: [enum](https://docs.python.org/3/library/enum.html#creating-an-enum).

CTrando:
Fixes #1419.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Handlers/AssignmentHandler.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Handlers/AssignmentHandler.cs
@@ -84,7 +84,7 @@ namespace Microsoft.Python.Analysis.Analyzer.Handlers {
             if (Eval.GetInScope(name) is ILocatedMember m) {
                 // Class and function definition are processed first, so only override
                 // if assignment happens after declaration
-                if(loc.IndexSpan.Start < m.Location.IndexSpan.Start) {
+                if (loc.IndexSpan.Start < m.Location.IndexSpan.Start) {
                     return false;
                 }
             }
--- a/src/Analysis/Ast/Impl/Analyzer/Handlers/AssignmentHandler.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Handlers/AssignmentHandler.cs
@@ -82,7 +82,8 @@ namespace Microsoft.Python.Analysis.Analyzer.Handlers {
 
         private bool IsValidAssignment(string name, Location loc) {
             if (Eval.GetInScope(name) is ILocatedMember m) {
-                // If assignment happens before class or function declaration then ignore
+                // Class and function definition are processed first, so only override
+                // if assignment happens after declaration
                 if(loc.IndexSpan.Start < m.Location.IndexSpan.Start) {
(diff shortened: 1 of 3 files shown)


src/Analysis/Ast/Impl/Analyzer/Handlers/AssignmentHandler.cs
```diff
@@ -71,12 +71,26 @@ internal sealed class AssignmentHandler : StatementHandler {
...
             }
 
             TryHandleClassVariable(node, value);
         }
 
+        private bool IsValidAssignment(string name, Location loc) {
+            if (Eval.GetInScope(name) is ILocatedMember m) {
+                // Class and function definition are processed first, so only override
+                // if assignment happens after declaration
+                if(loc.IndexSpan.Start < m.Location.IndexSpan.Start) {
```
CTrando: Should I worry about assignment and class/function declarations on the same line? If not, then simply checking their line numbers is enough.

MikhailArkhipov: Technically we may simply avoid assignments to classes and functions completely... I.e. if `A is a class`, ignore `A= ...`

CTrando: But what if we have:
```
class A: ...


A = None
A = another_class_dec
```

It seems weird from a library creator perspective, but surely it could happen.

src/Analysis/Ast/Test/AssignmentTests.cs
```diff
@@ -308,6 +308,48 @@ class C:
...
+
+        [TestMethod, Priority(0)]
+        public async Task AssignBeforeFunctionDef() {
+            const string code = @"
+D = 5
+def D():
+    pass
+";
+            var analysis = await GetAnalysisAsync(code);
+            analysis.Should().HaveVariable("D").OfType("D");
```
jakebailey: Is "having D of type D" the right way to check that `D` is a function? Seems weird to me.

src/Analysis/Ast/Impl/Analyzer/Handlers/AssignmentHandler.cs
jakebailey: Nit: formatting of `if(` vs `if (`.

src/Analysis/Ast/Test/AssignmentTests.cs
CTrando: Ok

PR #1424, opened 2019-08-10
Special handling of variables representing class members

### LINKED ISSUE #1231: `self.` is not required to access class variables inside the class

MikhailArkhipov:
Fixes #1231

Root cause is that class evaluator was declaring members as variables and then collected variables into members. Smaller change than earlier.

- Add `IsClassMember` to variable. `VariableSource` does not fit since `class member` behaves like a flag.
- Add option to scope search to return class member variables.
- In linter and hover use option to obtain all variables. In completions use regular option.

### DIFF
--- a/src/Analysis/Ast/Impl/Types/PythonClassType.cs
+++ b/src/Analysis/Ast/Impl/Types/PythonClassType.cs
@@ -30,7 +30,7 @@ using Microsoft.Python.Parsing.Ast;
 
 namespace Microsoft.Python.Analysis.Types {
     [DebuggerDisplay("Class {Name}")]
-    internal partial class PythonClassType {
+    internal partial class PythonClassType : PythonType, IPythonClassType, IGenericType, IEquatable<IPythonClassType> {
         private static readonly string[] _classMethods = { "mro", "__dict__", @"__weakref__" };
 
         private readonly ReentrancyGuard<IPythonClassType> _memberGuard = new ReentrancyGuard<IPythonClassType>();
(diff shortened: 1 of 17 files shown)


jakebailey:
Using my examples in #1231, I can see that the analysis and completion reflect this change, but hovering doesn't:

image

image

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -30,10 +30,10 @@
 
 namespace Microsoft.Python.Analysis.Types {
     [DebuggerDisplay("Class {Name}")]
-    internal partial class PythonClassType : PythonType, IPythonClassType, IGenericType, IEquatable<IPythonClassType> {
+    internal partial class PythonClassType {
```
jakebailey: A nit, but I think that it should get removed from `PythonClassType.Generics.cs` rather than `PythonClassType.cs`, if anywhere.

jakebailey:
Not sure why, but `ReferencesTests.FindAllReferences_MultipleOpenFiles` and `ReferencesTests.UnrelatedFiles` both fail after that change.

PR #1435, opened 2019-08-13
When autogenerating function, do not add return for __init__

### LINKED ISSUE #1082: Auto-completion of __init__(self, ) should not automatically include a "return" statement
@PowerSnail commented on [Mon May 06 2019](https://github.com/microsoft/vscode-python/issues/5568)

## Environment data

- VS Code version: 1.33.1
- Extension version (available under the Extensions sidebar): 2019.4.12954
- OS and version: Ubuntu 19.04
- Python version (& distribution if applicable, e.g. Anaconda): Python 3.7.3 (system)
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): any 
- Relevant/affected Python packages and their versions: NA

## Expected behaviour

Auto-completion for `__init__` method should not generate `return super().__init__()`. 

`__init__` should always return None. And, explicit return in `__init__` is probably not appropriate anyway.

Maybe it should just generate an empty method with a `pass`, and let the user fill in the the necessary initialization. First, many classes don't inherit anything (other than `object`) and `super().__init__()` should not be there. Second, some libraries recommend initializing super by `super(MyClass, self).__init__()`. Users have to rewrite this anyway. 

## Actual behaviour

Auto-completion generates something like this when click "tab" when typing `__init__`:

```python
class NewClass(object):
    
    def __init__(self, *args, **kwargs):
        return super().__init__(*args, **kwargs)
```

## Steps to reproduce:

1. Make a new class
2. Type `def __ini`
3. Press `<tab>`

Thanks.

---

@Jonas1312 commented on [Wed May 08 2019](https://github.com/microsoft/vscode-python/issues/5568#issuecomment-490471319)

Same issue here

CTrando:
Fixes #1082

### DIFF
--- a/src/LanguageServer/Test/CompletionTests.cs
+++ b/src/LanguageServer/Test/CompletionTests.cs
@@ -123,6 +123,63 @@ class oar(list):
                 .And.HaveInsertTextFormat(InsertTextFormat.PlainText);
         }
 
+        [DataRow(PythonLanguageVersion.V37)]
+        [DataRow(PythonLanguageVersion.V36)]
+        [DataTestMethod]
+        public async Task OverrideInit3X(PythonLanguageVersion version) {
+            const string code = @"
+class Test():
+    def __
+";
+            var analysis = await GetAnalysisAsync(code, version);
+            var cs = new CompletionSource(new PlainTextDocumentationSource(), ServerSettings.completion);
+            var result = cs.GetCompletions(analysis, new SourceLocation(3, 10));
+
+            result.Should().HaveItem("__init__")
+                .Which.Should().HaveInsertText($"__init__(self, *args, **kwargs):{Environment.NewLine}    super().__init__(*args, **kwargs)")
+                .And.HaveInsertTextFormat(InsertTextFormat.PlainText);
+        }
+
+        [DataRow(PythonLanguageVersion.V26, "value")]
+        [DataRow(PythonLanguageVersion.V27, "value")]
+        [DataTestMethod]
+        public async Task OverrideCompletions2X(PythonLanguageVersion version, string parameterName) {
+            const string code = @"
+class oar(list):
+    def 
+    pass
+";
+            var analysis = await GetAnalysisAsync(code, version);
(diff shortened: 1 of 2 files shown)


src/LanguageServer/Test/CompletionTests.cs
```diff
@@ -123,6 +123,63 @@ class oar(list):
...
+            var analysis = await GetAnalysisAsync(code, version);
+            var cs = new CompletionSource(new PlainTextDocumentationSource(), ServerSettings.completion);
+            var result = cs.GetCompletions(analysis, new SourceLocation(3, 10));
+
+            result.Should().HaveItem("__init__")
+                .Which.Should().HaveInsertText($"__init__(self, *args, **kwargs):{Environment.NewLine}    super().__init__(*args, **kwargs)")
+                .And.HaveInsertTextFormat(InsertTextFormat.PlainText);
+        }
+
+        [DataRow(PythonLanguageVersion.V26, "value")]
```
MikhailArkhipov: Seems to use same "value" in both data cases

src/LanguageServer/Test/CompletionTests.cs
```diff
@@ -123,6 +123,63 @@ class oar(list):
                 .And.HaveInsertTextFormat(InsertTextFormat.PlainText);
         }
 
+        [DataRow(PythonLanguageVersion.V37)]
+        [DataRow(PythonLanguageVersion.V36)]
+        [DataTestMethod]
```
MikhailArkhipov: Up to, but you can just use `LatestAvailable3X`

src/LanguageServer/Test/CompletionTests.cs
CTrando: The function signature remained the same between the two. The parameter name actually seems to come from the list class. Between 36 and 37, the function parameter changed to object.

src/LanguageServer/Test/CompletionTests.cs
CTrando: Ok

Issue #1401, opened 2019-08-02
Failure analyzing decorator module, high CPU/RAM usage


uriva:
Hi,

I have `"python.jediEnabled": false` in my `settings.json`.

Whenever I open a project, I see "analyzing in background" for a bit, then it stops. Usually it's 1200 targets or something similar.

Then from time to time I have a spike of cpu (200%) and ram usage from the process, and the code is again "analyzed in background".

This is coming from `.vscode/extensions/ms-python.python-2019.6.24221/languageServer.0.3.43/Microsoft.Python.LanguageServer`.

Over the last month the cpu/ram issues were on and off. At first they were making the computer stuck (I had to restart) and other times it went away after awhile. I saw some issues were opened and the closed, but over the last two weeks it again became hard to work with vscode because of this, so probably the problem isn't completely fixed.

MikhailArkhipov:
Could you provide
- Amount of RAM used (few hundred meg is not unusual)
- List of packages you use beside system (i.e. tensorflow, numpy, plotly?)
- How long CPU spike may last (LS is heavily multi-threaded so using multiple cores is normal unless it lasts for a long time).

uriva:
I can't easily reproduce, but an hour ago I edited some code and it seems to re-analyze on every code change, then the entire cpu is unresponsive and I have to power down, so it's more than a few hundred megs.

Now by just opening the project and waiting for awhile, it reaches 6GB ram consumption.

I think spacy and selenium are the heaviest dependencies I have (no tensorflow).

uriva:
Another thing that may be of interest is that I have two projects opened sometimes, and one depends on the other (pip install -e).

uriva:
I think the cpu is less of a worry than ram is (makes the mouse unusable).

jakebailey:
If they depend on each other and you open them separately, then you'll have two language servers running watching both such that changing one causes the other to be reanalyzed. If that's happening often, then this may be the module reloading memory usage issue I'm currently working on.

uriva:
Now for example it's analyzing >5 minutes, and the computer intermittently freezes.

uriva:
@jakebailey I think this is related, as it's aggravated when I'm working on the two projects at the same time, but it's not explaining the very long analysis even when there's no code change.

uriva:
I'm also using ipython with autoreload in case it matters.

jakebailey:
[...]

PR #1438, opened 2019-08-14
Log full exception on session failure

jakebailey:
For #1401.

We should really have more info here than just the message, but I'm also worried that this might be _too_ much for long stack traces. Thoughts?


MikhailArkhipov:
Add stack trace?

jakebailey:
Unless I'm mistaken, `Exception.ToString()` includes that info.

PR #1441, opened 2019-08-14
String forward reference support

### LINKED ISSUE #1186: String forward references not working as generic type parameters
When using "stringy" types such as `List["SomeClass"]` and a `for` loop over that (or any other operation that returns the element type), the element in the loop is not resolved as type `SomeClass`

Full repro example (see the comments for where the issue appears):

```python
from typing import List

class Bar:
    def bar_method(self) -> int:
        return 1

def list_of_bars() -> List["Bar"]:
    return []

def no_string_list_of_bars() -> List[Bar]:
    return []

def a_single_bar() -> "Bar":
    return Bar()


def test():
    single = a_single_bar()

    single.bar_method()

    l = list_of_bars()

    for item in l: # Hover: Type of item is unresolved
        item.bar_method() # Go to definition does not work

    l2 = no_string_list_of_bars()

    for item2 in l2: # Type is resolved here
        item2.bar_method() # Ctrl+click works

```

image

image

<details><summary>Extension information:</summary>

Microsoft Python Language Server version 0.2.96.0

Name: Python
Id: ms-python.python
Description: Linting, Debugging (multi-threaded, remote), Intellisense, code formatting, refactoring, unit tests, snippets, and more.
Version: 2019.5.18875
Publisher: Microsoft
VS Marketplace Link: https://marketplace.visualstudio.com/items?itemName=ms-python.python
</details>

CTrando:
Fixes #1186.
Handles cases such as:

```
class A: ...
class Bar: ...
def tmp(v: 'List[str]') -> 'List[str]'
def tmp(v: 'Bar') -> List['Bar']
def tmp(v: 'str') -> 'A'
```

So instead of returning unknown, we try to evaluate what is inside the quotes.
* Updated the type annotation hint code to handle forward references so parameters get the correct types.
* Handled values in quotes when creating Generic types
* Handled forward reference return value hints

### DIFF
--- a/src/Analysis/Ast/Test/GenericsTests.cs
+++ b/src/Analysis/Ast/Test/GenericsTests.cs
@@ -1116,6 +1116,30 @@ d = Dict['float', 'str']
         }
 
         [TestMethod, Priority(0)]
+        public async Task GenericListForwardRef() {
+            const string code = @"
+from typing import List, Dict
+class A: ...
+class B: ...
+
+def test() -> 'List[A]':
+    pass
+
+def test1() -> 'Dict[A, B]':
+    pass
+
+x = test()
+y = test1()
+";
+            var analysis = await GetAnalysisAsync(code, PythonVersions.LatestAvailable3X);
+            analysis.Should().HaveVariable("x")
+                .Which.Should().HaveType("List[A]");
+            analysis.Should().HaveVariable("y")
+                        .Which.Should().HaveType("Dict[A, B]");
+        }
+
+
+        [TestMethod, Priority(0)]
         public async Task GenericClassForwardRef() {
             const string code = @"
 from typing import Generic, TypeVar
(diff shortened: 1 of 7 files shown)


src/Analysis/Ast/Test/GenericsTests.cs
```diff
@@ -1075,6 +1075,109 @@ class Box(Generic[_T]):
...
+            const string code = @"
+from typing import List, Dict
+class A:
+    def test(self) -> int:
+        pass
+class B:
+    def test(self) -> int:
+        pass
+l = List['A']
+d = Dict['A', 'B']
```
MikhailArkhipov: Does it work if types are declared later in the file - since it is supposed to be 'forward' reference?

CTrando: Ok I'll check

CTrando:
I updated the tests because the return documentation for a method inside a class is our own analyzed return value, but for unbound functions, we just use the regular return documentation. I think to be consistent we should also return our analyzed return value.

Issue #1156, opened 2019-06-03
Unresolved import for builtin packages after reload


DonJayamanne:
@schlamar commented on [Mon Jun 03 2019](https://github.com/microsoft/vscode-python/issues/5881)

## Environment data

- VS Code version: 1.34.0
- Extension version (available under the Extensions sidebar): 2019.5
- OS and version: Windows 7
- Python version (& distribution if applicable, e.g. Anaconda): 2.7
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): N/A
- Relevant/affected Python packages and their versions: N/A

## Expected behaviour

No errors for builtin packages.

## Actual behaviour

Import of builtin packages like `time`, `itertools` and `operator` are sometimes marked as unresolved import.

## Steps to reproduce:

Don't know, this happens randomly.

## Logs

```
Starting Microsoft Python language server.
##########Linting Output - flake8##########
[Info  - 12:19:26] GetCurrentSearchPaths C:\Python27\python.exe 
[Info  - 12:19:26] Python search paths:
[Info  - 12:19:26]     c:\python27\dlls
[Info  - 12:19:26]     c:\python27\lib
[Info  - 12:19:26]     c:\python27\lib\lib-tk
[Info  - 12:19:26]     c:\python27
[Info  - 12:19:26]     c:\python27\lib\site-packages
[Info  - 12:19:26] Configuration search paths:
[Info  - 12:19:26]     ...
[Info  - 12:19:33] Microsoft Python Language Server version 0.2.87.0
[Info  - 12:19:33] Initializing for C:\Python27\python.exe
```   

(Right now, I don't have a Console Log in case of this error. I will submit them as soon as this happens again.)

thernstig:
This happens to our project as well, e.g. for the `sys` module, but other modules work just fine. Example of our imports:

```python
from __future__ import (absolute_import, division, generators, nested_scopes,
                        print_function, with_statement)

import os
import re
import subprocess
import sys
```

I am using the language server (i.e this is our setting ` "python.jediEnabled": false,`).

johnybx:
Same problem with python 3.6. I do not really think that this problem is related to python version though I did not test all versions. I noticed this problem after using 

`"python.autoComplete.extraPaths": [
        "${workspaceFolder}/src"
    ],`
and using autosave with autopep8 few times (seem like some pep8 formating must occure but not sure). Simple folder structure:
```
root/
    \_env
    \_src
    \_.vscode
```

[...]

PR #1446, opened 2019-08-16
Merges from db

AlexanderSher:
- Fix the order of builtins initialization during reload
- Remove some of the dead code
- Add file size information to the module import


AlexanderSher:
May fix #1156

PR #1458, opened 2019-08-19
Fixing partial generic params for specialized generic types

### LINKED ISSUE #1454: Filling generic arguments partially creates incorrect class
Some fallout of https://github.com/microsoft/python-language-server/pull/1404

```python
from typing import Dict, Generic, TypeVar

T = TypeVar('T')
class A(Genetic[T], Dict[T, int]) : ...
```

`A` actually have base of `Dict[T]` rather than proper dictionary with 2 parameters. `Dict[T]` cannot be re-created b/c number of arguments in it is wrong.

```cs
        [TestMethod, Priority(0)]
        public async Task DictPartialParams() {
            const string code = @"
from typing import Dict, Generic, TypeVar

T = TypeVar('T')
class A(Genetic[T], Dict[T, int]) : ...
";

            analysis.Should().HaveClass("A")
                .Which.Should().HaveBase("Dict[T, int]");
        }
```

fails. @CTrando fyi

CTrando:
Fixes #1454

### DIFF
--- a/src/Analysis/Ast/Test/FluentAssertions/MemberAssertions.cs
+++ b/src/Analysis/Ast/Test/FluentAssertions/MemberAssertions.cs
@@ -61,7 +61,7 @@ namespace Microsoft.Python.Analysis.Tests.FluentAssertions {
             return new AndWhichConstraint<MemberAssertions, IMember>(this, Subject);
         }
 
-        public AndWhichConstraint<MemberAssertions, IPythonClassType> HaveBase(string name, string because = "", params object[] reasonArgs) {
+        public AndWhichConstraint<MemberAssertions, IPythonType> HaveBase(string name, string because = "", params object[] reasonArgs) {
             NotBeNull();
 
             var cls = Subject.GetPythonType<IPythonClassType>();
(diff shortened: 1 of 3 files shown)


src/Analysis/Ast/Test/FluentAssertions/MemberAssertions.cs
```diff
@@ -61,20 +61,20 @@ internal class MemberAssertions : ReferenceTypeAssertions<IMember, MemberAsserti
             return new AndWhichConstraint<MemberAssertions, IMember>(this, Subject);
         }
 
-        public AndWhichConstraint<MemberAssertions, IPythonClassType> HaveBase(string name, string because = "", params object[] reasonArgs) {
+        public AndWhichConstraint<MemberAssertions, IPythonType> HaveBase(string name, string because = "", params object[] reasonArgs) {
```
MikhailArkhipov: If you want, you can borrow https://github.com/MikhailArkhipov/python-language-server/blob/functools/src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericClassBase.cs which is former `GenericTypeParameter` and is actually a class - as it is supposed to be.

src/Analysis/Ast/Test/GenericsTests.cs
```diff
@@ -1091,6 +1090,37 @@ class Simple(unittest.TestCase):
             analysis.Should().HaveVariable("x").Which.Should().HaveType("TypeError");
         }
 
+        [TestMethod, Priority(0)]
+        public async Task SpecializedGenericPartialParams() {
+            const string code = @"
+from typing import Dict, List, Tuple, Iterator, KeysView, ValuesView, ItemsViews, Sequence, Generic, TypeVar
```
MikhailArkhipov: Does the test provided in the issue description pass?

CTrando: Yes it does.

src/Analysis/Ast/Test/FluentAssertions/MemberAssertions.cs
CTrando: If we just want to check if a base is inside the Bases list, then we should just need to pass in the name of the base. By having it as PythonClassType, we can have cases with typing or builtins where the type is a PythonType but we look for PythonClassType. Unless we change Bases to a list of IPythonClassTypes, then I think this is clearer.

src/Analysis/Ast/Test/GenericsTests.cs
CTrando: Your test is included in this one.

PR #1471, opened 2019-08-22
Handling dependencies in module persistence

### LINKED ISSUE #1468: Integrate persistent modules into dependency analysis
Dependency analysis needs to know what module imports and since AST is not present in db modules, dependency analysis does not load modules in proper order.

MikhailArkhipov:
Fixes #1468

Simplify integration of persistence into the current pipeline:
- Module dependencies (imports) are stored in the database.
- Instead of restoring module, restore global scope during analysis. This way analysis session looks the same as for regular modules.
- Implement `IDependencyProvider` so regular modules can employ AST walk for imports and database-backed modules restore dependencies from the database.
- Move `Analyze` to module (add to `IAnalyzable`) since only regular modules employ `ModuleWalker`.
- Separate declaration of variables during restore from population of members. This is necessary to handle forward declarations so we know classes and functions early and can add members to them later.


src/LanguageServer/Test/SymbolIndexTests.cs
```diff
@@ -67,124 +68,130 @@ public class SymbolIndexTests {
...
-            };
-            cancelled.Should().Throw<OperationCanceledException>();
+            using (var index = MakeSymbolIndex()) {
+                var path = TestData.GetDefaultModulePath();
+
+                index.Add(path, DocumentWithAst("x = 1"));
+                index.MarkAsPending(path);
+                using (var cts = new CancellationTokenSource()) {
+                    var t = index.HierarchicalDocumentSymbolsAsync(path, cts.Token);
+                    t.IsCompleted.Should().BeFalse();
```
AlexanderSher: `t.Should().NotBeCompleted()`

src/LanguageServer/Test/IndexParserTests.cs
```diff
@@ -81,8 +81,9 @@ public class IndexParserTests : LanguageServerTestBase {
             _fileSystem.FileExists(testFilePath).Returns(true);
             SetFileOpen(_fileSystem, testFilePath, _ => throw new FileNotFoundException());
 
-            IIndexParser indexParser = new IndexParser(_fileSystem, _pythonLanguageVersion);
-            var ast = await indexParser.ParseAsync(testFilePath);
+            using (IIndexParser indexParser = new IndexParser(_fileSystem, _pythonLanguageVersion)) {
```
AlexanderSher: `var`

src/Caching/Impl/TypeNames.cs
```diff
@@ -101,7 +100,12 @@ internal static class TypeNames {
                         parts.MemberNames = typeName == "..." ? new[] { "ellipsis" } : typeName.Split('.').ToArray();
                         break;
                     default:
-                        parts.ModuleName = typeName;
+                        if (typeName.EndsWith("(stub)")) {
```
AlexanderSher: Do we have test for this?

src/Caching/Impl/RestoredGlobalScope.cs
```diff
@@ -22,44 +22,45 @@
...
-            }
-            foreach (var cm in _model.Classes) {
-                var cls = cm.Construct(mf, null);
-                _scopeVariables.DeclareVariable(cm.Name, cls, VariableSource.Declaration, mf.DefaultLocation);
-            }
-            foreach (var fm in _model.Functions) {
-                var ft = fm.Construct(mf, null);
-                _scopeVariables.DeclareVariable(fm.Name, ft, VariableSource.Declaration, mf.DefaultLocation);
+            var mf = new ModuleFactory(_model, Module, this);
+            var models = _model.TypeVars.Concat<MemberModel>(_model.NamedTuples).Concat(_model.Classes).Concat(_model.Functions).ToArray();
```
AlexanderSher: Do we need `ToArray()` here?

src/Caching/Impl/Properties/AssemblyInfo.cs
```diff
@@ -16,3 +16,4 @@
 using System.Runtime.CompilerServices;
 
 [assembly: InternalsVisibleTo("Microsoft.Python.Analysis.Caching.Tests, PublicKey=002400000480000094000000060200000024000052534131000400000100010007d1fa57c4aed9f0a32e84aa0faefd0de9e8fd6aec8f87fb03766c834c99921eb23be79ad9d5dcc1dd9ad236132102900b723cf980957fc4e177108fc607774f29e8320e92ea05ece4e821c0a5efe8f1645c4c0c93c1ab99285d622caa652c1dfad63d745d6f2de5f17e5eaf0fc4963d261c8a12436518206dc093344d5ad293")]
+[assembly: InternalsVisibleTo("Microsoft.Python.LanguageServer, PublicKey=002400000480000094000000060200000024000052534131000400000100010007d1fa57c4aed9f0a32e84aa0faefd0de9e8fd6aec8f87fb03766c834c99921eb23be79ad9d5dcc1dd9ad236132102900b723cf980957fc4e177108fc607774f29e8320e92ea05ece4e821c0a5efe8f1645c4c0c93c1ab99285d622caa652c1dfad63d745d6f2de5f17e5eaf0fc4963d261c8a12436518206dc093344d5ad293")]
```
AlexanderSher: What is the type that required `InternalsVisibleTo`?

src/Caching/Impl/ModuleFactory.cs
```diff
@@ -31,22 +31,26 @@ namespace Microsoft.Python.Analysis.Caching {
     /// Constructs module from its persistent model.
     /// </summary>
     internal sealed class ModuleFactory {
+        /// <summary>For use in tests so missing members will assert.</summary>
+        internal static bool EnableMissingMemberAssertions { get; set; }
```
AlexanderSher: It will be shared between tests

MikhailArkhipov: Yes, that is the intent. Basically, assert if cache tests cannot restore a member. At runtime this is normal until dependencies are fully analyzed.

src/Caching/Impl/Properties/AssemblyInfo.cs
MikhailArkhipov: `ModuleDatabase` so it can be added to services. It could be made public, but then via `IDependencyProvider` the `AnalysisModuleKey` becomes public.

src/Caching/Impl/TypeNames.cs
MikhailArkhipov: Yes, some of `LibraryModulesTests` would fail otherwise.

src/LanguageServer/Test/IndexParserTests.cs
MikhailArkhipov: Yes

src/LanguageServer/Test/SymbolIndexTests.cs
MikhailArkhipov: OK

src/Analysis/Ast/Impl/Types/PythonType.cs
```diff
@@ -159,16 +159,14 @@ public virtual IMember Call(IPythonInstance instance, string memberName, IArgume
             }
         }
 
-        internal void MakeReadOnly() => _readonly = true;
+        internal void MakeReadOnly() {
+            lock (_lock) {
+                _readonly = true;
```
AlexanderSher: Should `AddMember` notify somehow about `PythonType` being sealed?

src/Caching/Impl/RestoredGlobalScope.cs
MikhailArkhipov: Nope

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
```diff
@@ -127,7 +127,7 @@ internal partial class PythonClassType {
                     }
 
                     // Set specific class bases 
-                    classType.SetBases(specificBases.Concat(newBases), args.Eval.CurrentScope);
+                    classType.SetBases(specificBases.Concat(newBases), args.Eval?.CurrentScope);
```
AlexanderSher: When does the `args.Eval` become `null`?

MikhailArkhipov: When class is restored from the database, there is no evaluator. However, in the restore case there is no need to disambiguate bases and unroll loops since they were unrolled during analysis and before the persistence.

MikhailArkhipov: Added comment

src/Analysis/Ast/Impl/Types/PythonType.cs
MikhailArkhipov: This can probably be removed. It was protection from builtins and some other types being modified during stub merge. With the reversal of stub merge direction this probably no longer needed. https://github.com/microsoft/python-language-server/issues/1503

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
```diff
@@ -106,8 +107,12 @@ internal sealed class GenericTypeParameter : PythonType, IGenericTypeParameter {
             switch (rawBound) {
                 case IPythonType t:
                     return t;
-                case IPythonConstant c when c.GetString() != null:
-                    return eval.GetTypeFromString(c.GetString());
+                case IPythonConstant c:
+                    var s = c.GetString();
+                    if (!string.IsNullOrEmpty(s)) {
+                        return eval.GetTypeFromString(c.GetString()) ?? argSet.Eval.UnknownType;
```
AlexanderSher: You can reuse `s` here

src/Analysis/Ast/Impl/Modules/PythonVariableModule.cs
```diff
@@ -80,5 +81,10 @@ internal sealed class PythonVariableModule : LocatedMember, IPythonModule, IEqua
         public SourceLocation IndexToLocation(int index) => (Module as ILocationConverter)?.IndexToLocation(index) ?? default;
         public int LocationToIndex(SourceLocation location) => (Module as ILocationConverter)?.LocationToIndex(location) ?? default;
         #endregion
+
+        #region IDependencyProvider
+        public HashSet<AnalysisModuleKey> GetDependencies(PythonAst ast) 
+            => (Module as IDependencyProvider)?.GetDependencies(ast) ?? new HashSet<AnalysisModuleKey>();
```
AlexanderSher: Is it possible that `Module` isn't a `IDependencyProvider`?

src/Caching/Impl/ModuleDatabase.cs
```diff
@@ -191,7 +208,52 @@ public Task StoreModuleAnalysisAsync(IDocumentAnalysis analysis, CancellationTok
...
+            for (var retries = 50; retries > 0; --retries) {
+                try {
+                    // TODO: make combined db rather than per module?
+                    var dbPath = FindDatabaseFile(moduleName, filePath);
+                    if (string.IsNullOrEmpty(dbPath)) {
+                        return false;
+                    }
+
+                    using (var db = new LiteDatabase(dbPath)) {
+                        if (!db.CollectionExists("modules")) {
```
AlexanderSher: Are there async versions of these methods?

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
```diff
@@ -248,20 +247,20 @@ public Task<IDocumentAnalysis> GetAnalysisAsync(CancellationToken cancellationTo
         }
 
         private HashSet<AnalysisModuleKey> FindDependencies(IPythonModule module, PythonAst ast, int bufferVersion) {
+            var dependencies = new HashSet<AnalysisModuleKey>();
             if (_bufferVersion > bufferVersion) {
-                return new HashSet<AnalysisModuleKey>();
+                return dependencies;
+            }
+
+            var moduleDeps = (module as IDependencyProvider)?.GetDependencies(ast).ToArray();
```
AlexanderSher: `dependencies = module.GetDependencies(ast)` should be enough, cause `dependencies` is an empty hashset at this point

src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -585,12 +658,25 @@ public Task<IDocumentAnalysis> GetAnalysisAsync(int waitTime = 200, Cancellation
         public virtual int LocationToIndex(SourceLocation location) => this.GetAst()?.LocationToIndex(location) ?? default;
         #endregion
 
+        #region IDependencyProvider
+        public virtual HashSet<AnalysisModuleKey> GetDependencies(PythonAst ast) {
+            var dbs = GetDbService();
+            if (dbs != null && dbs.TryRestoreDependencies(this, out var dp)) {
```
AlexanderSher: So `TryRestoreDependencies` is used only here, and `GetDependencies` is used only in `PythonAnalyzerEntry`. It feels like this logic can be left inside `PythonAnalyzerEntry`.

src/Analysis/Ast/Impl/Modules/PythonVariableModule.cs
MikhailArkhipov: Generally no... But it may be simpler for modules that do not have dependencies like builtins or specialized. Sort of like we handle `IAnalyzable`.

src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
MikhailArkhipov: Yeah, that was the intent...

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
MikhailArkhipov: OK

src/Caching/Impl/ModuleDatabase.cs
MikhailArkhipov: No. Should we have them?

AlexanderSher: If LiteDB doesn't provide async IO, then there is not much we can do.

src/Analysis/Ast/Impl/Modules/PythonModule.cs
MikhailArkhipov: It was mostly to isolate dependency analysis from knowledge if particular module has AST or has other means of fetching dependencies. This can be extended to handle that `LoadFunctionDependencyModules` - module that has that kind of functions can supply additional dependencies in a normal way.

MikhailArkhipov: Updated per discussion

src/Analysis/Ast/Impl/Modules/PythonVariableModule.cs
MikhailArkhipov: Updated per discussion

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
```diff
@@ -248,20 +246,21 @@ public Task<IDocumentAnalysis> GetAnalysisAsync(CancellationToken cancellationTo
         }
 
         private HashSet<AnalysisModuleKey> FindDependencies(IPythonModule module, PythonAst ast, int bufferVersion) {
+            var dependencies = new HashSet<AnalysisModuleKey>();
             if (_bufferVersion > bufferVersion) {
-                return new HashSet<AnalysisModuleKey>();
+                return dependencies;
+            }
+
+            var dependencyProvider = (module as IAnalyzable)?.DependencyProvider;
```
AlexanderSher:
```cs
var dependencies = _bufferVersion <= bufferVersion && module is IAnalyzable analyzable && analyzable.DependencyProvider != null
    ? dependencyProvider.GetDependencies()
    : new HashSet<AnalysisModuleKey>();
```

MikhailArkhipov: OK

PR #1477, opened 2019-08-23
Adding support for egg and zip files

### LINKED ISSUE #196: Support egg and zip files
From Microsoft/vscode-python#2532

`.egg` files are zip files and Python supports having `.zip` files on `sys.path` to package up source code.

CTrando:
Fixes #196


src/Analysis/Ast/Impl/get_search_paths.py
```diff
@@ -68,13 +68,18 @@ def clean(path):
...
-            print("%s|stdlib|" % p)
-        elif p in AFTER_SITE:
-            if p in SITE_PKGS:
-                print("%s|site|" % p)
-            else:
-                print("%s|pth|" % p)
+
+    if not os.path.isdir(p) and not (os.path.isfile(p) and zipfile.is_zipfile(p)):
+	    continue
+
```
CTrando: Added this so interpreter paths would include .egg and .zip

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
```diff
@@ -38,10 +38,7 @@ internal static class CacheFolders {
             return filePath;
         }
 
-        public static string GetCacheFolder(IServiceContainer services) {
-            var platform = services.GetService<IOSPlatform>();
-            var logger = services.GetService<ILogger>();
-
+        public static string GetCacheFolder(IOSPlatform platform, ILogger logger = null) {
```
CTrando: Was logging every time I called it, so wanted to have an option not to log with the optional parameter.

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
```diff
@@ -103,6 +100,12 @@ internal static class CacheFolders {
                 .Replace('/', '_').Replace('+', '-');
         }
 
+        /// <summary>
+        /// Given a path to a zip file, extracts it to a cache directory and returns the file path to that directory
+        /// </summary>
+        public static string GetZipCacheRoot(IOSPlatform platform)
+            => Path.Combine(GetCacheFolder(platform), "Zips/Eggs");
```
CTrando: Can use a better name if needed

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
```diff
@@ -190,6 +192,19 @@ public IPythonModule GetSpecializedModule(string name)
...
+            _log?.Log(TraceEventType.Information, "Aliased search paths:");
+            var zipRoot = CacheFolders.GetZipCacheRoot(_services.GetService<IOSPlatform>());
+            foreach (var s in zipPaths) {
+                // Alias ../test.zip => ../test folder
+                var aliasedPath = Path.Combine(zipRoot, Path.GetFileNameWithoutExtension(s.Path));
+                if(!Directory.Exists(aliasedPath)) {
+                    _fs.ExtractZip(s.Path, aliasedPath);
+                }
+                // User path has the .egg or .zip path, redirect it to have the new alias path
+                _userPaths = _userPaths.Add(aliasedPath);
```
CTrando: I add to userPaths, would there be a difference in adding to interpreter path if the zip came from there?

CTrando: Now that I think about it the aliased path is always under the analysis cache path, so should I add it to interpreter paths?

src/LanguageServer/Test/ImportsTests.cs
```diff
@@ -830,5 +832,39 @@ import module2
...
+
+            var uriPath = Path.Combine(root, "egg_user.py");
+            var code = await File.ReadAllTextAsync(uriPath);
+            var moduleUri = TestData.GetTestSpecificUri(uriPath);
+            var module = rdt.OpenDocument(moduleUri, code);
+
+            await analyzer.WaitForCompleteAnalysisAsync();
+            var analysis = await module.GetAnalysisAsync(-1);
+            analysis.Should().HaveVariable("i").OfType(BuiltinTypeId.Int);
+        }
```
CTrando: So these are for adding the .zip and .egg files on the user search paths, I'm not exactly sure how to test for interpreter search paths. @AlexanderSher could you give me some suggestions?

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
AlexanderSher: `Path.Combine` has overload with several parameters: `Path.Combine(GetCacheFolder(platform), "Zips", "Eggs")`

CTrando: I'm doing Zips/Eggs to mean a folder that contains zips and eggs. Maybe `ZipsAndEggs` is better?

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
MikhailArkhipov: Caller then should be logging. I.e. we output cache folder every time so user knows what to clear if needed.

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
```diff
@@ -103,6 +100,12 @@ internal static class CacheFolders {
                 .Replace('/', '_').Replace('+', '-');
         }
 
+        /// <summary>
+        /// Given a path to a zip file, extracts it to a cache directory and returns the file path to that directory
+        /// </summary>
+        public static string GetZipCacheRoot(IOSPlatform platform)
```
MikhailArkhipov: I think we should try and make it work without explicit extraction and changing paths. Ex files can be enumerated and read directly via `https://docs.microsoft.com/en-us/dotnet/api/system.io.compression.ziparchiveentry?view=netframework-4.8`.

MikhailArkhipov:
I think we should try and make it work without explicit extraction and changing paths. Ex files can be enumerated and read directly via https://docs.microsoft.com/en-us/dotnet/api/system.io.compression.ziparchiveentry?view=netframework-4.8. See ZipArchiveEntry.Open

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
```diff
@@ -190,6 +192,19 @@ public IPythonModule GetSpecializedModule(string name)
...
             }
+
+            _log?.Log(TraceEventType.Information, "Aliased search paths:");
+            var zipRoot = CacheFolders.GetZipCacheRoot(_services.GetService<IOSPlatform>());
+            foreach (var s in zipPaths) {
+                // Alias ../test.zip => ../test folder
+                var aliasedPath = Path.Combine(zipRoot, Path.GetFileNameWithoutExtension(s.Path));
+                if(!Directory.Exists(aliasedPath)) {
+                    _fs.ExtractZip(s.Path, aliasedPath);
+                }
```
MikhailArkhipov: How would that work when zips have same names in different modules?

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
CTrando: It was being called twice in initialization and outputting the same information so I had one call log and the other was silent.

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
CTrando: Good point

CTrando:
Should I add support for goto definitions in egg files and zips?

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
CTrando: I think it's much easier if we extract them, but I can try and make it work in memory

MikhailArkhipov:
Goto Def should work automatically since root definition should have been loaded and analyzed. Find all references only scans user files in workspace and symbol it is searching for is already in analysis and hence the module is in RDT (it is based on goto def anyway). So I dont think anything else is needed if module is loaded.

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
MikhailArkhipov: No it is not easier. We'd have to manage the lifetime of extracted files - what if user places new zip with new files? What if there are different versions with the same name but different content? What if we fail to extract all for whatever reasons and now user is stuck with invalid cache and does not know why imports are not getting resolved? What if user has search paths that override what we add or we stomp on their imports by importing something from cache b/c it appeared earlier in search paths? Modifying search paths is pretty risky proposition.

CTrando:
I looked into it and goto does not work. What happens is that because everything in the egg/zip file is analyzed in memory, there is no concrete implementation of it. It looks for the uri of the module, but because the module is in memory it has no URI (technically it does but it is inaccessible because it is inside the zip/egg file). Brett suggested creating a temporary file with the contents of the zip file for the purposes of goto definition. 

I plan to make a new module class `ZippedModule` that will take care of resolving `FilePath` and separating it into `zipPath` (the path to the zip folder) and `zipFilePath` the path to the file inside the zip folder.

MikhailArkhipov:
They should have URI reflecting their place on disk and in zip, like `root/a/b/zip/c/d/x.py`. It may need special service inside `FileSystem` to make sure it can load content from inside zip.

MikhailArkhipov:
I don't think that would work. How are we going to handle CompiledModule or StubModule if it is inside zip? Can't derive from both ZipModule and StubModule. I think it is sufficient to have `FileSystem.ReadTextWithRetry` to handle zipped content. Basically transparently, like it works in Windows Explorer for example. Zip just opens like a folder.

CTrando:
In DefinitionSource, we return a Reference with a Uri. That Uri (to my knowledge) must be openable by vscode. If the uri contains a zip folder, e.g `test.zip/a.py`, then vscode cannot open the file. I just tested, and vscode cannot open zip folders. 

That means we cannot do go to definition and return a Uri inside a zip folder. 

For the second comment:
    1. That's a good point, and I'm willing to go along with it; however, I would like some kind of way to get both the `zipPath` and `zipFilePath` strings in an object. I wanted a `ZippedModule` because I thought it would very nice to have `FilePath` return them concatenated, and when calling `ZippedModule.LoadContent` it would already have access to `zipPath` and I wouldn't have to do the parsing I did in `FileSystem.ReadTextWithRetry`. I think the parsing I did is very hacky. The second idea I had was to change `FilePath` to an interface that I could implement specifically for zipped modules, but it seemed too big a change. 
    2. They do have proper paths. What I'm doing now is I am not editing user/interpreter search paths. What I'm doing is when I encounter a zip/egg file is that I open them up and merge their contents together.
e.g
```
test.zip
    test
        a.py
        b.py
```
Gets added to the path resolver as `test.zip/test/a` and `test.zip/test/b`. 
Now take for example we do:
```
import test.a
```
The path resolver will see that we have a user/interpreter path `test.zip`, and look for `test.zip/test/a`, which we have added already, so the import will be correctly resolved. Even with relative imports, we will be able to analyze within the same zip file (perhaps even go outside of the zip file). I've already tested this on torchvision (the library that prompted this bug report).

MikhailArkhipov:
Needs merge and review from @AlexanderSher since it is all about imports.

src/Analysis/Ast/Impl/Modules/Resolution/ModuleResolutionBase.cs
```diff
@@ -105,11 +103,24 @@ public IPythonModule GetImportedModule(string name)
...
+                    PathResolver.TryAddModulePath(moduleFile.FullName, moduleFile.Length, false, out _);
+                }
+
+                if (PathUtils.TryGetZipFilePath(root, out var zipFilePath, out var _) && File.Exists(zipFilePath)) {
+                    foreach (var moduleFile in PathUtils.EnumerateZip(zipFilePath)) {
+                        if (!PathUtils.PathStartsWith(moduleFile.FullName, "EGG-INFO")) {
+                            PathResolver.TryAddModulePath(
+                                Path.Combine(zipFilePath,
+                                PathUtils.NormalizePath(moduleFile.FullName)),
+                                moduleFile.Length, false, out _
```
AlexanderSher: So what would be the `fullModuleName` returned by the `TryAddModulePath`?

CTrando: `fullModuleName` returns the empty string, not sure if that's what is supposed to happen

AlexanderSher: Is that an expected full module name?

CTrando: Sorry I should have clarified that a root was `test.zip/test`. I ran the test on the library `simplejson` and for files like `simplejson/decoder.py` it returns `simplejson.decoder` and for `simplejson/tests/test_bitsize_int` it gives `simplejson.tests.tesT_bitsize_int` etc, so I believe this is correct.

AlexanderSher: Ok. Then let's add an import test to ensure that import of the module from egg is resolved correctly.

CTrando: Yes I did that in:
 https://github.com/microsoft/python-language-server/pull/1477/files#diff-fc769198933ce6a497889cd518874a9dR897-R928

It loads up the simplejson library and checks to see if some variables have members

MikhailArkhipov:
Merging into db first

PR #1478, opened 2019-08-23
Fix couple of concurrency issues

AlexanderSher:


src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
```diff
@@ -139,9 +140,10 @@ public Task<IDocumentAnalysis> GetAnalysisAsync(CancellationToken cancellationTo
                 _analysisDependencies = null;
                 UpdateAnalysisTcs(version);
                 _previousAnalysis = analysis;
+                tcs = _analysisTcs;
             }
 
-            _analysisTcs.TrySetResult(analysis);
+            tcs.TrySetResult(analysis);
```
AlexanderSher: `_analysisTcs` can be changed by another call of `UpdateAnalysisTcs` with a newer version. Shouldn't affect the old one.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerSession.cs
```diff
@@ -128,10 +128,6 @@ internal sealed class PythonAnalyzerSession {
             }
 
             var stopWatch = Stopwatch.StartNew();
-            foreach (var affectedEntry in _walker.AffectedValues) {
-                affectedEntry.Invalidate(Version);
```
AlexanderSher: Old code that doesn't required anymore. Also causes the issues invalidating library modules that have been analyzed by previous session.

PR #1492, opened 2019-08-28
Handle on-the-fly search path configuration changes, read search paths from config directly

### LINKED ISSUE #915: Throws an exception if workspace/didChangeConfiguration settings is empty/null
It tries to get `python` key from settings even if settings is null. Which makes `Newtonsoft.Json` throw an exception (`Newtonsoft.Json.Linq.JToken.get_Item(Object key) InvalidOperationException`).

### LINKED ISSUE #1388: Read search paths from configuration in didChangeConfiguration, update on the fly
With #1289, we can now do what I mentioned in #794:

- Don't read searchPaths in initialization options (note: need to preserve backward compat for a time).
- Accept our own list (or lists) of paths in the main configuration sent in didChangeConfiguration.
- Read PYTHONPATH from the environment.
- Combine info ourselves.

This would mean:

- We have one fewer thing "required" in our initialization (#469).
- We can update the search paths and reload on a configuration change (#507).

But, before doing this, we'll need to settle on our choices of configuration options, as `python.autoComplete.extraPaths` is likely not what we'd want to call it.

jakebailey:
Fixes #1388.
Fixes #915. (Null check added.)
Updates #469. (Less needed from `initializationOptions`, `PYTHONPATH` potential.)

User configured search paths are now pulled from the user configuration (and on the fly), rather than solely through the initialization options.

To do this, the LS initialization logic has been moved from `initialize` RPC call into the `initialized` notification handler. There, we can ask the editor for the python configuration, and then use the real configuration to set the initial search paths. If the configuration is changed later, then it will be re-processed, and a reload will be triggered if it changes (just like path watching).

The old `python.autoComplete.extraPaths` will still work. `python.analysis.searchPaths` and `python.analysis.usePYTHONPATH` have been added as a newer interface. Also in this change is a fix which makes `watchSearchPaths` accessible again, as `python.analysis.watchSearchPaths`; previously it was not accessible.

Config items:
- `searchPaths` in the initialization options - Old, last resort.
- `python.autoComplete.extraPaths` - Old, but still read.
- `python.analysis.searchPaths` - New, replacement for `extraPaths`. If present, extraPaths is ignored entirely.
- `python.analysis.usePYTHONPATH` - New, true (default) if `PYTHONPATH` should be added to the end of `searchPaths`/`extraPaths` (which the extension did)
- `python.analysis.watchSearchPaths` - Fixed, true (default) if library search paths should be watched for file changes.


jakebailey:
I can see a way to handle the issues with the configuration coming in too late: The server can make a `workspace/configuration` call back to the editor and ask for the configuration. However, the server is not allowed to make any calls until `initialize` returns, which means that any initialization RPC calls back to the client need to happen later, which the spec says to do in the `initialized` handler. It's very likely that what we're doing in the `initialize` handler can be moved later.

That might be too invasive to do in this change, but, if it's done, the server can just ask the editor for the config, pre-populate the settings (before it ever gets `workspace/didChangeConfiguration`), and initialize based on that.

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
```diff
@@ -41,14 +41,14 @@ internal sealed class MainModuleResolution : ModuleResolutionBase, IModuleManage
         private readonly IUIService _ui;
         private IRunningDocumentTable _rdt;
 
-        private ImmutableArray<string> _userPaths = ImmutableArray<string>.Empty;
+        private ImmutableArray<string> _userConfiguredPaths;
 
-        public MainModuleResolution(string root, IServiceContainer services)
+        public MainModuleResolution(string root, IServiceContainer services, ImmutableArray<string> userConfiguredPaths = default)
```
AlexanderSher: I'd rather not have this parameter optional.

jakebailey: Oops, that was a leftover. Removed.

Issue #1428, opened 2019-08-12
Add periodic GC-compact when user is idle


jakebailey:
To help out in cases like #1298 and #1408, we should periodically run a full GC and compact when the user is idle and no analysis is running. Doing it on a big session or a reload is helpful, but small, incremental changes do not benefit.

Opening so we don't forget.

PR #1496, opened 2019-08-29
Fix #1156: Unresolved import for builtin packages after reload

### LINKED ISSUE #1156: Unresolved import for builtin packages after reload

AlexanderSher:


src/Analysis/Ast/Impl/Analyzer/PythonInterpreter.cs
```diff
@@ -92,6 +83,24 @@ public sealed class PythonInterpreter : IPythonInterpreter {
...
+            get {
+                lock (_lock) {
+                    var type = _unknownType;
+                    if (type != null) {
+                        return type;
+                    }
+
+                    _unknownType = new PythonType("Unknown", new Location(_moduleResolution.BuiltinsModule), string.Empty);
+                    _builtinTypes[BuiltinTypeId.Unknown] = _unknownType;
+                    return _unknownType;
```
AlexanderSher: `Unknown` type switched to lazy initialization

src/Analysis/Core/Impl/Interpreter/PythonLibraryPath.cs
```diff
@@ -109,31 +109,30 @@ public sealed class PythonLibraryPath : IEquatable<PythonLibraryPath> {
         /// <param name="library">Root of the standard library.</param>
         /// <returns>A list of search paths for the interpreter.</returns>
         /// <remarks>New in 2.2, moved in 3.3</remarks>
-        private static List<PythonLibraryPath> GetDefaultSearchPaths(IFileSystem fs, string library) {
-            var result = new List<PythonLibraryPath>();
+        private static ImmutableArray<PythonLibraryPath> GetDefaultSearchPaths(IFileSystem fs, string library) {
```
jakebailey: Not entirely related to this PR, but I feel like we should come up with some agreement about which list types we're going to use across the board so we don't end up switching them back and forth between PRs...

jakebailey:
When I do the same memory testing I did for #1402, the memory usage after a big session or a reload doesn't seem to go down very much (or at all), and requires something else to happen to do another GC/compact for the memory to go down again. Maybe there's some timing problem where the new code is a bit faster and things aren't ready to be compacted yet, or something, but it might hurt memory usage. (Another case #1428 would help.)

src/Analysis/Ast/Impl/Analyzer/PythonInterpreter.cs
```diff
@@ -106,31 +115,34 @@ public sealed class PythonInterpreter : IPythonInterpreter {
...
-                    return res;
+                if (_builtinTypes.TryGetValue(id, out var type) && type != null) {
+                    return type;
                 }
 
-                var bm = ModuleResolution.BuiltinsModule;
-                var typeName = id.GetTypeName(LanguageVersion);
-                if (typeName != null) {
-                    res = bm.GetMember(typeName) as IPythonType;
+                if (id == BuiltinTypeId.Unknown) {
```
MikhailArkhipov: Might as well move this before dictionary check

jakebailey:
Just an example of what I'm describing:

image

1. `pip install numpy`, reloads, memory goes up.
2. `pip install -e .` (small package, no imports), reloads and redoes everything, memory is the same.
3. `pip unintsall numpy`, memory usage goes down, but not all the way.
4. `pip uninstall mypackage`, memory usage goes down to the baseline.

There are variations where the initial install doesn't drop after its session GC/compact, and some where the reload GC/compact doesn't have any affect either.

src/Analysis/Core/Impl/Interpreter/PythonLibraryPath.cs
MikhailArkhipov: I am OK either way but I tend to use interfaces on public surface. Is there a benefit of `Immutable` over `IReadOnly`?

AlexanderSher: There are 3 issues with `IReadOnlyList<T>`:
1. It is read-only, but not immutable. If an object is provided to the outer context, but owner still has a reference to the object, the is no way to synchronize it anymore. E.g.:
    ```cs
    private List<SomeType> _someList;
    public IReadOnlyList<SomeType> SomeList=> _someList;
    ```
    If any code gets reference to the list through `SomeList`, it still can be changed by the owner, concurrently to the reader.
2. Adding elements. With `ImmutableArray`, adding elements create new instance of it, but it is as cheap as adding to `List<T>`. For `IReadOnlyList<T>`, it requires to create a copy of the list.
3. Enumerator. For any interface, enumerator is an object. For `ImmutableArray<T>` and `List<T>` it would be a struct. This becomes a problem when we have `foreach` loops in analysis where enumerators are promoted to Gen2.

jakebailey:
Here's what happens when I repeat the above but modify the ForceGC function to wait a bit:

image

```csharp
private static void ForceGCIfNeeded(ILogger logger, int originalRemaining, int remaining, bool force) {
	if (force || originalRemaining - remaining > 1000) {
		Task.Run(async () => {
			await Task.Delay(500);
			logger?.Log(TraceEventType.Verbose, "Forcing full garbage collection and heap compaction.");
			GCSettings.LargeObjectHeapCompactionMode = GCLargeObjectHeapCompactionMode.CompactOnce;
			GC.Collect();
		}).DoNotWait();
	}
}
```

PR #1497, opened 2019-08-30
Improve handling of circular and imported class bases

### LINKED ISSUE #1495: Stack overflow when importing GPy
# Environment data
VS Code version: 1.37.1
Microsoft Python Language Server version: 0.3.66.0
Extension version (available under the Extensions sidebar):2019.8.30787
OS and version: MacOs Mojave 10.14.5 (18F132)
Python version (& distribution if applicable, e.g. Anaconda): Anaconda 3.7.3
Type of virtual environment used (N/A | venv | virtualenv | conda | ...): conda
Relevant/affected Python packages and their versions: GPy https://sheffieldml.github.io/GPy/
Jedi or Language Server? (i.e. what is "python.jediEnabled" set to; more info microsoft/vscode-python#3977): python.jediEnabled：false

# Expected behaviour
Language server doesn't hang on 'Analyzing in background'

# Actual behaviour
CPU usage is *not* high but process keeps terminating due to StackOverflowException. Then the 'The Python Tools server crashed 5 times in the last 3 minutes. The server will not be restarted.' warning appears . I've isolated the problem to be due to importing the GPy package (problem goes away if I comment out the imports). I've also tried adding the '# noqa' tag after the imports and it doesn't work. Is it possible to allow users to specify packages for the language server to ignore?

# Steps to reproduce:
Open a blank python file and import gpy

# Logs
User belongs to experiment group 'AlwaysDisplayTestExplorer - experiment'
User belongs to experiment group 'ShowPlayIcon - control'
User belongs to experiment group 'ShowExtensionSurveyPrompt - control'
> conda --version
> pyenv root
> python3.7 -c "import sys;print(sys.executable)"
> python3.6 -c "import sys;print(sys.executable)"
> python3 -c "import sys;print(sys.executable)"
> python2 -c "import sys;print(sys.executable)"
> python -c "import sys;print(sys.executable)"
> ~/miniconda3/envs/ml3/bin/python -c "import sys;print(sys.executable)"
> conda info --json
> conda env list
> conda info --json
> ~/miniconda3/envs/ml3/bin/python -m pylint --errors-only --enable=W0614 "--generated-members=numpy.* ,torch.* ,cv2.* , cv.*" --msg-template='{line},{column},{category},{symbol}:{msg}' --reports=n --output-format=text ~/dev/workspace/test.py
cwd: ~/dev/workspace
> ~/miniconda3/envs/ml3/bin/python -m pylint --errors-only --enable=W0614 "--generated-members=numpy.* ,torch.* ,cv2.* , cv.*" --msg-template='{line},{column},{category},{symbol}:{msg}' --reports=n --output-format=text ~/dev/workspace/test.py
cwd: ~/dev/workspace
> conda env list
> ~/miniconda3/envs/ml3/bin/python -m pylint --errors-only --enable=W0614 "--generated-members=numpy.* ,torch.* ,cv2.* , cv.*" --msg-template='{line},{column},{category},{symbol}:{msg}' --reports=n --output-format=text ~/dev/workspace/test.py
cwd: ~/dev/workspace
> ~/miniconda3/envs/ml3/bin/python -m pylint --errors-only --enable=W0614 "--generated-members=numpy.* ,torch.* ,cv2.* , cv.*" --msg-template='{line},{column},{category},{symbol}:{msg}' --reports=n --output-format=text ~/dev/workspace/test.py
cwd: ~/dev/workspace
[Info  - 2:51:34 PM] Analysis cache path: /Users/vmasrani/Library/Caches/Microsoft/Python Language Server
[Info  - 2:51:34 PM] GetCurrentSearchPaths /Users/vmasrani/miniconda3/envs/ml3/bin/python
[Info  - 2:51:34 PM] Interpreter search paths:
[Info  - 2:51:34 PM]     /private/var/folders/k4/mvr6htsd0kg3wvx5vzttctg00000gr/T/ymeysyn1.vjb
[Info  - 2:51:34 PM]     /Users/vmasrani/miniconda3/envs/ml3/lib/python3.7
[Info  - 2:51:34 PM]     /Users/vmasrani/miniconda3/envs/ml3/lib/python3.7/lib-dynload
[Info  - 2:51:34 PM]     /Users/vmasrani/.python
[Info  - 2:51:34 PM]     /Users/vmasrani/miniconda3/envs/ml3/lib/python3.7/site-packages
[Info  - 2:51:34 PM] User search paths:
[Info  - 2:51:38 PM] Microsoft Python Language Server version 0.3.66.0
[Info  - 2:51:38 PM] Initializing for /Users/vmasrani/miniconda3/envs/ml3/bin/python
> conda info --json
Process is terminating due to StackOverflowException.
[Info  - 2:52:04 PM] Connection to server got closed. Server will restart.
[Info  - 2:52:04 PM] Analysis cache path: /Users/vmasrani/Library/Caches/Microsoft/Python Language Server
[Info  - 2:52:04 PM] GetCurrentSearchPaths /Users/vmasrani/miniconda3/envs/ml3/bin/python
[Info  - 2:52:05 PM] Interpreter search paths:
[Info  - 2:52:05 PM]     /private/var/folders/k4/mvr6htsd0kg3wvx5vzttctg00000gr/T/zcsujvkk.oig
[Info  - 2:52:05 PM]     /Users/vmasrani/miniconda3/envs/ml3/lib/python3.7
[Info  - 2:52:05 PM]     /Users/vmasrani/miniconda3/envs/ml3/lib/python3.7/lib-dynload
[Info  - 2:52:05 PM]     /Users/vmasrani/.python
[Info  - 2:52:05 PM]     /Users/vmasrani/miniconda3/envs/ml3/lib/python3.7/site-packages
[Info  - 2:52:05 PM] User search paths:
[Info  - 2:52:08 PM] Microsoft Python Language Server version 0.3.66.0
[Info  - 2:52:08 PM] Initializing for /Users/vmasrani/miniconda3/envs/ml3/bin/python
Process is terminating due to StackOverflowException.
[Info  - 2:52:44 PM] Connection to server got closed. Server will restart.

MikhailArkhipov:
Fixes #1495 

- Since we don't handle reassignments, improve handling of imported bases like 
```python
from Base import A, B

class A(A, B):
    def methodA(self, x):
        return x

class C(B): ...

class B(A):
    def methodB(self, x):
        return x
```
by declaring imported variables in regular collection as well as in special `imported` collection so they can be found when we need to resolve class base.
- Decide on declaration or import depending on location
- Implement deep filtering of circular base chains
- Tests

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Scopes.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Scopes.cs
@@ -47,20 +47,23 @@ namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
             => DeclareVariable(name, value, source, GetLocationOfName(location), overwrite);
 
         public void DeclareVariable(string name, IMember value, VariableSource source, Location location, bool overwrite = true) {
+            if (source == VariableSource.Import) {
+                // Duplicate declaration so if variable gets overwritten it can still be retrieved. Consider:
+                //    from X import A
+                //    class A(A): ...
+                CurrentScope.DeclareImported(name, value, location);
+            }
+
             var member = GetInScope(name);
             if (member != null && !overwrite) {
-                if (source == VariableSource.Import) {
-                    // Duplicate declaration so if variable gets overwritten it can still be retrieved. Consider:
-                    //    from X import A
-                    //    class A(A): ...
-                    CurrentScope.DeclareImported(name, value, location);
-                }
                 return;
             }
+            
             if (source == VariableSource.Import && value is IVariable v) {
                 CurrentScope.LinkVariable(name, v, location);
                 return;
             }
+            
             if (member != null) {
                 if (!value.IsUnknown()) {
                     CurrentScope.DeclareVariable(name, value, source, location);
--- a/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Scopes.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Scopes.cs
@@ -49,6 +49,12 @@ namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
         public void DeclareVariable(string name, IMember value, VariableSource source, Location location, bool overwrite = true) {
             var member = GetInScope(name);
             if (member != null && !overwrite) {
+                if (source == VariableSource.Import) {
+                    // Duplicate declaration so if variable gets overwritten it can still be retrieved. Consider:
+                    //    from X import A
+                    //    class A(A): ...
+                    CurrentScope.DeclareImported(name, value, location);
+                }
                 return;
             }
             if (source == VariableSource.Import && value is IVariable v) {
--- a/src/Analysis/Ast/Impl/Extensions/EvaluatorExtensions.cs
+++ b/src/Analysis/Ast/Impl/Extensions/EvaluatorExtensions.cs
@@ -27,9 +27,9 @@ namespace Microsoft.Python.Analysis {
 
         public static IMember LookupImportedNameInScopes(this IExpressionEvaluator eval, string name, out IScope scope) {
             scope = null;
-            foreach(var s in eval.CurrentScope.EnumerateTowardsGlobal) {
+            foreach (var s in eval.CurrentScope.EnumerateTowardsGlobal) {
                 var v = s.Imported[name];
-                if(v != null) {
+                if (v != null) {
                     scope = s;
                     return v.Value;
                 }
--- a/src/Analysis/Ast/Impl/Extensions/EvaluatorExtensions.cs
+++ b/src/Analysis/Ast/Impl/Extensions/EvaluatorExtensions.cs
@@ -25,18 +25,6 @@ namespace Microsoft.Python.Analysis {
         public static IMember LookupNameInScopes(this IExpressionEvaluator eval, string name, LookupOptions options = LookupOptions.Normal)
             => eval.LookupNameInScopes(name, out _, out _, options);
 
-        public static IMember LookupImportedNameInScopes(this IExpressionEvaluator eval, string name, out IScope scope) {
-            scope = null;
-            foreach (var s in eval.CurrentScope.EnumerateTowardsGlobal) {
-                var v = s.Imported[name];
-                if (v != null) {
-                    scope = s;
-                    return v.Value;
-                }
-            }
-            return null;
-        }
-
         public static IDisposable OpenScope(this IExpressionEvaluator eval, IPythonClassType cls)
          => eval.OpenScope(cls.DeclaringModule, cls.ClassDefinition);
         public static IDisposable OpenScope(this IExpressionEvaluator eval, IPythonFunctionType ft)
--- a/src/Analysis/Ast/Impl/Extensions/EvaluatorExtensions.cs
+++ b/src/Analysis/Ast/Impl/Extensions/EvaluatorExtensions.cs
@@ -24,6 +24,19 @@ namespace Microsoft.Python.Analysis {
             => eval.LookupNameInScopes(name, out scope, out _, options);
         public static IMember LookupNameInScopes(this IExpressionEvaluator eval, string name, LookupOptions options = LookupOptions.Normal)
             => eval.LookupNameInScopes(name, out _, out _, options);
+
+        public static IMember LookupImportedNameInScopes(this IExpressionEvaluator eval, string name, out IScope scope) {
+            scope = null;
+            foreach(var s in eval.CurrentScope.EnumerateTowardsGlobal) {
+                var v = s.Imported[name];
+                if(v != null) {
+                    scope = s;
+                    return v.Value;
+                }
+            }
+            return null;
+        }
+
         public static IDisposable OpenScope(this IExpressionEvaluator eval, IPythonClassType cls)
          => eval.OpenScope(cls.DeclaringModule, cls.ClassDefinition);
         public static IDisposable OpenScope(this IExpressionEvaluator eval, IPythonFunctionType ft)
--- a/src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
+++ b/src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
@@ -124,7 +124,7 @@ namespace Microsoft.Python.Analysis.Types {
                     }
 
                     // Set specific class bases 
-                    classType.SetBases(specificBases.Concat(newBases), args.Eval);
+                    classType.SetBases(specificBases.Concat(newBases), args.Eval.CurrentScope);
                     // Now that parameters are set, check if class is generic
                     classType._parameters = classType._genericParameters.Values.Distinct().OfType<IGenericTypeParameter>().ToList();
                     classType.DecideGeneric();
--- a/src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
+++ b/src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
@@ -61,23 +61,19 @@ namespace Microsoft.Python.Analysis.Types {
         /// </summary>
         public IPythonType CreateSpecificType(IArgumentSet args) {
             lock (_genericParameterLock) {
-                var genericTypeParameters = GetTypeParameters();
+                var newGenericTypeParameters = GetTypeParameters();
                 var newBases = new List<IPythonType>();
 
                 // Get map of generic type parameter to specific type - fill in what type goes to what
                 // type parameter T -> int, U -> str, etc.
-                var genericTypeToSpecificType = GetSpecificTypes(args, genericTypeParameters, newBases);
+                var genericTypeToSpecificType = GetSpecificTypes(args, newGenericTypeParameters, newBases);
 
                 var classType = new PythonClassType(BaseName, new Location(DeclaringModule));
                 // Storing generic parameters allows methods returning generic types 
                 // to know what type parameter returns what specific type
-                classType.StoreGenericParameters(genericTypeParameters, genericTypeToSpecificType);
+                classType.StoreGenericParameters(GenericParameters, newGenericTypeParameters, genericTypeToSpecificType);
 
                 // Set generic name
-                if (!classType._genericParameters.IsNullOrEmpty()) {
-                    classType._genericName = CodeFormatter.FormatSequence(BaseName, '[', classType._genericParameters.Values);
-                }
-
                 // Locking so threads can only access class after it's been initialized
                 // Store generic parameters first so name updates correctly, then check if class type has been cached
                 _specificTypeCache = _specificTypeCache ?? new Dictionary<string, PythonClassType>();
@@ -106,8 +102,8 @@ namespace Microsoft.Python.Analysis.Types {
                     // Get list of bases that are generic but not generic class parameters, e.g A[T], B[T] but not Generic[T1, T2]
                     var genericTypeBases = bases.Except(genericClassParameters).OfType<IGenericType>().Where(g => g.IsGeneric).ToArray();
 
-                    // Removing all generic bases, and will only specialize genericTypeBases, remove generic class paramters entirely
-                    // We remove generic class paramters entirely because the type information is now stored in GenericParameters field
+                    // Removing all generic bases, and will only specialize genericTypeBases, remove generic class parameters entirely
+                    // We remove generic class parameters entirely because the type information is now stored in GenericParameters field
                     // We still need generic bases so we can specialize them 
                     var specificBases = bases.Except(genericTypeBases).Except(genericClassParameters).ToList();
 
@@ -215,7 +211,7 @@ namespace Microsoft.Python.Analysis.Types {
                     // for the copy constructor. Consider 'class A(Generic[K, V], Mapping[K, V])'
                     // constructed as 'd = {1:'a', 2:'b'}; A(d)'. Here we look through bases
                     // and see if any matches the builtin type id. For example, Mapping or Dict
-                    // will have BultinTypeId.Dict and we can figure out specific types from
+                    // will have BuiltinTypeId.Dict and we can figure out specific types from
                     // the content of the collection.
                     var b = _bases.OfType<IGenericType>().Where(g => g.IsGeneric).FirstOrDefault(x => x.TypeId == type.TypeId);
                     if (b != null && !b.Parameters.IsNullOrEmpty()) {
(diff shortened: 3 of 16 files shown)


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Scopes.cs
```diff
@@ -49,6 +49,12 @@ public void DeclareVariable(string name, IMember value, VariableSource source, N
...
             if (member != null && !overwrite) {
+                if (source == VariableSource.Import) {
+                    // Duplicate declaration so if variable gets overwritten it can still be retrieved. Consider:
+                    //    from X import A
+                    //    class A(A): ...
+                    CurrentScope.DeclareImported(name, value, location);
+                }
                 return;
             }
             if (source == VariableSource.Import && value is IVariable v) {
```
CTrando: In the conditional below, shouldn't you also declare anything with variable source import as an imported variable to be consistent? Otherwise the collection of imported variables only consists of those that would override already existing variables and not include those that just happened to be declared in other modules.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -30,7 +31,7 @@
 
 namespace Microsoft.Python.Analysis.Types {
     [DebuggerDisplay("Class {Name}")]
-    internal partial class PythonClassType : PythonType, IPythonClassType, IGenericType, IEquatable<IPythonClassType> {
```
CTrando: Did we decide to make class types not generic anymore?

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -192,12 +191,15 @@ internal PythonClassType(string name, Location location)
...
+        internal void SetBases(IEnumerable<IPythonType> bases, IExpressionEvaluator eval = null) {
             if (_bases != null) {
                 return; // Already set
             }
 
-            bases = bases != null ? bases.Where(b => !b.GetPythonType().IsUnknown()).ToArray() : Array.Empty<IPythonType>();
+            // Consider
+            //    from X import A
+            //    class A(A): ...
+            bases = DisambiguateBases(bases, eval).ToArray();
```
CTrando: perhaps just a name change to `ResolveBases` or maybe even consider just putting these methods in `ClassEvaluator` because much of the work that it is doing is similar to the `ClassEvaluator.ProcessBases`

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -281,5 +284,55 @@ internal PythonClassType(string name, Location location)
...
+                if (!b.IsUnknown()) {
+                    newBases.Add(b);
+                }
+            }
+            return FilterCircularBases(newBases);
+        }
+
+        private IEnumerable<IPythonType> FilterCircularBases(IEnumerable<IPythonType> bases) {
+            // Inspect each base chain and exclude bases that chains to this class.
+            foreach(var b in bases.Where(x => !Equals(x))) {
```
CTrando: So this is dropping bases that have chains that lead back to the original class type. Didn't we want to save those bases to process them later? If we don't, then do we still reentrancy guards?

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -281,5 +284,55 @@ internal PythonClassType(string name, Location location)
...
+            var newBases = new List<IPythonType>();
+            foreach (var b in bases) {
+                var imported = eval.LookupImportedNameInScopes(b.Name, out _);
+                if (imported is IPythonType importedType) {
+                    // Variable with same name as the base was imported.
+                    // If there is also a local declaration, we need to figure out which one wins.
+                    var declared = eval.LookupNameInScopes(b.Name, out var scope);
+                    if (declared != null && scope != null) {
+                        var v = scope.Variables[b.Name];
+                        if (v.Source != VariableSource.Import && v.Value is IPythonClassType cls && cls.Location.IndexSpan.Start >= Location.IndexSpan.Start) {
```
CTrando: Maybe consider an extension method for checking if a declaration/assignment occurs after a certain point, I made a similar call in `AssignmentHandler` so maybe it's getting popular enough

CTrando: Consider putting this as a method, it's sort of confusing to read. Maybe `PythonClassType.DeclaredBefore(ILocatedMember m)` or something so you could rewrite it as:
```
if(v.Source != VariableSource.Import && DeclaredBefore(v.Value as IPythonClassType)) ...
```
or something similar.

You could rename `v` to `local` to further clarify `v` is the class declared inside the current module.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Scopes.cs
MikhailArkhipov: Probably... I guess like in `base.py`
```python
class A: ...
a = A
```
and in `Main.py`
```python
from Base import a
class a(a): ...
```

src/Analysis/Ast/Impl/Types/PythonClassType.cs
jakebailey: Yeah, I don't think this should be removed. Probably just moved to `PythonClassType.Generics.cs` where this interface is being implemented.

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
```diff
@@ -23,13 +23,14 @@
 
 namespace Microsoft.Python.Analysis.Types {
     internal partial class PythonClassType {
```
jakebailey: Add IGenericType here, since the interface is implemented below?

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -281,5 +284,55 @@ internal PythonClassType(string name, Location location)
         public bool Equals(IPythonClassType other)
             => Name == other?.Name && DeclaringModule.Equals(other?.DeclaringModule);
 
+        private IEnumerable<IPythonType> DisambiguateBases(IEnumerable<IPythonType> bases, IExpressionEvaluator eval) {
+            if (bases == null) {
+                return Array.Empty<IPythonType>();
```
jakebailey: `Enumerable.Empty` instead of `Array.Empty`, maybe.

src/Analysis/Ast/Impl/Extensions/EvaluatorExtensions.cs
```diff
@@ -24,6 +24,19 @@ public static IMember LookupNameInScopes(this IExpressionEvaluator eval, string
             => eval.LookupNameInScopes(name, out scope, out _, options);
         public static IMember LookupNameInScopes(this IExpressionEvaluator eval, string name, LookupOptions options = LookupOptions.Normal)
             => eval.LookupNameInScopes(name, out _, out _, options);
+
+        public static IMember LookupImportedNameInScopes(this IExpressionEvaluator eval, string name, out IScope scope) {
+            scope = null;
+            foreach(var s in eval.CurrentScope.EnumerateTowardsGlobal) {
```
jakebailey: Nit: formatting of `foreach(` and `if(` in this function/.

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
```diff
@@ -67,10 +68,10 @@ internal partial class PythonClassType {
                 // type parameter T -> int, U -> str, etc.
                 var genericTypeToSpecificType = GetSpecificTypes(args, genericTypeParameters, newBases);
 
-                PythonClassType classType = new PythonClassType(BaseName, new Location(DeclaringModule));
+                var classType = new PythonClassType(BaseName, new Location(DeclaringModule));
                 // Storing generic parameters allows methods returning generic types 
                 // to know what type parameter returns what specific type
-                StoreGenericParameters(classType, genericTypeParameters, genericTypeToSpecificType);
+                classType.StoreGenericParameters(genericTypeParameters, genericTypeToSpecificType);
```
jakebailey: Doesn't changing this mean that `StoreGenericParameters` is now accessing the wrong class's `GenericParameters` property? Inside, it uses `GenericParameters` to copy over the original parameters, which before where the current class', but I think it'll be empty now that it's the new type's.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
MikhailArkhipov: Because `IPythonClassType : IPythonClassMember, IGenericType` and R# shows it as unnecessary

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
MikhailArkhipov: `IPythonClassType : IPythonClassMember, IGenericType`

src/Analysis/Ast/Impl/Extensions/EvaluatorExtensions.cs
MikhailArkhipov: Yep

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
jakebailey: Aha, yes.

MikhailArkhipov: `IPythonClassType` is derived from it

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
MikhailArkhipov: No, it effectively used `classType` as `this`

src/Analysis/Ast/Impl/Types/PythonClassType.cs
MikhailArkhipov: `SetBases` is internal so every call should guarantee that there won't be cycles. `ProcessBases` is just happens to be one of the cases. Otherwise every caller to `SetBases` will have to remember to resolve.

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
```diff
@@ -259,16 +260,15 @@ internal partial class PythonClassType {
         /// Points the generic type parameter in class type to their corresponding specific type (or a generic
         /// type parameter if no specific type was provided)
         /// </summary>
-        private void StoreGenericParameters(PythonClassType classType, IGenericTypeParameter[] genericParameters, IReadOnlyDictionary<IGenericTypeParameter, IPythonType> genericToSpecificTypes) {
+        private void StoreGenericParameters(IGenericTypeParameter[] genericParameters, IReadOnlyDictionary<IGenericTypeParameter, IPythonType> genericToSpecificTypes) {
             // copy original generic parameters over and try to fill them in
-            classType._genericParameters = new Dictionary<IGenericTypeParameter, IPythonType>(GenericParameters.ToDictionary(k => k.Key, k => k.Value));
+            _genericParameters = new Dictionary<IGenericTypeParameter, IPythonType>(GenericParameters.ToDictionary(k => k.Key, k => k.Value));
```
jakebailey: GenericParameters should be taken from the original type, rather than from here since the method receiver is essentially being changed. (Leaving here so we don't forget)

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
```diff
@@ -331,24 +331,24 @@ internal partial class PythonClassType {
         /// Transfers members from generic class to the specific class type
         /// while instantiating specific types for the members.
         /// </summary>
-        private void SetClassMembers(PythonClassType classType, IArgumentSet args) {
+        private void SetClassMembers(IArgumentSet args) {
             // Add members from the template class (this one).
             // Members must be clones rather than references since
             // we are going to set specific types on them.
-            classType.AddMembers(this, true);
+            AddMembers(this, true);
```
jakebailey: The meaning of `this` is changing here, since before `this` meant the parent type, while now it means the child. Similar to the other one above.

jakebailey: The comment here would need to be modified too.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -281,5 +284,55 @@ internal PythonClassType(string name, Location location)
...
+            }
+
+            if (eval == null) {
+                return FilterCircularBases(bases).Where(b => !b.IsUnknown());
+            }
+
+            var newBases = new List<IPythonType>();
+            foreach (var b in bases) {
+                var imported = eval.LookupImportedNameInScopes(b.Name, out _);
```
AlexanderSher: Does it matter in which scope we have found the import?

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
```diff
@@ -259,33 +256,40 @@ internal partial class PythonClassType {
...
+            IReadOnlyDictionary<IGenericTypeParameter, IPythonType> currentGenericParameters,
+            IEnumerable<IGenericTypeParameter> newGenericParameters,
+            IReadOnlyDictionary<IGenericTypeParameter, IPythonType> genericToSpecificTypes) {
+
             // copy original generic parameters over and try to fill them in
-            classType._genericParameters = new Dictionary<IGenericTypeParameter, IPythonType>(GenericParameters.ToDictionary(k => k.Key, k => k.Value));
+            _genericParameters = currentGenericParameters.ToDictionary(k => k.Key, k => k.Value);
 
             // Case when creating a new specific class type
             if (Parameters.Count == 0) {
```
jakebailey: This is also now a different `Parameters`. Maybe these two functions should just stay as they were since they access so much of the original class...

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
MikhailArkhipov: It actually needs to pass the template class

src/Analysis/Ast/Impl/Types/PythonClassType.cs
MikhailArkhipov: The search try to see if there is something imported before it is being declared. Outer scopes matter, i.e.
```python
from X import A

def func():
    class A(A): ...
    return A()
```
should find imported `A` up above and resolve base to it.
  
```

File: src/Analysis/Ast/Impl/Types/PythonClassType.cs (hunk 5 above)
Comment (MikhailArkhipov) [reply src/Analysis/Ast/Impl/Types/PythonClassType.cs]: Sure

File: src/Analysis/Ast/Impl/Types/PythonClassType.cs (hunk 4 above)
Comment (MikhailArkhipov) [reply src/Analysis/Ast/Impl/Types/PythonClassType.cs]: Nah, we don't want to process loops. Depends on guards. MRO guard might still apply...

File: src/Analysis/Ast/Impl/Types/PythonClassType.cs (hunk 12 above)
Comment (AlexanderSher) [reply src/Analysis/Ast/Impl/Types/PythonClassType.cs]: What about the opposite case:
```python
class A(A): ...

def func():
    from X import A
    return A()
```

src/Analysis/Ast/Impl/Types/PythonClassType.cs
CTrando: if eval is being passed into set bases then the work done in process bases can probably be put into set bases

src/Analysis/Ast/Impl/Types/PythonClassType.Generics.cs
MikhailArkhipov: It needs to receive template class from which it will only access public members.

MikhailArkhipov: `SetClassMembers(IPythonClassType templateClass, ...)` and `void StoreGenericParameters(
            IPythonClassType templateClass, ...`

src/Analysis/Ast/Impl/Types/PythonClassType.cs
MikhailArkhipov: I doesn't actually need eval, only scope.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
MikhailArkhipov:
```cs
class A(A): ...

def func():
    from Base import A
    return A()

x = func()
";
            analysis.Should().HaveVariable("x").OfType("A").Which.Should().HaveMember("methodABase");
```
passes

src/Analysis/Ast/Impl/Types/PythonClassType.cs
CTrando: Hmm perhaps the generic resolution guard is unnecessary now. Not sure about mro, I don't see how it could get loops if all the bases with loops are removed

src/Analysis/Ast/Impl/Types/PythonClassType.cs
CTrando: if eval is there then I think it would be better to move all processing inside SetBases. Otherwise we end up with the case where sometimes the bases are fully processed and sometimes they aren't

src/Analysis/Ast/Impl/Types/PythonClassType.cs
MikhailArkhipov: Maybe. But I'll leave it now for safety. We'll see f it can be removed later.

src/Analysis/Ast/Impl/Types/PythonClassType.cs
MikhailArkhipov: Replaced by scope

src/Analysis/Ast/Impl/Types/PythonClassType.cs
CTrando: OK

PR #1498, opened 2019-08-31
Unpacking typing Tuples and Lists correctly

### LINKED ISSUE #411: Single-element tuple assignments have no type information
When the LHS is a tuple with a single item, the assigned type is not propagated. Examples:

```python
(foo) = 1234
((x, y)) = 1, "2"
```

`foo`, `x`, and `y` all do not have known types.

Found in #334.

### LINKED ISSUE #1293: "Undefined variable" with multiple assignment and unpacking
Seems like the combination of multiple assignment and unpacking assignment confuses the parser or symbol tracker. Examples like this code:
```
a = b, _ = [0, 1]
print(b)
```
are giving me `Undefined variable: 'b'`

### LINKED ISSUE #1382: False positive undefined variable when destructuring function return values
The language server complains about `var1` being undefined.

fp

### LINKED ISSUE #1457: Tuple[int, int] not unpacked correctly into lvar tuple
# Description

In the following piece of code, the Microsoft Python Analysis Engine incorrectly infers the types of `a` and `b` when hovering over them.

# Minimal reproducible example

```python
from typing import Tuple, List

def foo() -> List[Tuple[int, int]]:
    return [(1, -1)]

result = foo()
(a, b) = result[0]
```

# Expected behavior

When hovering over `a` or `b`, I expect to see `a: int` and `b: int`.

# Actual behavior

When hovering over `a` or `b`, it shows `a: Tuple[int, int]` and `b: Tuple[int, int]`. I also tested this using PyCharm CE and it correctly inferred the types (i.e. `a: int` and `b: int`).

# System information

**OS**: macOS "Mojave" 10.14.6 (18G87)

## Visual Studio Code

**Python extension version**: 2019.8.30787

**Version**: 1.37.1
**Commit**: f06011ac164ae4dc8e753a3fe7f9549844d15e35
**Date**: 2019-08-15T16:16:34.800Z
**Electron**: 4.2.7
**Chrome**: 69.0.3497.128
**Node.js**: 10.11.0
**V8**: 6.9.427.31-electron.0
**OS**: Darwin x64 18.7.0

### LINKED ISSUE #1494: Intellisense not working when iterating over typed collection
This is with the daily download channel:

```python
from typing import List

def f(xs: List[str]):
    for x in xs:
        x.
```
The string methods are not being suggested after `x.`.

However, they *are* suggested after `xs[0].`

Screen Shot 2019-08-28 at 7 42 48 PM

CTrando:
Fixes #1457 
Fixes #1494 
Fixes #411
Fixes #1382 
Fixes #1293 

Also swapped all the PythonInstance constructor calls (when they are not needed) to the `IPythonType.CreateInstance`


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Callables.cs
```diff
@@ -53,7 +54,7 @@ internal sealed partial class ExpressionEval {
                     value = GetValueFromInstanceCall(pi, expr);
                     break;
                 case IPythonFunctionType ft: // Standalone function or a class method call.
-                    var instance = ft.DeclaringType != null ? new PythonInstance(ft.DeclaringType) : null;
+                    var instance = ft.DeclaringType != null ? ft.DeclaringType.CreateInstance(args) : null;
```
MikhailArkhipov: `ft.DeclaringType?.CreateInstance(args)`, no need for conditional anymore.

src/Analysis/Ast/Impl/Analyzer/Handlers/SequenceExpressionHandler.cs
```diff
@@ -65,20 +65,29 @@ internal sealed class SequenceExpressionHandler : StatementHandler {
                     case NameExpression nex when !string.IsNullOrEmpty(nex.Name):
                         eval.DeclareVariable(nex.Name, valueEnum.Next, VariableSource.Declaration, nex);
                         break;
+                    // Nested sequence expression in sequence, Tuple[Tuple[int, str], int], List[Tuple[int], str]
+                    // TODO: Because of bug with how collection types are constructed, they don't make nested collection types
+                    // into instances, meaning we have to create it here
+                    case SequenceExpression se when valueEnum.Peek is IPythonCollection || valueEnum.Peek is IPythonCollectionType:
```
MikhailArkhipov: You may be able to `case SequenceExpression se when valueEnum.Next is IPythonCollection pc || valueEnum.Next is IPythonCollectionType pct:`

src/Analysis/Ast/Test/TypingTests.cs
```diff
@@ -135,7 +135,7 @@ import io
 ";
             var analysis = await GetAnalysisAsync(code);
             analysis.Should().HaveVariable("Foo").OfType("Foo")
-                .And.HaveVariable("foo").OfType("Foo");
+                .And.HaveVariable("foo").OfType("dict");
```
MikhailArkhipov: Actually `NewType` is a distinct type so it should remain Foo (with base of dict). It acts like `class Foo(dict)`

CTrando: When I run the following script:
```
from typing import NewType

Foo = NewType('Foo', dict)
foo: Foo = Foo({ })

print(type(foo))
```
It prints out `<class 'dict'>` so I think it's safe to change this test.

MikhailArkhipov: From the point of runtime it is a `dict` like in C `typedef int BOOL` does not actually introduce boolean type. But from type checking perspective it is not a plain `dict` so calling `def func(x: Foo)` with `dict` is wrong as is calling `void func(BOOL x)` in C with `int`. Linters are augmented to check for this.

MikhailArkhipov: Another thing is what tooltip is going to show. As with `x: List[str]` where it says that `x` is `List[str]` and not `list` (which it actually is at runtime), in `x:Foo` the tooltip should display that `x` is `Foo` and not `dict`.

CTrando: Ok I've changed the test

src/Analysis/Ast/Impl/Analyzer/Handlers/SequenceExpressionHandler.cs
CTrando: I don't think I can - calling next will shift the index up one so the next time we call next it will have a different value

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Callables.cs
CTrando: ok

PR #1519, opened 2019-09-04
Ensure FileInfo is not relative in DirectoryInfoProxy

### LINKED ISSUE #1512: IndexManager.WorkspaceFiles returns paths rooted by cwd instead of workspace root
Causes errors like this, notice the file path:
```
System.IO.FileNotFoundException
  HResult=0x80070002
  Message=Could not find file 'C:\Projects\GitHub\Microsoft\python-language-server\output\bin\Debug\PythonApplication7.py'.
  Source=System.Private.CoreLib
  StackTrace:
   at System.IO.FileStream.ValidateFileHandle(SafeFileHandle fileHandle)
   at System.IO.FileStream.CreateFileOpenHandle(FileMode mode, FileShare share, FileOptions options)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share, Int32 bufferSize, FileOptions options)
   at System.IO.FileStream..ctor(String path, FileMode mode, FileAccess access, FileShare share)
   at System.IO.File.Open(String path, FileMode mode, FileAccess access, FileShare share)
   at Microsoft.Python.Core.IO.FileSystem.FileOpen(String path, FileMode mode, FileAccess access, FileShare share) in C:\Projects\GitHub\Microsoft\python-language-server\src\Core\Impl\IO\FileSystem.cs:line 35

```

See how my `_workspaceRootPath` is properly set, and also the `cwd` which I've added FYI.

image

And now the result, which is incorrectly rooted:

image

Problem seems to be in `DirectoryInfoProxy` class:
```
        public IEnumerable<IFileSystemInfo> EnumerateFileSystemInfos(string[] includePatterns, string[] excludePatterns) {
            var matcher = GetMatcher(includePatterns, excludePatterns);
            PatternMatchingResult matchResult = SafeExecuteMatcher(matcher);
            return matchResult.Files.Select((filePatternMatch) => {
                var path = PathUtils.NormalizePath(filePatternMatch.Path);
                return CreateFileSystemInfoProxy(new FileInfo(path));
            });
        }
```

The `FileInfo` is created with just the file name.

If I change that line to this, it works fine:
```
return CreateFileSystemInfoProxy(new FileInfo(Path.Combine(_directoryInfo.FullName, path)));
```

jakebailey:
Fixes #1512.

Since we're constructing `FileInfo` ourselves, and the matcher returns things relative to the directory the matching started at, ensure that we combine things to not not return a relative FileInfo.

Needs some testing to ensure that the proxy interface returns something consistent.

### DIFF
--- a/src/Core/Impl/IO/DirectoryInfoProxy.cs
+++ b/src/Core/Impl/IO/DirectoryInfoProxy.cs
@@ -55,7 +55,7 @@ namespace Microsoft.Python.Core.IO {
             var matcher = GetMatcher(includePatterns, excludePatterns);
             PatternMatchingResult matchResult = SafeExecuteMatcher(matcher);
             return matchResult.Files.Select((filePatternMatch) => {
-                var path = PathUtils.NormalizePath(filePatternMatch.Path);
+                var path = PathUtils.NormalizePath(Path.Combine(_directoryInfo.FullName, filePatternMatch.Path));
                 return CreateFileSystemInfoProxy(new FileInfo(path));
             });
         }
--- a/src/Core/Test/AssemblySetup.cs
+++ b/src/Core/Test/AssemblySetup.cs
@@ -1,0 +1,34 @@
+﻿// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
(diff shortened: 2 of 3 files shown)


CTrando:
If you revert your changes and run these tests, did it show that the path was appended to your current working directory?

jakebailey:
Yes, I did test that; the paths were instead rooted at the `Debug` directory.

PR #1522, opened 2019-09-05
Automatically check for a `src` directory in the workspace root

### LINKED ISSUE #1440: Automatically add common project roots (src) to search paths
Requires #1388. Based on work in #1384.

Once search path configuration is done outside of the initialization options we should:

- Add a new boolean to toggle a search path "heuristic".
- If enabled (and the old backwards-compat searchPaths/extraPaths/etc are not specified), then we should detect a common import root and automatically add it to the search path list.

To start, this can be as simple as looking in the workspace's top level for a `src` directory, then add to the top of the user search path lists.

In the future, this heuristic can either be expanded, or eliminated (if our import handling becomes loose enough to not require it).

jakebailey:
Fixes #1440.

Very basic implementation - if there's a top level `src` directory that doesn't contain a `__init__.py` file, then add it as an import root. This heuristic can be disabled by setting `python.analysis.autoSearchPaths` to false.


src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -103,6 +105,11 @@ public sealed partial class Server : IDisposable {
             _initParams = @params;
             _log = _services.GetService<ILogger>();
 
+            Root = _initParams?.rootUri != null ? _initParams.rootUri.ToAbsolutePath() : _initParams?.rootPath;
+            if (Root != null) {
+                Root = PathUtils.NormalizePathAndTrim(Root);
+            }
+
```
MikhailArkhipov: Up to you I guess it is `_initParams?.rootUri?.ToAbsolutePath() ?? _initParams?.rootPath`

src/LanguageServer/Impl/LanguageServer.Configuration.cs
```diff
@@ -162,5 +171,29 @@ public sealed partial class LanguageServer {
 
             return ImmutableArray<string>.Empty;
         }
+
+        private ImmutableArray<string> AutoSearchPathHeuristic() {
```
MikhailArkhipov: Do you want to add tests for it?

jakebailey: Sure, I can see about moving this to another place to test directly given a `IFilesystem`.

PR #1548, opened 2019-09-16
Add support for Python 3.8 fstring debug `=`

### LINKED ISSUE #1102: Support 3.8 fstring debug syntax in parser
Python 3.8 comes with a new fstring feature:

```python
print(f"{name=}")
print(f"{name =}")
print(f"{name = }")
print(f"{foo.bar()=}")
# etc
```

Is equivalent to writing:

```python
print(f"name={repr(name)}")
print(f"name ={repr(name)}")
print(f"name = {repr(name)}")
print(f"foo.bar()={repr(foo.bar())}")
# etc
```

We don't particularly need to know about the string that gets printed, but we'll have to handle the extra `=` at the end in 3.8+.

Note that these can be combined with `!` stuff, so:

```python
print(f"{foo=!s}")
```

https://bugs.python.org/issue36817

jakebailey:
Fixes #1102.

Plus a few extra cleanups (methods to properties, error reporting fixes, etc)

### DIFF
--- a/src/Parsing/Impl/FStringParser.cs
+++ b/src/Parsing/Impl/FStringParser.cs
@@ -50,8 +50,8 @@ namespace Microsoft.Python.Parsing {
         }
 
         public void Parse() {
-            var bufferStartLoc = CurrentLocation();
-            while (!EndOfFString()) {
+            var bufferStartLoc = CurrentLocation;
+            while (!EndOfFString) {
                 if (IsNext(DoubleOpen)) {
                     _buffer.Append(NextChar());
                     _buffer.Append(NextChar());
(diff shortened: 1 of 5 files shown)


src/Parsing/Impl/FStringParser.cs
```diff
@@ -132,31 +136,50 @@ private bool IsNext(StringSpan span)
...
             }
 
-            Debug.Assert(CurrentChar() == '}' || CurrentChar() == '!' || CurrentChar() == ':');
+            Debug.Assert(CurrentChar == '}' || CurrentChar == '!' || CurrentChar == ':' || (_langVersion >= PythonLanguageVersion.V38 && CurrentChar == '='));
 
+            MaybeReadEqualSpecifier();
             var conversion = MaybeReadConversionChar();
             var formatSpecifier = MaybeReadFormatSpecifier();
+
+            if (CurrentChar != '}') {
```
MikhailArkhipov: I think you can do `_incomplete =  !Read('}')`;

PR #1565, opened 2019-09-18
Implement declaration of submodules in module members

### LINKED ISSUE #1098: Intellisense doesnt work for all Pytorch methods
@Coderx7 commented on [Wed May 15 2019](https://github.com/microsoft/vscode-python/issues/5643)

## Environment data

- VS Code version: **1.33.1/1.34.0**
- Extension version (available under the Extensions sidebar): **2019.4.12954**
- OS and version: **Windows_NT x64 10.0.17763**
- Python version (& distribution if applicable, e.g. Anaconda): **3.6.7 Anaconda3**
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): **N/A**
- Relevant/affected Python packages and their versions: **Pytorch 1.0.0** and **1.1.0**

## Expected behaviour

Intellisense should work on Pytorch methods as well but it does not!

## Actual behaviour

It wont work on  methods such as `mm`, `matmul`, `mul`, `from_numpy`, `sum`, `randn`, `tensor`, `exp`, `empty` `arange`, `zeros`, `cat` ,etc as described [here ](https://github.com/pytorch/pytorch/issues/16996) in details.

## Steps to reproduce:
1. Simply try to use one of the aforementioned torch methods such as `sum`, `empty`,`randn`,`tensor`,etc

## Logs
Output for `Python` in the `Output` panel (`View`→`Output`, change the drop-down the upper-right of the `Output` panel to `Python`)

```
Starting Jedi Python language engine.
************* Module lesson4_1
9,19,error,no-member:Module 'torch' has no 'exp' member
15,11,error,no-member:Module 'torch' has no 'randn' member
16,10,error,no-member:Module 'torch' has no 'randn_like' member
17,7,error,no-member:Module 'torch' has no 'randn' member
20,4,error,no-member:Module 'torch' has no 'sum' member
23,0,error,no-member:Module 'torch' has no 'empty' member

----------------------------------------------------------------------

Your code has been rated at -11.43/10 (previous run: -11.43/10, +0.00)
```   

---

@DonJayamanne commented on [Thu May 16 2019](https://github.com/microsoft/vscode-python/issues/5643#issuecomment-493269706)

Please could you try switching to the `Language Server`.
Please check here for instructions https://github.com/microsoft/vscode-python/issues/3977

Let me know whether this resolves the problem for you.

---

@Coderx7 commented on [Thu May 16 2019](https://github.com/microsoft/vscode-python/issues/5643#issuecomment-493327549)

@DonJayamanne : I had a look at the link you provided. the problem is I cant find a way to install Language Server, as the instructions given there doesn't work for me. 
I dont have a `Language Server` folder, to remove in first place! 
This is how my extension directory looks like : 
image
and this is the contents of `ms-python.python` folder : 
image

I just updated the VSCode to the latest version couple of hours ago and this has not changed (i.e. no language server is installed)
I'd appreciate any feedback on how to get that installed so I can give you the result you need. 

---

@DonJayamanne commented on [Thu May 16 2019](https://github.com/microsoft/vscode-python/issues/5643#issuecomment-493335164)

All you need to do is:
>  Go into your settings.json, set "python.jediEnabled": false.

---

@Coderx7 commented on [Fri May 17 2019](https://github.com/microsoft/vscode-python/issues/5643#issuecomment-493361716)

Thanks, I did so and language server seem to have its own share of issues, while the mentioned methods now show up in intellisense, the presentation is different and more importantly, intellisense doesnt work flawlessly here either, here is a demonstration show casing the issues:  
vscode_pytorchintellisense2

(side note: jediEnabled presentation/look is much more desirable than what is being provided by language server! apart from its bugs! I myself prefer jediEnabled over language sever)

---

@DonJayamanne commented on [Fri May 17 2019](https://github.com/microsoft/vscode-python/issues/5643#issuecomment-493518731)

> here is a demonstration show casing the issues:

Please note, you have `jedi` enabled. Please change the value to `false`.
Once done, please provide an animated gif as done previously. Thanks a lot.

---

@Coderx7 commented on [Fri May 17 2019](https://github.com/microsoft/vscode-python/issues/5643#issuecomment-493520013)

@DonJayamanne In the gif I posted above, I tested both scenarios, both with jediEnabled set to true and false! 
Just watch the whole gif to the end and you'll see the difference between the  two. 
I start with JediEnabled and show how it works, and then disable it and show how language server works

---

@DonJayamanne commented on [Fri May 17 2019](https://github.com/microsoft/vscode-python/issues/5643#issuecomment-493551270)

When using Jedi, the auto completion list is provided by an external package and we do not have much control over that. Please feel free to report an issue on their repo https://github.com/davidhalter/jedi.

On the other hand the language server is something that's built by us.
Hence I'll be moving this issue to the language server repo so they can address the issues/concerns you have.

### LINKED ISSUE #1301: torch.nn missing from torch module and `from torch import` completions
Hi,
I'm the author of type stubs for PyTorch, and it seems VSCode is not suggesting some autocompletes of methods and modules that the stubs clearly state exist and that mypy and PyCharm make use of (using the 'daily' download channel). I haven't been able to construct a MWE yet.

To reproduce, install PyTorch master (as of 046c4589df15fc4e7b832d594d3d8e3a54cd47c9) following the instructions at https://github.com/pytorch/pytorch#get-the-pytorch-source. *NOT* the release version, which doesn't yet ship with the type stubs the language server seems to be having trouble using.

In a buffer, after waiting for the analysis to complete, try

```python
import torch
torch.
```

the list of autocompletes does not include the `nn` module, even though `from . import nn as nn` appears in `torch/__init__.pyi`.

Further,

```python
from torch import nn
nn.
```

doesn't provide any suggestions from the language server, even though `torch/nn/__init__.pyi` contains

```python
from .modules import * 
from .parameter import Parameter as Parameter
from .parallel import DataParallel as DataParallel
from . import init as init
from . import utils as utils
```

### LINKED ISSUE #1395: Autocomplete doesn't list pandas.testing
VScode version: 1.36.1
ms-python.python: 2019.6.24221
Microsoft Python Language Server version 0.3.43.0

In the terminal I have a virtual environment which was set by `conda` with Python version 3.7.3. The packages in the environment were installed from a `requirements.txt` file:

```
pytest
dvc[s3]==0.51.2
click==7.0
numpy==1.16.4
pandas==0.24.2
pyarrow==0.14.0
dask[complete]==2.1.0
graphviz==0.11.1
```

In the following, 

image

I was expecting to see the suggestion for `.testing` and not only for `test`. Is it a bug or a problem in my settings?

MikhailArkhipov:
Fixes #1098
Fixes #1301
Fixes #1395

Add submodules to module members so they show up in completions on the module name. Otherwise we show them in

image

but not in
```python
import torch
torch.
```


src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -157,7 +158,9 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
         #endregion
 
         #region IMemberContainer
-        public virtual IMember GetMember(string name) => GlobalScope.Variables[name]?.Value;
+        public virtual IMember GetMember(string name)
+            => GlobalScope.Variables[name]?.Value ?? this.GetSubmodule(name);
```
jakebailey: Is it possible for `Value` to be null on purpose such that we'd be better off doing a try-get from `Variables`, then resorting to the module name?

MikhailArkhipov: Unusual (normally it would be `Unknown`) but I guess it does not hurt to handle.

src/Analysis/Ast/Test/ImportTests.cs
```diff
@@ -234,5 +237,19 @@ import sys
...
+            var appUri = TestData.GetTestSpecificUri("app.py");
+            await TestData.CreateTestSpecificFileAsync(Path.Combine("package", "__init__.py"), "import m1");
+            await TestData.CreateTestSpecificFileAsync(Path.Combine("package", "m1", "__init__.py"), string.Empty);
+
+            await CreateServicesAsync(PythonVersions.LatestAvailable3X);
+            var rdt = Services.GetService<IRunningDocumentTable>();
+            var doc = rdt.OpenDocument(appUri, "import package");
+
+            var analysis = await doc.GetAnalysisAsync(Timeout.Infinite);
+            analysis.Should().HaveVariable("package").Which.Should().HaveMember("m1");
```
AlexanderSher: This is wrong. 
This example works (output: `module 'package.m1' from 'D:\\Temp\\package\\m1\\__init__.py'`):
```
from package import m1
print(m1)
```

This example doesn't (output: `AttributeError: module 'package' has no attribute 'm1'`):
```
import package
print(package.m1)
```

AlexanderSher: There are tests `InitPyVsModuleNameImport_AbsoluteImport` and `PackageModuleImport` that show correct behavior

jakebailey: Yeah, I think we're going to have to handle this another way, like ensuring that modules make it out as members of the module. Since that's going to get big (with every module that some other module imports as a member), maybe we can do something like "export it if it's a module below this one", which would probably handle many common cases of libraries re-exporting submodules.

AlexanderSher: Submodules and subpackages are available as members only if they either are exported from module or package's `__init__.py` or if they are imported by previous code (that is what `PythonVariableModule` is for):
```py
from package import m1
import package
print(package.m1)
```

AlexanderSher: @jakebailey , I don't think there is anything we can address. Original bug #1301 has issue with `import torch` probably because in `torch/__init__.py` it doesn't list `nn` in the `__all__` explicitly.

jakebailey: `__all__` is only applicable for `*` imports. It does not affect member accesses, only what is copied into the scope when doing something like `from torch import *`.

jakebailey: I think the actual issue is that we don't offer modules in other modules as members, but getting that right while not also being annoying is tricky.

MikhailArkhipov: True but if user types it anyway we don't squiggle member saying it does not exist. Specific case is in the bug, i.e. people want `pandas.testing` to show up so they can type `pandas.testing.assert`

jakebailey: But it may not exist, so showing it is going to be misleading since it'd crash at runtime.

AlexanderSher: Yep, I was wrong about `__all__`. The problem is that we ignore modules that are imported by `torch` itself or any other modules. If I change `package/__init__.py` to 
```py
import package.m1
```
then
```py
import package
print(package.m1)
```
will work despite the fact that there is no member `m1` in `package`.

AlexanderSher: This information is collected in `ImportHandler._variableModules` during analysis, but it isn't shared between modules. And passing it through is tricky indeed. For example, if we add a module `package/m2.py`:
```py
from package import m1 as p
```
and use the following code in `package/__init__.py`:
```py
import package.m2
```
then 
```py
import package
print(package.m1)
```
will work again.

src/Analysis/Ast/Impl/Extensions/PythonModuleExtensions.cs
```diff
@@ -70,5 +74,39 @@ internal static void AddAstNode(this IPythonModule module, object o, Node n)
 
         internal static bool IsNonUserFile(this IPythonModule module) => module.ModuleType.IsNonUserFile();
         internal static bool IsCompiled(this IPythonModule module) => module.ModuleType.IsCompiled();
+
+        public static IEnumerable<string> GetSubmoduleNames(this IPythonModule module) {
```
MikhailArkhipov: Technically no longer used... Keep or remove? May come useful at some point.

src/Analysis/Ast/Test/ImportTests.cs
```diff
@@ -234,5 +238,20 @@ import sys
...
+            await TestData.CreateTestSpecificFileAsync(Path.Combine("package", "__init__.py"), "from . import m1\nimport sys");
+            await TestData.CreateTestSpecificFileAsync(Path.Combine("package", "m1", "__init__.py"), string.Empty);
+
+            await CreateServicesAsync(PythonVersions.LatestAvailable3X);
+            var rdt = Services.GetService<IRunningDocumentTable>();
+            var doc = rdt.OpenDocument(appUri, "import package");
+
+            await Services.GetService<IPythonAnalyzer>().WaitForCompleteAnalysisAsync();
+            var analysis = await doc.GetAnalysisAsync(Timeout.Infinite);
+            analysis.Should().HaveVariable("package").Which.Should().HaveMember("m1").And.NotHaveMember("sys");
```
AlexanderSher: This assertion is incorrect in both cases:
- `package.sys` is a real member of the package
- `m1` doesn't have to be submodule. It is actually doesn't matter where module is located. What matters is if the name is present in `sys.modules`, which effectively depends on loading order. Here is an example:
   - File structure
    ```
    ├─ package
    │  ├─ sub1
    │  │  └─ m1
    │  │     └─ __init__.py
    │  ├─ sub2
    │  │  └─ m2.py
    │  ├─ m3.py
    │  └─ __init__.py
    └─ app.py
    ```
  - `package/sub1/m1/__init__.py`:

    ```py
    import sys
    ```
  - `package/sub2/m2.py`:

    ```py
    from package.sub1 import m1 as p
    ```
  - `package/__init__.py`:

    ```py
    import package.sub2.m2
    ```
  - `app.py`:

    ```py
    from sys import modules
    import package
    print('package.sub1.m1' in modules, hasattr(package.sub1, 'm1'), sep=', ') # True, True
    print('package.sub1.m1.sys' in modules, hasattr(package.sub1.m1, 'sys'), sep=', ') # False, True
    print('package.sub1.m3' in modules, hasattr(package.sub1, 'm3'), sep=', ') # False, False
    ```

MikhailArkhipov: Sys is indeed a member but we don’t want to show it since otherwise we will show everything module imports. We never showed it before. User can type it if they wish but that is not something we offer. Basically we relax the filter and allow imported sub modules as users want

MikhailArkhipov: So the point is to resolve user request to see pandas.testing and torch.mm

AlexanderSher: The problem is, as @jakebailey mentioned, that there can be no member, even if module itself exists (see the last line with `package.m3` in my example).

MikhailArkhipov: In the filter we already know it is present in the module as a variable. I.e. `GetMemberNames` goes through existing variables so it won't pick up something that is not imported. The difference is that before we did not show any of the imported modules and now we allow those variables that hold imported submodules.

src/Analysis/Ast/Impl/Modules/PythonModule.cs
MikhailArkhipov: Filtering changes since.

src/Analysis/Ast/Impl/Extensions/PythonModuleExtensions.cs
```diff
@@ -70,5 +74,39 @@ internal static void AddAstNode(this IPythonModule module, object o, Node n)
...
+            var searchResult = module.Interpreter.ModuleResolution.CurrentPathResolver.FindImports(module.FilePath, Enumerable.Repeat(module.Name, 1), 0, true);
```
AlexanderSher: You probably want to use `PathResolverSnapshot.GetModuleImportFromModuleName` here

src/Analysis/Ast/Test/ImportTests.cs
jakebailey: Yeah, this is the middle ground we came up with for the most common case (re-exporting something in the project out).

Another solution entirely might be to just let _everything_ be a member, then start filtering at completion time. My worry with not adding things as members is that it's valid code to write something like `from package.sub1.m1 import sys`, which might be important to preserve. If our issue with re-exporting every member is that it's annoying for users, maybe deferring the filter is the right way to go. (Sorry to suggest yet another solution, just brainstorming)

malmaud: Maybe worth noting the rules on when imports are considering exporting a type stub (from https://www.python.org/dev/peps/pep-0484/):

```
Additional notes on stub files:

Modules and variables imported into the stub are not considered exported from the stub unless the import uses the import ... as ... form or the equivalent from ... import ... as ... form.

However, as an exception to the previous bullet, all objects imported into a stub using from ... import * are considered exported. (This makes it easier to re-export all objects from a given module that may vary by Python version.)
```

jakebailey: Okay, that's definitely useful knowledge as I don't think we treat stubs any differently when handling `*` imports.

src/Analysis/Ast/Impl/Extensions/PythonModuleExtensions.cs
MikhailArkhipov: It was copied verbatim from `ImportCompletions`. Since no one actually uses it anymore, I'll remove it.

src/Analysis/Ast/Test/ImportTests.cs
MikhailArkhipov: Yes, I thought about filtering at completion, which is more appropriate, but it requires creating something more complex than GetMemberNames, like `GetMemberInfos` which would also provide additional data on what the member is exactly.

src/Analysis/Ast/Test/ImportTests.cs
jakebailey: After some experimentation, the example of `import package.m1` inside of `package`'s `__init__.py` causing `m1` to be available to importers of `package` via `import package` as `package.m1` is definitely true; we're seeing it for some parts of `pandas` as well. Seems like that needs to be fixed as well as making sure that they're saved as members for completion.

jakebailey: image

Where `top/__init__.py` is:

```python
from top import sub1

import top.sub2
import top.sub3.sub4
```

Running:

```python
import top
print(dir(top))
```

Gives:
```
[
    '__builtins__',
    '__cached__',
    '__doc__',
    '__file__',
    '__loader__',
    '__name__',
    '__package__',
    '__path__',
    '__spec__',
    'sub1',
    'sub2',
    'sub3',
    'top'
]
```

jakebailey: So essentially, if you make an import within your module of a submodule, then there's a member that's added to that module for the submodule, such that you can access it externally, but no more than that single level.

src/Analysis/Ast/Test/ImportTests.cs
MikhailArkhipov: Should be resolved in the latest update

src/Analysis/Ast/Test/ImportTests.cs
MikhailArkhipov: Done

MikhailArkhipov:
image

image

Python 3.7, pip-installed 1.2

MikhailArkhipov:
image

image

src/Analysis/Ast/Test/ImportTests.cs
```diff
@@ -234,5 +238,57 @@ import sys
             var analysis = await GetAnalysisAsync(code);
             analysis.Should().HaveVariable("x").OfType(BuiltinTypeId.Int);
         }
+
+        [TestMethod, Priority(0)]
+        public async Task ModuleInternalImportSys() {
+            var appUri = TestData.GetTestSpecificUri("app.py");
+            await TestData.CreateTestSpecificFileAsync(Path.Combine("package", "__init__.py"), "from . import m1\nimport sys");
```
AlexanderSher: Why do you need real files in file system?

MikhailArkhipov: Will Importing/dependency resolution need something to load?

src/Analysis/Ast/Test/ImportTests.cs
```diff
@@ -234,5 +238,57 @@ import sys
...
+            await TestData.CreateTestSpecificFileAsync(Path.Combine("top", "sub3", "__init__.py"), string.Empty);
+            await TestData.CreateTestSpecificFileAsync(Path.Combine("top", "sub3", "sub4.py"), string.Empty);
+
+            await CreateServicesAsync(PythonVersions.LatestAvailable3X);
+            var rdt = Services.GetService<IRunningDocumentTable>();
+            var doc = rdt.OpenDocument(appUri, "import top");
+
+            await Services.GetService<IPythonAnalyzer>().WaitForCompleteAnalysisAsync();
+            var analysis = await doc.GetAnalysisAsync(Timeout.Infinite);
+            analysis.Should().HaveVariable("top").Which.Should().HaveMembers("sub1", "sub2", "sub3", "top");
```
AlexanderSher: What should happen here in `__init__.py` itself, e.g.:
```cs
var top = rdt.GetDocument(TestData.GetTestSpecificUri("top", "__init__.py"));
var topAnalysis = await top.GetAnalysisAsync(-1);
analysis.Should().HaveVariable("sub2").And.NotHaveVariable("sub3");
```

AlexanderSher: Also, I would've check `sub3` with something like this:
```cs
await TestData.CreateTestSpecificFileAsync(Path.Combine("top", "sub3", "__init__.py"), @"
import top.sub3.sub4
");
```
and then
```cs
var sub3 = rdt.GetDocument(TestData.GetTestSpecificUri("top", "__init__.py"));
var sub3Analysis = await sub3.GetAnalysisAsync(-1);
analysis.Should().HaveVariable("sub4").And.NotHaveVariable("sub3");
```

MikhailArkhipov: `analysis.Should().HaveVariable("sub2").And.NotHaveVariable("sub3");` - `top` actually has `sub3` which is produced by `import top.sub3.sub4`, as tested in the console.

MikhailArkhipov: And `import top.sub3.sub4` does not seem to declared `sub4`

MikhailArkhipov: I.e. in the 
```
Python 3.7.2 (tags/v3.7.2:9a3ffc0492, Dec 23 2018, 23:09:28) [MSC v.1916 64 bit (AMD64)] on win32
>>> import top
>>> dir(top)
['__builtins__', '__cached__', '__doc__', '__file__', '__loader__', '__name__', '__package__', '__path__', '__spec__', 'sub1', 'sub2', 'sub3', 'top']
>>>
```
So basically
```python
from top import sub1
import top.sub2
import top.sub3.sub4
```
yields members `top`, `sub1`, `sub2` and `sub3` but not `sub4`.

src/LanguageServer/Impl/Completion/ExpressionCompletion.cs
```diff
@@ -41,23 +44,57 @@ internal static class ExpressionCompletion {
...
+            var variables = module.Analysis?.GlobalScope?.Variables;
+            if (variables == null) {
+                return module.GetMemberNames();
+            }
+
+            var fs = context.Services.GetService<IFileSystem>();
+            var thisModuleDirectory = Path.GetDirectoryName(module.FilePath);
+            // drop imported modules and typing.
+            return variables
+                .Where(v => {
```
jakebailey: If I do `torch.`, I can see the builtin types in the completion. Should those be filtered? (It's also a little inconsistent, `torch.nn.` doesn't give `int`.)

src/LanguageServer/Impl/Completion/ImportCompletion.cs
```diff
@@ -60,7 +59,7 @@ internal class ImportCompletion {
 
         public static CompletionResult GetCompletionsInFromImport(FromImportStatement fromImport, CompletionContext context) {
             // No more completions after '*', ever!
-            if (fromImport.Names != null && fromImport.Names.Any(n => n?.Name == "*" && context.Position > n.EndIndex)) {
+            if (fromImport.Names.Any(n => n?.Name == "*" && context.Position > n.EndIndex)) {
```
jakebailey: Before, `Names` was checked for null, but now it isn't. Does this need something like `?.Any(...) ?? false)` or `MaybeEnumerate().Any()`?

src/LanguageServer/Impl/Completion/ExpressionCompletion.cs
MikhailArkhipov: They appear to be of a different type, so it is not filtering issue. Pandas don't show them either.

image

jakebailey: Ah, I see now, they're `dtype`. Wonder why you can't F12 to them...

MikhailArkhipov: https://pytorch.org/docs/stable/type_info.html

src/LanguageServer/Impl/Completion/ImportCompletion.cs
jakebailey: I now see that you only removed the null check, so maybe it's known not-null.

MikhailArkhipov: It is `ImmutableArray`, VS shows it as unnecessary check although I don't mind.

jakebailey: Ah, yeah. It's probably upgrading the ImmutableArray to ImmutableArray? and checking null, which is pointless.

MikhailArkhipov: It's a struct `ImmutableArray` so VS shows it as unnecessary check although I don't mind either way.

src/LanguageServer/Impl/Completion/ExpressionCompletion.cs
MikhailArkhipov: Most probably compiled. In-memory type came from stub which we don't show. `lib\site-packages\torch\__init__.pyi`

jakebailey:
Here's an interesting example where the top level module is able to essentially "add" members to child modules. Maybe we also need to be doing this via VariableModule.

```python
# a/__init__.py
import a.b.c.d
class A:
    pass

# a/b/__init__.py
class B:
    pass

# a/b/c/__init__.py
class C:
    pass

# a/b/c/d/__init__.py
class D:
    pass
```

Running:

```python
import a

def print_members(prefix, x):
    print(prefix, [m for m in dir(x) if not m.startswith("__")])

print_members("members of a:      ", a)
print_members("members of a.b:    ", a.b)
print_members("members of a.b.c:  ", a.b.c)
print_members("members of a.b.c.d:", a.b.c.d)
```

Gives:

```
members of a:       ['A', 'a', 'b']
members of a.b:     ['B', 'c']
members of a.b.c:   ['C', 'd']
members of a.b.c.d: ['D']`
```

jakebailey:
And unfortunately, this effect persists to imports of another form:

```python
import a

def print_members(prefix, x):
    print(prefix, [m for m in dir(x) if not m.startswith("__")])

print_members("members of a:      ", a)
print_members("members of a.b:    ", a.b)
print_members("members of a.b.c:  ", a.b.c)
print_members("members of a.b.c.d:", a.b.c.d)

from a.b import c

print_members("members of c:      ", c)
```

```
members of a:       ['A', 'a', 'b']
members of a.b:     ['B', 'c']
members of a.b.c:   ['C', 'd']
members of a.b.c.d: ['D']
members of c:       ['C', 'd']
```

jakebailey:
Using my example above, seeing this weird completion:

image

Also, `from a.b import c` resolves the variable `c` in `a.b`, but `c` should be the module `a.b.c` instead.

src/Analysis/Ast/Impl/Analyzer/Handlers/ImportHandler.cs
```diff
@@ -56,7 +56,7 @@ internal sealed partial class ImportHandler : StatementHandler {
             // import_module('fob.oar.baz')
             var importNames = ImmutableArray<string>.Empty;
             var lastModule = default(PythonVariableModule);
-            var modules = new PythonVariableModule[moduleImportExpression.Names.Count];
+            var modules = new (string, PythonVariableModule)[moduleImportExpression.Names.Count];
```
jakebailey: You should actually be able to give the tuple items names, like `string name, PythonVariableModule module` and avoid the `Item1`/`Item2` thing below.

MikhailArkhipov: Done

PR #1572, opened 2019-09-20
Allow packages in GetAllImportableModuleNames

### LINKED ISSUE #1392: Unresolved Import on 3rd Party Installed Library, But Still Receiving Partial Autocomplete
### Version Info

VSCode Version: 1.36.1
OS: Windows 10
Language Server Version: 0.3.40
Python: Anaconda 2.7

---------------------
When I import the library `pyteomics` or any of its submodules, I get an "unresolved-import" warning

image

 I know it is installed with `pip`
```
(base) C:\Users\jklein>pip list | grep pyteomics
pyteomics                          4.1
```
and the code successfully imports `pyteomics` and calls its functions.

What's more bizarre is that I can still get import suggestions while typing the import statement
image

Once the import statement is past the module name to import from, I do not get further autocomplete support.

If I import a name from a submodule, I *do* get intellisense:

image

This problem appears to be absent on Python 3.6.7, but the "Analyzing in background" queue never empties, so I do not know if that could be blocking the warning.

MikhailArkhipov:
Fixes #1392

### DIFF
--- a/src/Analysis/Ast/Test/ImportTests.cs
+++ b/src/Analysis/Ast/Test/ImportTests.cs
@@ -13,10 +13,14 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
+using System.IO;
 using System.Linq;
+using System.Threading;
 using System.Threading.Tasks;
 using FluentAssertions;
+using Microsoft.Python.Analysis.Analyzer;
 using Microsoft.Python.Analysis.Diagnostics;
+using Microsoft.Python.Analysis.Documents;
 using Microsoft.Python.Analysis.Modules;
 using Microsoft.Python.Analysis.Tests.FluentAssertions;
 using Microsoft.Python.Analysis.Types;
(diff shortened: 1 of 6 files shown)


src/Analysis/Ast/Test/ImportTests.cs
```diff
@@ -234,5 +238,22 @@ import sys
             var analysis = await GetAnalysisAsync(code);
             analysis.Should().HaveVariable("x").OfType(BuiltinTypeId.Int);
         }
+
+
+        [TestMethod, Priority(0)]
+        public async Task ImportPackageNoInitPy() {
+            var appUri = TestData.GetTestSpecificUri("app.py");
+            await TestData.CreateTestSpecificFileAsync(Path.Combine("top", "sub1", "__init__.py"), string.Empty);
```
jakebailey: Is this one of those things that only works in one of Python 2/3? Not sure what happens if you do this in the interpreter off hand.

MikhailArkhipov: Yes, in 3. Changed per @AlexanderSher suggestions since `GetAllModuleNames` actually did not return packages.

AlexanderSher: We need a parameter, something like `bool includeImplicitPackages`

MikhailArkhipov: For `GetAllImportable`?

AlexanderSher: Yes

MikhailArkhipov: Currently there is only one user, what scenario would need to exclude them? I don't mind either way.

jakebailey: For Python 2, right?

MikhailArkhipov: OK.

src/Analysis/Core/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -81,28 +81,26 @@ public PathResolverSnapshot(PythonLanguageVersion pythonLanguageVersion, string
             Version = version;
         }
 
-        public ImmutableArray<string> GetAllModuleNames() => GetModuleNames(_roots.Prepend(_nonRooted));
-        public IEnumerable<string> GetInterpreterModuleNames() => GetModuleNames(_roots.Skip(_userRootsCount).Append(_builtins));
-
-        private ImmutableArray<string> GetModuleNames(IEnumerable<Node> roots) {
+        public ImmutableArray<string> GetAllImportableModuleNames(bool includeImplicitPackages = true) {
```
AlexanderSher: Do you use this parameter in production?

jakebailey: Doesn't seem that way, is this what should be false if the interpreter is 2.x?

MikhailArkhipov: Nope

MikhailArkhipov: I can just use `LanguageVersionRequiresInitPy` instead.

jakebailey: Ah yes, that's better.

PR #1575, opened 2019-09-23
Fix Python 3.8 version checks in tests

CTrando:
Encountered this while I was working on the scope stuff. I think for version specific stuff like the := operator, we should specify the version of the test.

### DIFF
--- a/src/Analysis/Ast/Test/AssignmentTests.cs
+++ b/src/Analysis/Ast/Test/AssignmentTests.cs
@@ -686,8 +686,7 @@ from typing import List
 a: List[int]
 b = [(x := i) for i in a]
 ";
-            var analysis = await GetAnalysisAsync(code);
-
+            var analysis = await GetAnalysisAsync(code, PythonVersions.Required_Python38X);
             analysis.Should().HaveVariable("x").OfType(BuiltinTypeId.Int);
         }
     }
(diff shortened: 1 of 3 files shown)


src/Analysis/Ast/Test/AssignmentTests.cs
```diff
@@ -674,8 +674,7 @@ class D: ...
 if (x := 1234) == 1234:
     pass
 ";
-            var analysis = await GetAnalysisAsync(code);
-
+            var analysis = await GetAnalysisAsync(code, PythonVersions.Required_Python38X);
```
jakebailey: You'll want to do this to the test below this one as well.

CTrando: Done!

jakebailey:
Sorry to ask for more, but if you can, there's another one to add in HoverTests.AssignmentExpressions, then there are two NamedExpr tests in LintUndefinedVarsTests which should use `Required_Python38X` and not `Python38_x64`.

If you can't do it now, I can merge this and do those myself.

CTrando:
No problem very simple to do. Also in `AssignmentTests`, even with my scope changes it doesn't seem to pass because declaration of the variable in named expression doesn't occur outside the current scope. I'll let you handle that.

PR #1577, opened 2019-09-23
addressed csharp compiler breaking changes around default with binary…

heejaechang:
… operators

more detail
https://github.com/dotnet/roslyn/blob/master/docs/compilers/CSharp/Compiler%20Breaking%20Changes%20-%20post%20VS2019.md

breaking change number 3

and here is excerpt of the change

...

https://github.com/dotnet/roslyn/issues/35684 In C# 7.1 the resolution of a binary operator with a default literal could result in using an object equality and giving the literal a type object. For example, given a variable t of an unconstrained type T, t == default would be improperly allowed and emitted as t == default(object). Similarly, for a reference type without a custom == operator, x == default would be improperly allowed and emitted as x == default(object). In Visual Studio 2019 version 16.4 these scenarios will now produce an error.

....

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Handlers/ImportHandler.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Handlers/ImportHandler.cs
@@ -65,16 +65,16 @@ namespace Microsoft.Python.Analysis.Analyzer.Handlers {
                     break;
                 }
 
-                if (firstModule == default) {
+                if (firstModule == null) {
                     firstModule = lastModule;
                 }
             }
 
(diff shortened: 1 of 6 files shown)


MikhailArkhipov:
You can just do `null`, they are all classes.

jakebailey:
Even more frustrating is that `dotnet build` has no problem building everything. We thought the bug was the other way around.

heejaechang:
@MikhailArkhipov you prefer null over default? I can change to null.

@jakebailey this is post 2019 breaking changes, so you probably need most recent VS drop to see the errors.

jakebailey:
`null` would be most consistent, yes.

Issue #1511, opened 2019-09-04
Support positional-only parameters


jakebailey:
Requires at least a parser update, and potentially analysis changes (mainly linting).

https://docs.python.org/3.8/whatsnew/3.8.html#positional-only-parameters

PR #1578, opened 2019-09-23
Allowing any expression to declare a Scope

### LINKED ISSUE #1567: Treat comprehensions as scopes
Right now, we can only accept things that are `ScopeStatement` as scopes. However, comprehensions are also "scopes", in that they declare their own variables which should not leak out.

Additionally, with the introduction of `:=`, some variables _can_ escape out of a comprehension into the outer scope, which we currently do not handle either (see #1566).

See also: #786, #1041

CTrando:
Fixes #1567 and its associated issues. 

I created a `ScopeInfo` class that will store the free, global, closure, etc. variables and made it so that each `IScopeNode` stores a `ScopeInfo` object.

I did this because comprehensions should be treated just like functions in terms of scoping, and rather than create an interface and force the comprehensions to implement it (and any other future expressions), they can just use the `FunctionScopeInfo` object. 

Have some various testing as well for completions and scoping as well. Will add more tests when I get some more time to work on this.

There may be some formatting issues because I am using a different IDE, just point them out and I will fix them.


src/Analysis/Ast/Impl/Analyzer/Definitions/IExpressionEvaluator.cs
```diff
@@ -90,4 +91,4 @@ public interface IExpressionEvaluator {
         /// </summary>
         IPythonType UnknownType { get; }
     }
-}
+}
```
jakebailey: Here's an editor formatting difference; make sure you're setup to have files end with newlines.

src/Parsing/Impl/Ast/ScopeInfo.cs
```diff
@@ -1,41 +1,32 @@
...
-// permissions and limitations under the License.
-
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using Microsoft.Python.Core;
+using Microsoft.Python.Parsing.Definition;
 
 namespace Microsoft.Python.Parsing.Ast {
-    public abstract class ScopeStatement : Statement {
```
jakebailey: The parser's API shouldn't really be changing. Removing this entire type is going to cause issues for those who depend on our API (which is why its in its own assembly).

jakebailey: Deprecating behavior or adding more bases/interface implementations should be fine, but the overall API itself needs some care to change.

CTrando: Ok, so would it be enough to add `ScopeStatement` back in, but delegate its existing public interfaces to `ScopeInfo`? The other parts are internal, so I assume that it is fine to move them out of the class. 

I can have `ScopeStatement` implement `IScopeStatement` and then `ClassDefinition` and `FunctionDefinition` extend `ScopeStatement`.

src/Analysis/Ast/Impl/Analyzer/Definitions/IExpressionEvaluator.cs
CTrando: I'll get to it once I am closer to finishing

src/Parsing/Impl/Ast/ScopeInfo.cs
CTrando: I don't think FindScopes is going to be a problem, `ScopeStatement` implements `IScope` so people can still pass in their same objects. I also added the `ScopeStatement` class back and marked it as deprecated. The public fields should remain the same.

jakebailey: Sorry, I don't mean to say to actually "deprecate" the ScopeStatement class. It's just an AST node.

CTrando: Ok I'll change it back then. I think that people in the future should not use ScopeStatement in their code and it should ultimately be phased out because it doesn't do much.

CTrando:
Besides styling and refactoring, do you think there are any problems with my current strategy?

jakebailey:
Yes, building an interface above ScopeStatement and the comprehensions is what I had in mind. I would re-merge master just to make the diff clean. I'll leave more comments when I can, but my main concern would be the introduction of both `IScopeStatement` and `IScopeNode`; I had thought adding just one thing would be enough.

src/Parsing/Impl/Ast/FunctionScopeInfo.cs
```diff
@@ -0,0 +1,98 @@
...
+
+        internal override void Bind(PythonNameBinder binder) {
+            base.Bind(binder);
+            Verify(binder);
+        }
+
+        private void Verify(PythonNameBinder binder) {
+            if (ContainsImportStar && IsClosure) {
+                binder.ReportSyntaxError(
+                                         "import * is not allowed in function '{0}' because it is a nested function"
```
jakebailey: These would be good to move to Resources. Not sure why theyre hardcoded strings in the current code...

Not required, just looking.

CTrando:
Sure I added that because in some places we had `ScopeStatement.Body` which is a Statement object. So I could either make every ScopeNode require a Statement object as a body, but that doesn't really align with how Python works (or our code, as we have to convert an expression to a statement). For example, what is the statement body for a comprehension? So I made it so that some ScopeNodes can have bodies, which leads to the creation of ScopeStatement.

src/Parsing/Impl/Ast/FunctionScopeInfo.cs
CTrando: Ok I'll do that

jakebailey: You can hold off on it for now, I may need to modify the name binder in #1586 which won't merge well.

jakebailey: Though, if your PR can be done before mine, and the binding process works properly... Maybe I won't have to completely re-do scoping as well.

CTrando: I'm pretty sure the binding tests pass, I'm really just cleaning stuff up at this point unless you have any other issues.

CTrando: Actually I remember seeing some tests but I can't find them anymore, so how would I verify that the binding works?

jakebailey: I have to re-do what I have anyway, so you can ignore my comments. There are some binding tests in `ParserTests.Errors`, but otherwise you'd need to search for the exact strings.

CTrando: Because comprehensions have ScopeInfo in my changes, I think the scoping errors you are trying to address can be done in name binder. I'll try to get it done in the next few days, just tell me if there's anything else besides mild styling

CTrando:
Ready for a look.

src/Analysis/Ast/Impl/Analyzer/AnalysisWalker.cs
```diff
@@ -67,7 +69,7 @@ internal abstract class AnalysisWalker : PythonWalker {
                     AssignmentHandler.HandleAnnotatedExpression(ea, null);
                     return false;
                 case Comprehension comp:
-                    Eval.ProcessComprehension(comp);
+                    Eval.ProcessComprehensionInScope(comp, Eval.Interpreter.LanguageVersion.Is3x());
```
jakebailey: Eval contains Interpreter. Why pass explicitly when it can look up the version itself?

jakebailey: Could probably get away without changing the function name that way (smaller diff).

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
```diff
@@ -101,40 +102,47 @@ internal sealed partial class ExpressionEval {
...
-                        var v1 = GetValueFromExpression(lc.Item) ?? UnknownType;
-                        return PythonCollectionType.CreateList(Module, new[] { v1 });
-                    case SetComprehension sc:
-                        var v2 = GetValueFromExpression(sc.Item) ?? UnknownType;
-                        return PythonCollectionType.CreateSet(Module, new[] { v2 });
-                    case DictionaryComprehension dc:
-                        var k = GetValueFromExpression(dc.Key) ?? UnknownType;
-                        var v = GetValueFromExpression(dc.Value) ?? UnknownType;
-                        return new PythonDictionary(new PythonDictionaryType(Interpreter.ModuleResolution.BuiltinsModule), new Dictionary<IMember, IMember> { { k, v } });
+        public IMember GetValueFromComprehensionInScope(Comprehension node, bool is3X = true) {
```
jakebailey: Same as above. Could be the same name, and read is3X internaly, unless I've missed a usage.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
```diff
@@ -101,40 +102,47 @@ internal sealed partial class ExpressionEval {
...
+                case DictionaryComprehension dc:
+                    var k = GetValueFromExpression(dc.Key) ?? UnknownType;
+                    var v = GetValueFromExpression(dc.Value) ?? UnknownType;
+                    return new PythonDictionary(new PythonDictionaryType(Interpreter.ModuleResolution.BuiltinsModule), new Dictionary<IMember, IMember> { { k, v } });
+            }
+            return UnknownType;
+        }
+
+        internal void ProcessComprehensionInScope(Comprehension node, bool is3X = true) {
+            // In 2X there is a bug where comprehension variables get leaked to outer scopes, fixed in 3X
```
jakebailey: Nit: I don't think it was a bug, but just how comprehension scoping worked in 2x.

src/Analysis/Ast/Impl/Analyzer/AnalysisWalker.cs
CTrando: I did that because basically for 3X, we have to open a scope then process the comprehension, and for 2X we just have to process the comprehension. I didn't want to have the duplicated code so I wrapped it in another method `ProcessComprehension` and had `ProcessComprehensionInScope` handle whether to open a new scope or not. Do you know any constructs that would be nicer?

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
```diff
@@ -101,40 +102,47 @@ internal sealed partial class ExpressionEval {
...
+            if (is3X) {
+                using (OpenScope(Module, node)) {
```
jakebailey: I can see why you did it this way, in order to have a conditional `using`. I'm not sure if we can do it in this project (due to the .NET standard 2.0 requirement), but maybe one of those newfangled C# 8 usings would work in a conditional, where you can say "this disposable gets disposed at return".

Otherwise, it might be okay to do something like this:

```csharp
var scope = Disposable.Empty;
if (is3x) {
    scope = OpenScope(module, node)
}

using (scope) {
    // something
}
```

Or rename the private `ProcessComprehension` function and make this one `ProcessComprehension`.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
CTrando: OK

src/Analysis/Ast/Impl/Analyzer/AnalysisWalker.cs
CTrando: Keeping the same name and reading internally is fine though

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
CTrando: Ok I've run into situations like this in the past, and at my old company, we renamed the function to `Function0` for example in this case we would have `ProcessComprehension` and `ProcessComprehension0`. Is that ok with our coding guidelines? Do you have a suggestion for a name?

jakebailey: Not a fan of the `0` thing (not that I haven't seen it done). Maybe `ProcessComprehensionInScope` is actually a useful name, but reversed such that it's saying that "you're calling me in a scope".

CTrando: I might end up doing that nice trick with the disposable actually

jakebailey: FYI you can probably shorten this function a little by writing a using statement like:

```csharp
using (is3x ? OpenScope(module, node) : Disposable.Empty) {
    // work
}
```

And prevent a future error of forgetting to change two calls.

jakebailey: I think C# lets you pass a `null` disposable to a using and it'll skip over the dispose, but that feels very weird to me.

src/Parsing/Impl/Ast/PythonAst.cs
```diff
@@ -56,12 +58,12 @@ public sealed class PythonAst : ScopeStatement, ILocationConverter {
                 offset += a.NewLineLocations.Length + 1;
             }
             CommentLocations = comments.ToArray();
+            ScopeInfo = new AstScopeInfo(this);
         }
 
         public Uri Module { get; }
         public NewLineLocation[] NewLineLocations { get; private set; }
         public SourceLocation[] CommentLocations { get; private set; }
-        public override string Name => "<module>";
```
jakebailey: I've seen a lot of removal of `Name` in favor of `ScopeName`. Are you certain this would be a backwards compatible change?

src/Parsing/Impl/Definition/IScopeNode.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        /// <summary>
+        /// Gloabl scope
+        /// </summary>
+        PythonAst GlobalParent { get; }
+
+        /// <summary>
+        /// Holds information about the scope gathered from Parser, mainly used in  <see cref="PythonNameBinder"/>
+        /// </summary>
+        ScopeInfo ScopeInfo { get; }
```
jakebailey: This is a concrete type in an interface definition, which I'm not entirely sure about, though it is abstract. @MikhailArkhipov any thoughts on this?

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
```diff
@@ -167,7 +167,7 @@ internal sealed class ClassEvaluator : MemberEvaluator {
             // Do not use foreach since walker list is dynamically modified and walkers are removed
             // after processing. Handle __init__ and __new__ first so class variables are initialized.
             var constructors = SymbolTable.Evaluators
-                .Where(kvp => kvp.Key.Parent == cd && (kvp.Key.Name == "__init__" || kvp.Key.Name == "__new__"))
+                .Where(kvp => kvp.Key.Parent == cd && (kvp.Key.ScopeName == "__init__" || kvp.Key.ScopeName == "__new__"))
```
jakebailey: Similar question about `Name` vs `ScopeName`.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
```diff
@@ -25,6 +25,7 @@
 using Microsoft.Python.Core;
 using Microsoft.Python.Parsing;
 using Microsoft.Python.Parsing.Ast;
+using Microsoft.Python.Parsing.Definition;
```
MikhailArkhipov: We don't use 'Definitions' if interface is declared in `Definitions` subfolder. Just same namespace as the rest. Grouping in `definitions` folder is just a convenience thing.

src/Analysis/Ast/Impl/Analyzer/Expressions/ExpressionFinder.cs
```diff
@@ -189,6 +189,48 @@ private class NormalExpressionWalker : ExpressionWalker {
...
+                }
+
+                // Don't set scope if 2x 
+                if (!Tree.LanguageVersion.Is3x()) {
+                    return true;
+                }
+
+                Scope = node;
+                return true;
+            }
```
MikhailArkhipov: All three above is the same code, perhaps make function with `Comprehension` base parameter?

src/Parsing/Impl/Ast/ScopeInfo.cs
```diff
@@ -0,0 +1,297 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using Microsoft.Python.Core;
+using Microsoft.Python.Parsing.Definition;
+
+namespace Microsoft.Python.Parsing.Ast {
+    public abstract class ScopeInfo {
```
jakebailey: Did you modify the binding logic at all to handle comprehensions, or is this just the same logic but relocated? As noted in #1586, I'm not sure the binder was able to handle comprehension scoping properly, but if you're fixing it, the bulk of #1586 might be able to be removed in a future PR.

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
```diff
@@ -22,15 +22,17 @@
 using Microsoft.Python.Analysis.Values;
 using Microsoft.Python.Core;
 using Microsoft.Python.Core.OS;
+using Microsoft.Python.Parsing;
 using Microsoft.Python.Parsing.Ast;
+using Microsoft.Python.Parsing.Definition;
```
MikhailArkhipov: namespace

jakebailey:
Overall note that we probably shouldn't merge this until I can fix #1511, to make it easier for me to backport 3.8 into the PTVS branch. I don't want to also have to port the new scoping changes (though if all goes well, the AST should be fully compatible). We'll see.

src/Parsing/Impl/Definition/IScopeStatement.cs
```diff
@@ -0,0 +1,7 @@
+using Microsoft.Python.Parsing.Ast;
+
+namespace Microsoft.Python.Parsing.Definition {
```
MikhailArkhipov: namespace

MikhailArkhipov: If it is a folder, should be `Definitions` in the folder name, but no `Definitions` namespace.

src/Parsing/Impl/Ast/PythonAst.cs
MikhailArkhipov: Same here - is there a reason? Parser is public and used outside of the language server - IntelliCode and Python Tools for VS. We can't just change members.

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
MikhailArkhipov: +1

src/Parsing/Impl/Ast/PythonAst.cs
jakebailey: Given this is an override being removed, I'm not sure it's "gone" per se, but I haven't yet found the code that would redirect `ScopeName` to `Name`. My impression is that `INode` should have `string Name` and `ScopeName` can be dropped entirely.

src/Parsing/Impl/Definition/IScopeNode.cs
MikhailArkhipov: I am not sure about introduction of interfaces in the parser. It is all classes so it either needs to be done all interfaces or keep it with classes, otherwise it looks odd mix and match. Is there something that is `IScopeStatement` but not derived from `ScopeStatement`?

MikhailArkhipov: Ditto for `IScopeInfo` vs `ScopeInfo`. Looks like we are exposing `ScopeInfo` anyway so why the interface then.

jakebailey: I asked about this earlier, this was the response: https://github.com/microsoft/python-language-server/pull/1578#issuecomment-535150477

jakebailey: I'm personally not against interfacing some things. The problem is that comprehensions are expressions and don't share the same class tree as the existing scope statements, so something is going to have to be done...

src/Parsing/Impl/Ast/PythonAst.cs
CTrando: Ok I like the idea of adding Name to INode

src/Analysis/Ast/Impl/Analyzer/Symbols/ClassEvaluator.cs
CTrando: OK

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Generics.cs
CTrando: OK

src/Analysis/Ast/Impl/Analyzer/Expressions/ExpressionFinder.cs
CTrando: Yeah I was wondering - I would have to edit the interface in PythonWalker to do that. Otherwise I guess I could make a common function and have the overrides redirect to it.

src/Parsing/Impl/Ast/ScopeInfo.cs
CTrando: Nope all I handled was creating a new scope for comprehensions + framework for any Node. I could do that in a future PR

src/Analysis/Ast/Impl/Analyzer/Symbols/SymbolCollector.cs
CTrando: OK

src/Parsing/Impl/Definition/IScopeNode.cs
CTrando: I think we will have to interface this. The question is should we interface everything to be consistent or leave it partially done. I don’t particularly see a problem with having mix and match but if it’s a public API that might be a problem

src/Analysis/Ast/Impl/Analyzer/Expressions/ExpressionFinder.cs
jakebailey: Do you? All of the comprehension types have a common base, `Comprehension`. Generators, no (but I don't see that you've touched generators either).

src/Parsing/Impl/Definition/IScopeNode.cs
MikhailArkhipov: Yeah, `IScopeNode` may be inevitable, but "ScopeNode require a Statement object as a body" maybe less intrusive, may be just en empty body?

CTrando: Ok I can have null body meaning empty

src/Parsing/Impl/Ast/PythonAst.cs
jakebailey: If ScopeName gets eliminated, and everything continues to use Name, will there be any reason to have INode at all?

jakebailey: Well, given the number of things that IScopeNode picks up, maybe it would...

src/Parsing/Impl/Definition/IScopeNode.cs
MikhailArkhipov: Yeah, `IScopeNode` may be inevitable I thinking other cases like INode etc

src/Parsing/Impl/Ast/PythonAst.cs
CTrando: I have INode in order to have IScopeNode implement it and handle interfaces that used to take in Node and were given ScopeStatements so that they now take in INode and are given IScopeNodes

CTrando: Might actually have to leave Name in IScopeNode. Otherwise for other nodes, I could redirect Name to NodeName.

jakebailey: Leaving stuff in IScopeNode is probably fine, we're just trying to eliminate changes that look backward compatible, or interfaces that could be avoided.

src/Parsing/Impl/Ast/ClassDefinition.cs
```diff
@@ -20,31 +20,43 @@
...
 
+        #region IScopeStatement
+
         public override Statement Body => _body;
 
+        #endregion
+
+        #region ScopeStatement
+
+        public override string Name => _name?.Name ?? string.Empty;
```
jakebailey: Not: Now that this is restored, you can probably move it back to where it was before. I don't think `Name` is specifically related to `ScopeStatement` and doesn't need to be in this region.

src/Parsing/Impl/Ast/PythonVariable.cs
```diff
@@ -29,12 +31,12 @@ public class PythonVariable {
         /// <summary>
         /// The scope the variable was declared in.
         /// </summary>
-        public ScopeStatement Scope { get; }
+        public IScopeNode Scope { get; }
```
jakebailey: Unfortunately this is probably going to be a breaking change. Not entirely certain how bad it'd be, since it seems required to make this work.

src/Parsing/Impl/Ast/ScopeStatement.cs
```diff
@@ -15,274 +15,35 @@
...
 using System.Collections.Generic;
-using System.Diagnostics;
-using Microsoft.Python.Core;
+using Microsoft.Python.Parsing;
 
 namespace Microsoft.Python.Parsing.Ast {
-    public abstract class ScopeStatement : Statement {
-        // due to "exec" or call to dir, locals, eval, vars...
+    public abstract class ScopeStatement : Statement, IScopeStatement {
+        public IScopeNode Parent { get; set; }
```
jakebailey: Similarly not backwards compatible. Unfortunate...

src/Parsing/Impl/Definition/IScopeStatement.cs
CTrando: OK

src/Parsing/Impl/Ast/ScopeStatement.cs
```diff
@@ -18,9 +18,15 @@
...
+    public abstract class ScopeStatement : Statement, IScopeNode {
+        private Statement _body;
+
+        public ScopeStatement() {
+            _body = new EmptyStatement();
+        }
+        
         public IScopeNode Parent { get; set; }
-        public abstract Statement Body { get; }
+        public virtual Statement Body => _body;
```
jakebailey: I think you could leave this as abstract and require every implementation to give a value, rather than adding state to the abstract base.

src/Parsing/Impl/Ast/PythonVariable.cs
CTrando: Yeah I don't see any other way to do it while being backwards compatible, unless you guys have any other ideas

src/Parsing/Impl/Ast/ScopeStatement.cs
jakebailey: And I think you should be fine to use null rather than an empty statement, but that's less important.

src/Parsing/Impl/Ast/PythonAst.cs
MikhailArkhipov: On the other hand it becomes confusing. `NodeName` != `Name`, consider `NameExpression.Name` vs `NameExpression.NodeName`

src/Analysis/Ast/Impl/Analyzer/Expressions/ExpressionFinder.cs
CTrando: Ok I'll do the redirect way

src/Parsing/Impl/Ast/ClassDefinition.cs
CTrando: Ok

src/Parsing/Impl/Ast/ScopeStatement.cs
CTrando: Ok

src/Parsing/Impl/Ast/PythonVariable.cs
MikhailArkhipov: The problem is that intellicode refactoring and intellicode extension to LS take dependency on the parser. We can't change it and just ask them to recompile. New LS MUST work with existing IC in the wild. I.e. any release of LS must be compatible with existing IC. IC can be rebuilt guilg forward, but there we can't make coordinated releases.

Basically, if you are making any changes to AST you have to
- install intellicode and verify it works in VS Code
- compile PTVS with new parser or at least coordinate with @huguesv wrt of the changes.

jakebailey: I think there's a way, actually. Keep all of the old ones as they were (type + name). Add new ones, and then make the old ones use the new ones. The old ones are concretely typed to be ScopeStatement, so you can traverse the new parent properties up until you hit a scope that is ScopeStatement, and return that.

For example, if something's parent scope is a list comprehension, in the old code it would not have been, but would have been the containing scope of the list, so now it'd walk upward and find the same thing.

Backwards compatible in the API as well as the returned value, I think.

jakebailey:
```csharp

public ScopeStatement Scope;
public IScopeNode ScopeNode;
```

Here, and then in ScopeStatement:

```csharp
public ScopeStatement Parent;
public IScopeNode ParentNode;
```

Or similar.

MikhailArkhipov: The problem is that intellicode refactoring and intellicode extension to LS take dependency on the parser. We can't change it and just ask them to recompile. New LS MUST work with existing IC in the wild. I.e. any release of LS must be compatible with existing IC. IC can be rebuilt guilg forward, but there we can't make coordinated releases.

Basically, if you are making any changes to AST you have to
- install intellicode and verify it works in VS Code
- compile PTVS with new parser or at least coordinate with @huguesv wrt of the changes.

jakebailey: I think there's a way to do it backwards compatibly.

Put the code back the way it was, and add a new property like `public IScopeNode ScopeNode`. Then make the old one use the new one by walking up the scopes until a scope node of type `ScopeStatement` is found. This means that any extra scopes your code introduces won't be exposed via the old API, so for old users, they'd get the same thing.

jakebailey: Then similarly for ScopeStatement, keep the old Parent, then a new ParentNode or something which does the same.

An idea to think about.

jakebailey: Ugh, github glitched and I thought it didn't send my comments, so I typed out two replies... Oops

CTrando: Ok I'll give it a shot

CTrando: I think we will still have to do the steps mikhail talked about, but I tried your suggested changes. It doesn't look quite as good but we can always revert it.

src/Parsing/Impl/Ast/FunctionDefinition.cs
```diff
@@ -19,21 +19,18 @@
...
 namespace Microsoft.Python.Parsing.Ast {
     [DebuggerDisplay("{Name}")]
-    public class FunctionDefinition : ScopeStatement, IMaybeAsyncStatement {
+    public class FunctionDefinition : ScopeStatement, IMaybeAsyncStatement, IScopeNode {
         internal static readonly object WhitespaceAfterAsync = new object();
 
         private int? _keywordEndIndex;
 
-        protected Statement _body;
+        private Statement _body;
```
jakebailey: This is technically an API break, since anyone could subclass FunctionDefinition. I wouldn't change any of the existing protection levels.

src/Parsing/Impl/Ast/ScopeStatement.cs
```diff
@@ -15,274 +15,47 @@
...
-                var cur = this;
-                while (!(cur is PythonAst)) {
-                    Debug.Assert(cur != null);
-                    cur = cur.Parent;
-                }
-                return (PythonAst)cur;
-            }
-        }
-
-        protected void Clear() {
```
jakebailey: Similarly protected here, so you'd need to preserve this.

src/Parsing/Impl/Ast/PythonVariable.cs
jakebailey: The testing? Definitely. Testing with IntelliCode first for sure.

src/Parsing/Impl/Ast/FunctionDefinition.cs
CTrando: OK

src/Parsing/Impl/Ast/PythonAst.cs
CTrando: Then yeah I left `Name` field in `IScopeNode` so it only has name if it actually has something different than node name

src/Parsing/Impl/Ast/PythonVariable.cs
CTrando: Could you walk me through on how to do that, I'm not really sure what you guys mean by what version of intellicode and how to compile PTVS.

src/Parsing/Impl/Ast/ScopeStatement.cs
CTrando: Done

src/Parsing/Impl/Ast/PythonVariable.cs
jakebailey: For IntelliCode, basically just install this and run the LS and see if completion breaks. https://marketplace.visualstudio.com/items?itemName=VisualStudioExptTeam.vscodeintellicode

CTrando: Ok I'll give it a shot

CTrando: So I tried it out and I just see 
```
Acquiring model 'intellisense-members' for python
Querying IntelliCode service for available models.
Cached model is up to date.
vs-intellicode-python was passed a model: {"modelPath":"/home/ctrando/.vscode/extensions/visualstudioexptteam.vscodeintellicode-1.1.9/cache/88630A1B4C077752096ECAA576B92C7D3FA1_BC98A65409D042A99DDAF47DC30E1530","modelId":"88630A1B4C077752096ECAA576B92C7D3FA1","outputId":"BC98A65409D042A99DDAF47DC30E1530"}.
Loaded language server extension.
```

jakebailey: You'd need to go and try out some completions that fit their model. You'll know they're being sorted if you see a star next to the items in the completion list.

src/Parsing/Impl/Ast/PythonNameBinder.cs
```diff
@@ -85,34 +90,35 @@ internal class ParameterBinder : PythonWalkerNonRecursive {
...
+
         public override bool Walk(NameExpression node) {
             _binder.DefineDeleted(node.Name);
             return false;
         }
     }
 
-    class PythonNameBinder : PythonWalker {
-        private ScopeStatement _currentScope;
+    public class PythonNameBinder : PythonWalker {
```
jakebailey: `public` got added here, but before it was the default, which I think is "internal", so maybe use that if you'd like to be explicit.

src/Parsing/Impl/Ast/PythonVariable.cs
CTrando: Ok I verified that I still get star stuff with my changes

jakebailey: That's excellent. I think you've worked this to be API backwards compatible.

src/Parsing/Impl/Ast/PythonNameBinder.cs
CTrando: ok

src/Parsing/Impl/Ast/PythonVariable.cs
CTrando: I can barely take any credit since you helped so much with it! But you should try and verify on your end as well. Let me know if there's anything else you want me to do.

src/Parsing/Impl/Ast/Comprehension.cs
```diff
@@ -45,6 +51,24 @@ public abstract class Comprehension : Expression {
                 res.Append(end);
             }
         }
+
+        #region IScopeNode
+
+        public virtual string Name => "<comprehension>";
+        public IScopeNode ParentNode { get; set; }
```
MikhailArkhipov: `ParentScopeNode` since `ParentNode` sounds like node that is right above in the tree which is not the case for scopes.

src/Parsing/Impl/Ast/Comprehension.cs
```diff
@@ -45,6 +51,24 @@ public abstract class Comprehension : Expression {
...
+
+        public Statement Body => null;
+        
+        public PythonAst GlobalParent => ScopeInfo.GlobalParent;
```
MikhailArkhipov: Why not just say `Ast` since it is the `PythonAst`

src/Parsing/Impl/Ast/FunctionDefinition.cs
```diff
@@ -19,21 +19,18 @@
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Python.Core;
-using Microsoft.Python.Core.Text;
 
 namespace Microsoft.Python.Parsing.Ast {
     [DebuggerDisplay("{Name}")]
-    public class FunctionDefinition : ScopeStatement, IMaybeAsyncStatement {
+    public class FunctionDefinition : ScopeStatement, IMaybeAsyncStatement, IScopeNode {
```
MikhailArkhipov: I assume `ScopeStatement` implements `IScopeNode` so `IScopeNode` here is unnecessary.

src/Parsing/Impl/Ast/FunctionDefinition.cs
```diff
@@ -94,87 +86,19 @@ public FunctionDefinition(NameExpression name, Parameter[] parameters)
...
-        }
+        #endregion
 
+        #region ScopeStatement
 
-        internal override void Bind(PythonNameBinder binder) {
-            base.Bind(binder);
-            Verify(binder);
-        }
+        public override string Name => NameExpression.Name ?? string.Empty;
```
MikhailArkhipov: Probably `ScopeName` - I guess it matches name of the function or class.

src/Parsing/Impl/Ast/ScopeInfo.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        /// <summary>
+        /// True if variables can be set in a late bound fashion that we don't
+        /// know about at code gen time - for example via from fob import *.
+        /// 
+        /// This is tracked independently of the ContainsUnqualifiedExec/NeedsLocalsDictionary
+        /// </summary>
+        internal virtual bool HasLateBoundVariableSets { get; set; }
+
+        internal Dictionary<string, PythonVariable> Variables { get; private set; }
```
MikhailArkhipov: read/write types normally should not be exposed

src/Parsing/Impl/Ast/ScopeInfo.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        /// <summary>
+        /// Gets the variables for this scope.
+        /// </summary>
+        public ICollection<PythonVariable> ScopeVariables => Variables?.Values ?? Array.Empty<PythonVariable>() as ICollection<PythonVariable>;
```
MikhailArkhipov: Since `Variables` is read/write, attempt to enumerate the collection may throw. Therefore it should be `Variables?.Values.ToArray()`. Also, You can change `ICollection` to `ImmutableArray<>` which is preferable. There is `ToImmutableArray()` extension.

src/Parsing/Impl/Ast/ScopeInfo.cs
```diff
@@ -0,0 +1,297 @@
...
+
+        internal virtual bool IsGlobal => false;
+
+        public PythonAst GlobalParent {
+            get {
+                var cur = Node;
+                while (!(cur is PythonAst)) {
```
MikhailArkhipov: If you added extension like `EnumerateTowardsGlobal` in `IScope`, you can do `cur.EnumerateTowardsGlobal.Last()`

src/Parsing/Impl/Ast/ScopeInfo.cs
```diff
@@ -0,0 +1,297 @@
...
+            _cellVars?.Clear();
+            _freeVars?.Clear();
+            _globalVars?.Clear();
+            _nonLocalVars?.Clear();
+        }
+
+        internal void AddFreeVariable(PythonVariable variable, bool accessedInScope) {
+            _freeVars = _freeVars ?? new List<PythonVariable>();
+            if (!_freeVars.Contains(variable)) {
+                _freeVars.Add(variable);
```
MikhailArkhipov: What about concurrent access? Can someone be accessing collection and something else adding to it?

src/Parsing/Impl/Ast/ScopeInfo.cs
```diff
@@ -0,0 +1,297 @@
...
+        /// Variables that are bound in an outer scope - but not a global scope
+        /// </summary>
+        public IReadOnlyList<PythonVariable> FreeVariables => _freeVars;
+
+        protected virtual bool ExposesLocalVariable { get; }
+
+        internal bool TryGetVariable(string name, out PythonVariable variable) {
+            if (Variables != null && name != null) {
+                return Variables.TryGetValue(name, out variable);
+            } else {
```
MikhailArkhipov: No need to have `else` if you return from `if`.

src/Parsing/Impl/Ast/ScopeInfo.cs
```diff
@@ -0,0 +1,297 @@
...
+            // Hide scope contents by default (only functions expose their locals)
+            variable = null;
+            return false;
+        }
+
+        internal abstract PythonVariable BindReference(PythonNameBinder binder, string name);
+
+        internal virtual void Bind(PythonNameBinder binder) {
+            if (_references != null) {
+                foreach (var refList in _references.Values) {
```
MikhailArkhipov: `_references.MaybeEnumerate().Values` then no need for `_references != null`

src/Parsing/Impl/Ast/ScopeInfo.cs
```diff
@@ -0,0 +1,297 @@
...
+                    foreach (var reference in refList) {
```
MikhailArkhipov: ` foreach(var reference in _references.MaybeEnumerate().Values.SelectMany(v => v)`.

src/Parsing/Impl/Ast/ScopeInfo.cs
```diff
@@ -0,0 +1,297 @@
...
+            }
+        }
+
+        internal void FinishBind(PythonNameBinder binder) {
+            List<ClosureInfo> closureVariables = null;
+
+            if (_nonLocalVars != null) {
+                foreach (var variableName in _nonLocalVars) {
+                    var bound = false;
+                    for (var parent = Node.ParentNode; parent != null; parent = parent.ParentNode) {
```
MikhailArkhipov: Might use extension like `EnumerateTowardsGlobal`

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
```diff
@@ -102,10 +104,8 @@ internal sealed partial class ExpressionEval {
         }
 
         public IMember GetValueFromComprehension(Comprehension node) {
-            var oldVariables = CurrentScope.Variables.OfType<Variable>().ToDictionary(k => k.Name, v => v);
-            try {
+            using (Interpreter.LanguageVersion.Is3x() ? OpenScope(Module, node) : Disposable.Empty) {
```
MikhailArkhipov: Better to hide check inside `OpenScope` by the node type so if we get another place that needs to call `OpenScope` it won't forget the check.

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
```diff
@@ -119,49 +119,40 @@ internal sealed partial class ExpressionEval {
...
-                            }
-                            break;
-                        case TupleExpression tex when value is IPythonCollection c2 && tex.Items.Count > 0:
-                            var iter = c2.GetIterator();
-                            foreach (var item in tex.Items.OfType<NameExpression>().Where(x => !string.IsNullOrEmpty(x.Name))) {
-                                DeclareVariable(item.Name, iter.Next, VariableSource.Declaration, item);
-                            }
-                            break;
+        public void ProcessComprehension(Comprehension node) {
+            using (Interpreter.LanguageVersion.Is3x() ? OpenScope(Module, node) : Disposable.Empty) {
```
MikhailArkhipov: Aha, there is second check...

src/Parsing/Impl/Ast/INode.cs
```diff
@@ -0,0 +1,44 @@
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Python.Core.Text;
+
+namespace Microsoft.Python.Parsing.Ast {
+    public interface INode {
```
MikhailArkhipov: public interfaces should not have setters, especially on things that do not change like indices which are positions in the buffer AST was produced from.

src/Parsing/Impl/Ast/INode.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        int StartIndex { get; set; }
+
+        IEnumerable<Node> GetChildNodes();
+        void Walk(PythonWalker walker);
+        Task WalkAsync(PythonWalkerAsync walker, CancellationToken cancellationToken = default);
+
+        string NodeName { get; }
+
+        string ToCodeString(PythonAst ast);
```
MikhailArkhipov: I would drop `ToCodeString` from public. It is buggy and we'd rather remove it eventually.

src/Parsing/Impl/Ast/INode.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </summary>
+        string GetLeadingWhiteSpace(PythonAst ast);
+
+        /// <summary>
+        /// Sets the proceeding whitespace (newlines and comments) that shows up
+        /// before this node.
+        /// </summary>
+        /// <param name="ast"></param>
+        /// <param name="whiteSpace"></param>
+        void SetLeadingWhiteSpace(PythonAst ast, string whiteSpace);
```
MikhailArkhipov: Probably not needed on public surface

src/Parsing/Impl/Definitions/IScopeNode.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// Holds information about the scope gathered from Parser, mainly used in  <see cref="PythonNameBinder"/>
+        /// </summary>
+        ScopeInfo ScopeInfo { get; }
+
+        /// <summary>
+        /// Binds the current scope, giving <param name="binder">binder</param> access to stored variables in <see cref="ScopeInfo"/>
+        /// </summary>
+        void Bind(PythonNameBinder binder);
```
MikhailArkhipov: `Bind` and `FinishBind` should not be on a public interface

src/Parsing/Impl/Definitions/IScopeNode.cs
```diff
@@ -0,0 +1,48 @@
...
+
+        /// <summary>
+        /// Gloabl scope
+        /// </summary>
+        PythonAst GlobalParent { get; }
+
+        /// <summary>
+        /// Holds information about the scope gathered from Parser, mainly used in  <see cref="PythonNameBinder"/>
+        /// </summary>
+        ScopeInfo ScopeInfo { get; }
```
MikhailArkhipov: Why `ScopeInfo` is a separate class? I.e. why not expose members here directly since every scope has `ScopeInfo` ?

src/Parsing/Impl/Ast/ScopeStatement.cs
```diff
@@ -15,274 +15,43 @@
...
-                Variables = new Dictionary<string, PythonVariable>(StringComparer.Ordinal);
-            }
-        }
+        public virtual string Name => "<unknown>";
 
-        internal void AddVariable(PythonVariable variable) {
-            EnsureVariables();
-            Variables[variable.Name] = variable;
-        }
+        public ICollection<PythonVariable> ScopeVariables => ScopeInfo.ScopeVariables;
```
MikhailArkhipov: Should be readonly collection, such as `ImmutableArray` or at least `IReadOnlyCollection`, we don't want to expose `Remove(T)`

src/Parsing/Impl/Definitions/IScopeNode.cs
MikhailArkhipov: Basic rules on interfaces:
- on public there should be no setters unless there is a valid scenario for the consumer outside of the analysis engine. We provide evaluator on analysis that can be used for basic operations such as declaring variables.
- on public interface assume there is concurrency. So if setter is exposed, it must be thread-safe.
- on any interface enumeration must be safe. I.e. public interface user must be able to safely enumerate collections without worrying if they can be modified by the user code directly or indirectly via API calls or by concurrency.
- Above mostly applies to internal interfaces as well - otherwise why have interfaces...

CTrando: For the first, I added the `ScopeInfo` class so that different ScopeNode implementations can have the same ScopeInfo types. For example comprehension and functions both handle scopes the same ways, so instead of having both implement the same methods and have the same code body, they can just share `FunctionScopeInfo`

CTrando: Also I'll try to get to your comments by EOD tomorrow, thanks for the comments!

src/Parsing/Impl/Definitions/IScopeNode.cs
CTrando: Should I make them extension methods?

jakebailey: The comment here is that they have side effects, so having them be a part of a public interface is a bad idea (when they're only used internally by the parser).

CTrando: I see, so how would I make it so that it can only be used internally by the parser? I don't think I can make the interface internal. I guess I could have them as internal methods on ScopeInfo and have callers use that instead, would that be ok? Also I may have asked this before but do you agree on the idea of `ScopeInfo` or feel that I should extract the members to the `IScopeNode` interface?

CTrando: Another benefit of having `ScopeInfo` is that I can leave it in the interface and define internal methods on it so users of our assemblies cannot access them. If I had them on the interface then everyone would have access to those methods only used for the parser (assuming that there is not some way to have internal methods on a public interface)

MikhailArkhipov: Internally they should be used via the class itself. It is possible to have separate internal interface too, we have quite a few. ScopeInfo technically should be `IScopeInfo` as well... But I don't like much separate 'info'. Separate entity makes sense if it is passed around as a separate object or used in other places separately. If it is only used as part of an interface, then just promote methods up to the interface. Occam Razor :-)

jakebailey: Yeah, I wouldn't leave anything in the interface if it's not usable. Having something that's completely impossible to construct is a weird interface. If binding is something that's required for the parser to work but isn't generally usable outside (I think that's how it is), then it'd be better to have some separate internal interface or something.

CTrando: I see I think I'll add an internal interface then, perhaps `IBindable`. I'll try moving all the members up to the interface and moving the internal methods to the internal interface but it will take me some more time. Thanks for the help.

CTrando: I'm a little worried about code duplication however. Comprehensions should be treated like functions in terms of scoping, but without composition of their behavior then Comprehension will have to implement the `bind` and `finishBind` methods as well with the same code, do you have any suggestions for this?

jakebailey: I don't think we're suggesting duplicating the code all over the place. It's fine internally for the nodes to store some common thing which implements binding, but then expose that info through a common interface (e.g. IScopeNode).

CTrando: Ok now I understand. The issue is having ScopeInfo on a public interface, not the existence of ScopeInfo itsel. ScopeInfo can be part of the implementation details and its information can be exposed through the IBindable and IScopeNode interfaces. 

I’ll have to make FunctionDefinition, ClassDefinition and the like implement another internal interface because IScopeNode cannot implement an internal interface because it is public. 

I’ll get on it though thanks for the advice.

CTrando:
Could you look over the brief changes I did and see if it matches what you meant? I had some issue with a public interface also having internal members so I had to explicitly implement the internal interface. It doesn't compile either so don't worry about that :smile:

jakebailey:
My immediate reaction is to say not to us the word "delegate", which has a very specific meaning in C#.

src/Parsing/Impl/Definitions/INode.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// </summary>
+        string GetLeadingWhiteSpace(PythonAst ast);
+
+        /// <summary>
+        /// Sets the proceeding whitespace (newlines and comments) that shows up
+        /// before this node.
+        /// </summary>
+        /// <param name="ast"></param>
+        /// <param name="whiteSpace"></param>
+        void SetLeadingWhiteSpace(PythonAst ast, string whiteSpace);
```
MikhailArkhipov: Probably not for public interface

src/Parsing/Impl/Definitions/INode.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        SourceLocation GetEnd(PythonAst ast);
+
+        SourceSpan GetSpan(PythonAst ast);
+
+        /// <summary>
+        /// Returns the proceeding whitespace (newlines and comments) that
+        /// shows up before this node.
+        /// 
+        /// New in 1.1.
```
MikhailArkhipov: Comment about version is obsolete

src/Parsing/Impl/Ast/ScopeStatement.cs
```diff
@@ -13,276 +13,82 @@
...
-            public ClosureInfo(PythonVariable variable, bool accessedInScope) {
-                Variable = variable;
-                AccessedInScope = accessedInScope;
-            }
-        }
+        public IReadOnlyList<PythonVariable> FreeVariables => ScopeDelegate.FreeVariables;
+        #endregion
+
+        #region IBindableNode
+        void IBindableNode.Bind(PythonNameBinder binder) => ScopeDelegate.Bind(binder);
```
MikhailArkhipov: Normally we don't use explicit interface implementations. Is there anything that conflicts with the interface members?

CTrando: I read that when we have a public class that implements an internal interface, the methods from the internal interface are still available. I tried this by checking out some of the methods from the `LanguageServer` project and I could still access those methods. When I did explicit implementation then those methods were hidden. If you want I could switch them back to implicit implementation.

jakebailey: That's a good observation.

jakebailey:
Just a general note, but there's a merge conflict with this PR, so you may want to try and merge and fix some of the conflicts.

src/Parsing/Impl/Ast/ScopeStatement.cs
MikhailArkhipov: if class is public and method is public the method is, of course, accessible. Having it on an internal interface does not change this fact.

jakebailey: I think that's what he's saying, but by explicitly implementing the interface he's not actually exporting those names unless the class is used as the interface. (i.e. https://stackoverflow.com/a/6117670 but in our case the class is public, not internal)

The only other way I can think of to keep "IBindableNode" completely internal would be to have some internal abstract class that implements IBindableNode which each of these classes inherit from and complete...

MikhailArkhipov: Ah, OK

CTrando:
Yes I will do that!

src/Parsing/Impl/Ast/INode.cs
CTrando: Yes!

src/Parsing/Impl/Ast/INode.cs
CTrando: ok!

src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Collections.cs
CTrando: great, I'll do that

src/Parsing/Impl/Ast/ScopeStatement.cs
CTrando: Yes!

src/Parsing/Impl/Definitions/INode.cs
CTrando: yes!

src/Parsing/Impl/Definitions/INode.cs
CTrando: yes!

src/Parsing/Impl/Ast/ScopeInfo.cs
CTrando: ok I gave it a shot

src/Parsing/Impl/Ast/ScopeInfo.cs
CTrando: ok!

src/Parsing/Impl/Ast/ScopeDelegate.cs
```diff
@@ -0,0 +1,259 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Microsoft.Python.Core;
+using Microsoft.Python.Core.Collections;
+using Microsoft.Python.Core.Diagnostics;
+
+namespace Microsoft.Python.Parsing.Ast {
+    internal abstract class ScopeDelegate {
```
jakebailey: I still think that `Delegate` is the wrong name to use for these. Delegates are a thing in C# already. I think `Info` was the better suffix.

CTrando: Ok I'll rename it

jakebailey:
One thing that might need to be tested here is `:=`. If comprehensions are now scopes, then when `:=` does its assignment, it will assign its value to the wrong scope, as the "correct" scope is the first non-comprehension scope walking up the scope stack, typically a function/lambda/global scope.

CTrando:
I'll read over PEP 572. Also about the concurrency issues mentioned before, I don't know if those will be a big issue because only PythonNameBinder interacts with it and if each module is walked separate from each other by the PythonNameBinder I don't see anything getting messed up.

jakebailey:
To be clear, I don't mean in the parser (I already implemented that in #1586), I mean in the analyzer when it goes and tries to do the assignment and has to pick the scope.

Arguably we should be combining this scoping logic, but it's more likely that we remove some of it from the parser (along with other changes) and just use the analysis for everything.

CTrando:
What would a sufficient example be just to clarify what you mean.

I have:

```
def f(a):
    return a

def test():
    h = [y := f(2), y**2, y**3]

test();
```
And y is defined in the test scope.

jakebailey:
Yes, and also nesting like:

```python
def foo(lst):
    l = [[[(a := y) for y in z] for z in x] for x in lst]]]
    # a here

# no a here
```

But also:

```python
def foo(lst):
    l = lambda x: (a := x) + 1
    # no a here

# no a here
```

My worry was that we'd be seeing the comprehension scope, then stopping there, but if you've got it pinned down to where the scoping is only used to declare iteration variables, then that's great.

CTrando:
So what was happening is that we didn't actually handle comprehension scopes in analysis (no scope was being created for them so setting a variable y := something inside a nested comprehension caused the variable to have type unknown.

In order to fix this, I am debating having another member evaluator for comprehensions in the like. I tried adding a `PopScope` method to `Eval.Scopes` so I can call `OpenScope` when walking a comprehension and `PopScope` when post walking, but this method seems clumsy. 

Furthermore, when the scopes work correctly I have to worry about how to travel up the scope stack and declare a variable in a specific scope. I was thinking of passing in the scope I wanted to declare a variable in, but that would require changing a lot of method signatures and I don't think it's the cleanest. Another way to do it is to pop scopes until I find a non comprehension scope then add them back after I declare variables in that scope. (I could do this with disposable to help avoid invalid states). 

Any thoughts on this? This change is becoming very large, I hope you guys are still ok with it.

jakebailey:
How we handle and evaluate scopes is something we should be looking into, namely to try to not have the behavior of declaring parameters/iteration values in a scope then trying to delete them (which breaks with concurrency).

`:=` inside of comprehensions is a bit esoteric and the feature is only a week old. I'd personally prioritize getting scoping working, then try and get `:=` working right. Not being able to traverse up the scope stack confuses me, since each scope necessarily has to know what its parent is. I think we're in Eval already which does the declaration, so adding an optional "target scope" argument to the assignment function seems okay.

I'd probably try and go for mostly parser changes (plus the replacement of ScopeStatement with IScopeNode as you've done) and see how we stand in some tests.

CTrando:
So let me clarify.

In 

‘’’
def foo(lst):
    l = [[[(a := y) for y in z] for z in x] for x in lst]]]
    # a here

# no a here
‘’’

a exists in the function scope and doesn’t exist in the outer scope however the type of a in the function scope is unknown.

This is because in the analysis in FunctionEvaluator, we walk the function body and don’t open the scope of the comprehensions, so when it comes time to declare the variable a we cannot find the type of y because we are in the wrong scope.

So to clarify, is this something you want me to address in this PR? I can just as well say the existence of a in the function scope is good enough in the test and not check the test.

Otherwise I’ll have to figure out a way to open the comprehension scope when walking the function body so we know what type to assign the named expression variable.

jakebailey:
If it already doesn't work now, then I would not worry about fixing it. I was unsure if it did, and if it doesn't, then it's just more code in the same PR which I wouldn't want to do. I just don't want the additions of more scopes to cause a regression in some feature.

src/Parsing/Impl/Definitions/IBindableNode.cs
```diff
@@ -0,0 +1,59 @@
...
+        PythonVariable EnsureGlobalVariable(string name);
+
+        PythonVariable DefineParameter(string name);
+        #endregion
+
+        #region Contains
+        bool ContainsImportStar { get; set; }
+        bool ContainsExceptionHandling { get; set; }
+
+        bool ContainsUnqualifiedExec { get; set; }
```
CTrando: Can I expose these in `IScopeNode`? `IScopeNode` already has a few members similar to these and it would simplify the code in `PythonNameBinder` because I wouldn't have to cast.

jakebailey: What casts are you referring to? I don't see any more around these specific properties than existed previously in the binder.

jakebailey: I would prefer to about it, as it's more of an implementation detail of the parser/binder rather than a property of the scope.

CTrando: I created a new property `BoundScope` which casts the scope to an `IBindableNode`. But I'll leave it in there if you feel it's best.

CTrando:
Can I get some more feedback? Was thinking should probably get this in soon if that's alright with you guys.

jakebailey:
You're going to have to fix the merge conflicts (to start), but I can give this a final test and review.

CTrando:
I know it's been a while but could I get a final review if one of you have some time?

src/Analysis/Ast/Test/FluentAssertions/ScopeAssertions.cs
```diff
@@ -53,21 +55,49 @@ internal class ScopeAssertions<TScope, TScopeAssertions> : ReferenceTypeAssertio
...
+
+            return new AndWhichConstraint<TScopeAssertions, List<IScope>>((TScopeAssertions) this, scopeMatch);
         }
 
         public AndWhichConstraint<TScopeAssertions, IPythonClassType> HaveClass(string name, string because = "", params object[] reasonArgs) {
             var v = HaveVariable(name, because, reasonArgs).Which;
             v.Value.Should().BeAssignableTo<IPythonClassType>();
 
-            return new AndWhichConstraint<TScopeAssertions, IPythonClassType>((TScopeAssertions)this, (IPythonClassType)v.Value);
+            return new AndWhichConstraint<TScopeAssertions, IPythonClassType>((TScopeAssertions) this, (IPythonClassType) v.Value);
```
jakebailey: Super extreme nit, but these cast spacing changes appear to be an editor setting difference.

src/LanguageServer/Test/CompletionTests.cs
```diff
@@ -838,8 +834,8 @@ class x(object):
 
             result = cs.GetCompletions(analysis, new SourceLocation(3, 22));
             result.Should().HaveLabels("import")
-                .And.NotContainLabels("abc", @"unittest", "abs", "dir")
-                .And.Subject.ApplicableSpan.Should().Be(3, 20, 3, 26);
+                  .And.NotContainLabels("abc", @"unittest", "abs", "dir")
```
jakebailey: Another style nit, but a lot of files are getting changed with spacing things. You may want to see what settings are missing to prevent this.

src/Parsing/Impl/Extensions/AstExtensions.cs
```diff
@@ -20,16 +20,16 @@
 
 namespace Microsoft.Python.Parsing.Extensions {
     public static class AstExtensions {
-        public static IEnumerable<Node> ChildNodesDepthFirst(this Node node)
+        public static IEnumerable<INode> ChildNodesDepthFirst(this INode node)
```
jakebailey: I'm not sure that this is backwards compatible, since this changes the signature of non-internal stuff. Do we need to retain the `Node` variants? @MikhailArkhipov

jakebailey:
Other than style nits and my question about the public extensions that were previously on `Node`, I don't have any objections to this. Very happy to see this done compatibly with only new tests.

Thanks again for doing this and sorry about the review delay.

CTrando:
No worries, partially my fault because I got busy with school. Hope the style changes are correct, I'll run the tests on my end but I suggest you do as well just to be sure.

src/Parsing/Impl/Extensions/AstExtensions.cs
jakebailey: @heejaechang says this isn't backwards compatible. You'll have to bring back the old definitions, it seems.

heejaechang: it is changing signature which is explicitly stated in IL, runtime will throw method not found exception.

CTrando: So do you think I should just duplicate the methods but have them for node as well?

jakebailey: Yes, you can just leave the old code and also have the new code. Copy/paste. I tested this and the compiler seemed fine with it.

CTrando: Alright sounds good!

PR #1580, opened 2019-09-24
Add back the State property getter to the Tokenizer.

huguesv:
PTVS needs to initialize the Tokenizer with an existing state, but that property was removed in January, probably because that looked unused in the language server code base.

### DIFF
--- a/src/Parsing/Impl/Tokens/Tokenizer.cs
+++ b/src/Parsing/Impl/Tokens/Tokenizer.cs
@@ -111,6 +111,12 @@ namespace Microsoft.Python.Parsing {
             return tokens;
         }
 
+        /// <summary>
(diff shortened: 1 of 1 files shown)


src/Parsing/Impl/Tokens/Tokenizer.cs
```diff
@@ -111,6 +111,12 @@ public sealed class Tokenizer {
             return tokens;
         }
 
+        /// <summary>
+        /// Current state, to be passed to <see cref="Initialize(object, TextReader, SourceLocation)"/>
+        /// or <see cref="Initialize(object, TextReader, SourceLocation, int)"/>./>
```
jakebailey: Can you make a note about why this is needed in PTVS so we don't accidentally remove it again? (Sorta like what you did in #1541.)

PR #1581, opened 2019-09-24
use LinkedList instead of List in PriorityProducerConsumer

heejaechang:
simple clean up. PriorityProducerConsumer was using List but removed element from head. but never took advantage of indexer so changed it to use LinkedList.

### DIFF
--- a/src/Core/Impl/Extensions/TaskCompletionSourceExtensions.cs
+++ b/src/Core/Impl/Extensions/TaskCompletionSourceExtensions.cs
@@ -18,8 +18,10 @@ using System.Threading.Tasks;
 
 namespace Microsoft.Python.Core {
     public static class TaskCompletionSourceExtensions {
+        private const int NO_DELAY = -1;
+
         public static CancellationTokenRegistration RegisterForCancellation<T>(this TaskCompletionSource<T> taskCompletionSource, CancellationToken cancellationToken) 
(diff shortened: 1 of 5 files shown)


src/Core/Impl/Extensions/TaskCompletionSourceExtensions.cs
```diff
@@ -18,8 +18,10 @@
 
 namespace Microsoft.Python.Core {
     public static class TaskCompletionSourceExtensions {
+        private const int NO_DELAY = -1;
```
AlexanderSher: AFAIK, we don't use all caps notation for constants.

heejaechang: @AlexanderSher how should I change it? no_delay then?

heejaechang: changed to lower case.

src/Core/Impl/Extensions/TaskCompletionSourceExtensions.cs
```diff
@@ -18,8 +18,10 @@
 
 namespace Microsoft.Python.Core {
     public static class TaskCompletionSourceExtensions {
+        private const int no_delay = -1;
```
jakebailey: Too late, but probably should have been `NoDelay` for style consistency.

Issue #1576, opened 2019-09-23
NRE in MostRecentDocumentSymbols


CMLL:
## Environment data

- Language Server version: 0.4.22   
- OS and version: Ubuntu 19.04
- Python version (& distribution if applicable, e.g. Anaconda): 2.7.15

## Expected behaviour

Completion completes.

## Actual behaviour

The error happens when doing regular completion on a file, the logs shows the error related to caching. Tried removing the cache folder in `~/snap/common/.cache/Microsoft` but didn't help. Also changed the `python.analysis.cachingLevel` to `None` and `Library` but issue was still showing.

The first traceback shows when the cachingLevel is set to `Library`, second traceback is when I remove the `python.analysis.cachingLevel` fromt the options and let it have the default value or `None`.

## Logs

```
Analysis of models.core.server (User) queued. Dependencies: datetime, dateutil.parser, sqlobject, turbogears, turbogears.identity, turbogears.visit, turbogears.config, urlparse, ipaddress, itertools, models.masterdb, models.core.server_facts, models.core.metric_category, models.core.network_service_discovery, models.core.resource_matcher, models.core.snmp_traps, models.core.custom_incident_filter, models.core.snmp_port_discovery, models.core.countermeasures, models.core.maintenance_period, models.core.delta, models.core.elasticflow, models.core.outage_escalation, models.core.cloud, models.core.monitornode, models.core.availability, tasks.ops, pprint, models.core.dashboard, tasks.monitoring_policy
[Error - 3:19:07 PM] Request textDocument/documentSymbol failed.
  Message: One or more errors occurred. (Object reference not set to an instance of an object.)
  Code: -32000 
   at Microsoft.Python.LanguageServer.Indexing.MostRecentDocumentSymbols.<>c.<GetSymbolsAsync>b__11_0(Task`1 t) in E:\A\_work\8\s\src\LanguageServer\Impl\Indexing\MostRecentDocumentSymbols.cs:line 81
   at System.Threading.Tasks.ContinuationResultTaskFromResultTask`2.InnerInvoke()
   at System.Threading.ExecutionContext.RunFromThreadPoolDispatchLoop(Thread threadPoolThread, ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.Tasks.Task.ExecuteWithThreadLocal(Task& currentTaskSlot, Thread threadPoolThread)
[...]

PR #1582, opened 2019-09-25
[WIP] simplify priority producer consumer by using semaphoreSlim directly t…

heejaechang:
…o manage resource

### DIFF
--- a/src/Core/Test/PriorityProducerConsumerTest.cs
+++ b/src/Core/Test/PriorityProducerConsumerTest.cs
@@ -178,8 +178,10 @@ namespace Microsoft.Python.Core.Tests {
                 for (var i = 0; i < 3; i++) {
                     var task = ppc.ConsumeAsync();
                     tcsConsumer.TrySetResult(true);
-                    values[i] = await task;
+
+                    // make sure wait for producer first to remove race
                     await tcsProducer.Task;
+                    values[i] = await task;
                 }
             });
 
@@ -194,8 +196,10 @@ namespace Microsoft.Python.Core.Tests {
             await consumerTask;
 
             Assert.AreEqual(TaskStatus.RanToCompletion, consumerTask.Status);
-            Assert.AreEqual(5, values[0]);
-            Assert.AreEqual(7, values[1]);
+
+            // return values follow priority
+            Assert.AreEqual(7, values[0]);
+            Assert.AreEqual(5, values[1]);
             Assert.AreEqual(6, values[2]);
         }
 
--- a/src/Core/Impl/Threading/PriorityProducerConsumer.cs
+++ b/src/Core/Impl/Threading/PriorityProducerConsumer.cs
@@ -25,8 +25,8 @@ namespace Microsoft.Python.Core.Threading {
     public sealed class PriorityProducerConsumer<T> : IDisposable {
         private readonly int _maxPriority;
         private readonly object _syncObj;
+        private readonly SemaphoreSlim _pendingTasks;
         private readonly LinkedList<T>[] _queues;
-        private readonly Queue<Pending> _pendingTasks;
         private readonly DisposeToken _disposeToken;
         private readonly bool _excludeDuplicates;
         private readonly IEqualityComparer<T> _comparer;
@@ -51,7 +51,8 @@ namespace Microsoft.Python.Core.Threading {
             }
 
             _syncObj = new object();
-            _pendingTasks = new Queue<Pending>();
+            _pendingTasks = new SemaphoreSlim(initialCount: 0);
+
             _firstAvailablePriority = _maxPriority;
             _disposeToken = DisposeToken.Create<PriorityProducerConsumer<T>>();
             _excludeDuplicates = excludeDuplicates;
@@ -61,8 +62,8 @@ namespace Microsoft.Python.Core.Threading {
         public void Dispose() {
             if (_disposeToken.TryMarkDisposed()) {
                 lock (_syncObj) {
-                    Debug.Assert(_pendingTasks.Count == 0 || _firstAvailablePriority == _maxPriority);
-                    _pendingTasks.Clear();
+                    _pendingTasks.Dispose();
+
                     foreach (var queue in _queues) {
                         queue.Clear();
                     }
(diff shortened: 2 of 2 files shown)


heejaechang:
this change one behavior of the producer so, it is marked WIP

src/Core/Test/PriorityProducerConsumerTest.cs
```diff
@@ -178,8 +178,10 @@ public class PriorityProducerConsumerTest {
                 for (var i = 0; i < 3; i++) {
                     var task = ppc.ConsumeAsync();
                     tcsConsumer.TrySetResult(true);
-                    values[i] = await task;
+
+                    // make sure wait for producer first to remove race
                     await tcsProducer.Task;
+                    values[i] = await task;
```
heejaechang: this is where behavior got changed. previously if ConsumerAsync is called when the queue is empty, caller will hold onto TaskCompletionSource<T>.Task of pendingTcs directly

```csharp
                pendingTcs = new TaskCompletionSource<T>(TaskCreationOptions.RunContinuationsAsynchronously);	
                pending = new Pending(pendingTcs, _syncObj);	
                _pendingTasks.Enqueue(pending);	
            }	

            return pendingTcs.Task;
```

and when finally something enqueued through Produce

```csharp
            pendingTcs?.TrySetResult(value);
```

code set the value explicitly to the source, making the ConsumerAsync caller to get that value right away.

that is why existing test get the value in this order

```csharp
			Assert.AreEqual(5, values[0]);	
            Assert.AreEqual(7, values[0]);
            Assert.AreEqual(6, values[2]);
```

with the change, now ConsumeAsync always gets value in priority order among values at the moment in the queue, so now it gets in order of 7,5,6.

also, I moved await in the test to make sure test retrieve values once all Produce is called otherwise, we could have nondeterministic result since values are based on priority at the moment in the queue.

```csharp
	await tcsProducer.Task;
```

AlexanderSher: Ok, so here is the behavior of the original test:
- Step 1: one consumer task is registered, no tasks in the queue
- Step 2: item with priority 1 pushed to consumer. Since there is no item with priority 0, it is set to consumer. But consumer awaits for `tcsConsumer.Task`, so no new waiter are registered
- Step 3: items with priority 1 and 0 are registered, and `tcsProducer` is set
- Step 4: `consumerTask` drains all the items with respect of the order

This is an expected behavior, cause in ppc use cases, tasks with lower priority happen more often. For instance, in Language Server, initialization tasks have priority 0, configuration change tasks have priority 1, and document change tasks have priority 2.

heejaechang: @AlexanderSher 
so, consumer task that created when queue is empty, must get the very first item added to the queue when it was empty? is that an explicit contract or implementation contract code assumes in other parts of code base?

basically, what I am asking is if I change that behavior, will it break something?

the change I made basically make "consumer task" that created when queue is empty, will get an item in the queue once someone put items in the queue based on priority

AlexanderSher: > basically, what I am asking is if I change that behavior, will it break something?

**If** I correctly understand your changes, it will be a breaking change. PPC is used in LS to prioritize requests when they come faster than we process them, but when the queue is empty, consumer awaits for any incoming request, e.g. it doesn't wait for request with priority 0 to start processing requests with priority 1 and 2.

heejaechang: @AlexanderSher the change still doesn't wait for priority 0? it just wont special case the empty case anymore. the queue will pick up next time in the queue and process it as all other cases.

in reality, how the system knows this specific situation?

MikhailArkhipov:
I would like to learn reasoning behind changes. PPC has been working OK so what is that we want to improve or fix?

heejaechang:
@MikhailArkhipov I am just reading through code and clean up here and there to understand code base better.

fine not taking this PR.

that being said, my intention was trying to make implementation simpler. it looked like if we use semaphoreslim's natvie WaitAsync ability, we could remove our own TaskCompletionSource work since SemanphoreSlim basically does it for us.

MikhailArkhipov:
Since we don't have tests in this high traffic area, I think first step would be to add some for the current behavior so we don't change it to something different and possibly affecting user experience.

heejaechang:
@MikhailArkhipov sure I can add more tests. it looks like there are some tests specifically for this type. and it looks like 2 components (LS queue and dependency resolver) are using it. so I can add tests over those 2 components as well.

MikhailArkhipov:
Primary behavior as @AlexanderSher mentioned is
1. In a sequence of requests higher priority are always processed first.
2. Only one request is allowed to be processed at a time.
3. If there are no high priority requests, lower priority requests do not wait and are processed immediately.
4. On shutdown pending requests are dropped.

MikhailArkhipov:
It might be actually a good thing since we have some anecdotal evidence and hard to repro bugs like #1576 that requests may be getting processed out of order.

heejaechang:
@MikhailArkhipov 
> In a sequence of requests higher priority are always processed first.
this is still the behavior. 

> Only one request is allowed to be processed at a time.
I assume you meant queue is serialized? since it returns task, there is no guarantee that returned task is shared by 2 people "awaiting" on it. or caller call ConsumeAsync twice and hold onto the task and await on those tasks in reversed order. 

that being said, order of queue process is still preserved and handled one by one through lock.

> If there are no high priority requests, lower priority requests do not wait and are processed immediately.
this is still the behavior.

> On shutdown pending requests are dropped.
yep. same behavior.

…

about process ordering.

I see "SingleThreadSynchronizationContext" and "Prioritizer" that looks like trying to serialize processing request. are they not working as expected?

jakebailey:
`SingleThreadSynchronizationContext` from JsonRPC means that the requests are guaranteed to be single threaded _until the first await_, not that they're handled single threaded altogether. This gives us the freedom to do some exclusive work just before any async code, i.e. the prioritizer.

MikhailArkhipov:
So, to reiterate my points:
- Prioritizer organizes requests in a proper sequence and makes sure one request is handled by the LS code at a time.
- If there are bugs (or we _suspect_ there are) we should build a body of tests first to make sure everything works as expected after changes. Tests may also prove that there are indeed bugs.
- If there are no known bugs, we'd rather leave it alone since it is a high traffic sensitive piece which took a while to make working.

heejaechang:
@MikhailArkhipov sure.

PR #1584, opened 2019-09-25
Fix module cycle handling

AlexanderSher:
This change modifies the way we handle cycle dependencies of modules.
Previously, after finding the loop, we have determined some random path to walk over those modules and walked twice over it. With new approach, we first collect information about location of imports and variables in global scope of the modules in the loop, and then try to find a starting module. After that, we do the analysis starting from that module. Analyzes for modules are saved only after they are obtained for all modules in the loop.


AlexanderSher:
Please don't merge the PR until 9/30, I'd like to run the tests on my main machine.

MikhailArkhipov:
Most probably conflicts with https://github.com/microsoft/python-language-server/pull/1565 which changed how variables/members are exposed from variable modules.

MikhailArkhipov:
In the PR branch:
- Running analysis tests, `LibraryTests.Requests` failed consistently.
- In the app tests in full-cache mode `"python.analysis.cachingLevel": "Library"` only some of the numpy modules were saved so on reload analysis still took time.
- In partial-cache mode `"python.analysis.cachingLevel": "System"` results do appear consistently the same.

src/Analysis/Ast/Impl/Analyzer/Handlers/LoopImportedVariableHandler.cs
```diff
@@ -0,0 +1,72 @@
...
+            var key = new AnalysisModuleKey(module);
+            if (_walkers.TryGetValue(key, out var walker)) {
+                return walker.Eval.GlobalScope?.Variables[name];
+            }
+            
+            if (!_asts.TryGetValue(key, out var ast)) {
+                return _cachedVariables.TryGetValue(key, out var variables) ? variables[name] : default;
+            }
+            
+            walker = CreateWalker(module, ast);
```
MikhailArkhipov: This is equivalent to running analysis, no?

MikhailArkhipov: If this is just to get exports, the other day we discussed with @jakebailey possibility of running symbol collector instead to get variables like classes and functions but without evaluating bodies.

src/Analysis/Ast/Impl/Analyzer/Handlers/SimpleImportedVariableHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Microsoft.Python.Analysis.Analyzer.Handlers {
+    internal sealed class SimpleImportedVariableHandler : IImportedVariableHandler {
+        public static IImportedVariableHandler Instance { get; } = new SimpleImportedVariableHandler();
+
+        private SimpleImportedVariableHandler() {}
+        
+        public IVariable GetVariable(in PythonVariableModule module, in string name) =>
+            module.Analysis?.GlobalScope?.Variables[name];
+    }
```
MikhailArkhipov: https://github.com/microsoft/python-language-server/pull/1565

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerSession.cs
```diff
@@ -233,26 +241,50 @@ internal sealed class PythonAnalyzerSession {
...
-            => ExecutionContext.Run(ExecutionContext.Capture(), s => Analyze(node, null, stopWatch), null);
+                bool isCanceled;
+                lock (_syncObj) {
+                    isCanceled = _isCanceled;
+                }
 
-        private Task StartAnalysis(IDependencyChainNode<PythonAnalyzerEntry> node, AsyncCountdownEvent ace, Stopwatch stopWatch)
-            => Task.Run(() => Analyze(node, ace, stopWatch));
+                if (!isCanceled) {
+                    _progress.ReportRemaining(_walker.Remaining);
```
MikhailArkhipov: Having `_walker.Remaining` outside the lock caused occasional NREs

MikhailArkhipov:
Reopen from another branch

PR #1585, opened 2019-09-25
Prevent document reload with old content

MikhailArkhipov:
Found another case of buffer corruption. Happens when module reload comes during analysis. Repro:

0. Remove `black` formatter if you have one installed.
1. Open `scikit-learn`
2. Open `conftest.py`
3. While analysis is running, make changes like add space after dot in `del sys._is_pytest_session`
4. Format document
5. When prompted to install `black` say yes.
6. When formatter is installed and analysis still running, format document.

The buffer is now corrupted since changes comes over content reloaded by `ReloadAll`.

- Separated `Reset` into `Populate`, `MarkChanged` and `Clear`.
- Added exceptions to make sure `Populate` only happens once and there is no `Update` after `Clear`.
- `MarkChanged` advances version so analysis thinks buffer has changed without actually reloading it.

### DIFF
--- a/src/Analysis/Ast/Impl/Documents/DocumentBuffer.cs
+++ b/src/Analysis/Ast/Impl/Documents/DocumentBuffer.cs
@@ -13,7 +13,9 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
+using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using Microsoft.Python.Parsing;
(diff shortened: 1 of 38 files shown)


src/Analysis/Ast/Impl/Documents/DocumentBuffer.cs
```diff
@@ -59,12 +110,21 @@ internal sealed class DocumentBuffer {
 
                     var start = NewLineLocation.LocationToIndex(lineLoc, change.ReplacedSpan.Start, _sb.Length);
                     var end = NewLineLocation.LocationToIndex(lineLoc, change.ReplacedSpan.End, _sb.Length);
+
+                    //Debug.Assert(end >= start);
```
jakebailey: Do you want to remove these before merging? Seems like from debugging.

MikhailArkhipov: Maybe... They are useful to debug changes that's why I left them behind. I can nix them.

MikhailArkhipov:
Will reopen from another branch

PR #1586, opened 2019-09-25
Add syntax errors for `:=` operator scoping

### LINKED ISSUE #844: Expose syntax errors from := with a bad scope target
Defined in PEP 572 https://www.python.org/dev/peps/pep-0572/

jakebailey:
Fixes #844.

This works similarly to other walkers, like the name binder. Arguably, this work should be done in the name binder, but as far as I can tell, the name binder doesn't handle comprehension scoping correctly anyway (and we don't bother to use that info, choosing to reimplement scoping in the analysis anyway).

In the future, we might be able to move this step to the name binder, but probably not without #1567.

I added some useful AST node extensions for enumerating child nodes as well.


jakebailey:
Changed to WIP, need to handle this case:

```python
class LambdaTop:
	[(lambda: (z := x)) for x in range(1)]
```

jakebailey:
Redone using the walker to traverse everything (except places where it's known that only names can appear). Should be ready for a look.

src/Parsing/Impl/Resources.resx
```diff
@@ -258,6 +258,18 @@
   <data name="NamedExpressionCtxtErrorMsg" xml:space="preserve">
     <value>Named expression must be parenthesized in this context</value>
   </data>
+  <data name="NamedExpressionInClassBodyErrorMsg" xml:space="preserve">
+    <value>assignment expression within a comprehension cannot be used in a class body</value>
```
MikhailArkhipov: First letter capitalization?

src/Parsing/Impl/Resources.resx
```diff
@@ -258,6 +258,18 @@
...
+  </data>
+  <data name="NamedExpressionInComprehensionIteratorErrorMsg" xml:space="preserve">
+    <value>assignment expression cannot be used in a comprehension iterable expression</value>
+  </data>
+  <data name="NamedExpressionIteratorRebindsNamedErrorMsg" xml:space="preserve">
+    <value>comprehension inner loop cannot rebind assignment expression target '{0}'</value>
```
MikhailArkhipov: Some capitalized, some not

src/Parsing/Impl/Resources.resx
jakebailey: These are all the same as cpython, which doesn't capitalize at the start. The named ones that exist are the ones that are inconsistent (which I can fix).

PR #1587, opened 2019-09-26
Fix find references regression

MikhailArkhipov:
https://github.com/microsoft/python-language-server/pull/1565 missed case when imported item is a variable and it has to be variable and not a value so references can be added to it.


jakebailey:
Ah, yeah, I see the two failing references tests.

PR #1589, opened 2019-09-26
Fix goto definition in from import

MikhailArkhipov:


src/Analysis/Ast/Impl/Analyzer/Handlers/FromImportHandler.cs
```diff
@@ -109,7 +109,7 @@ internal sealed partial class ImportHandler {
             value = value ?? variableModule.GetMember(memberName);
             // Value may be variable or submodule. If it is variable, we need it in order to add reference.
             var variable = variableModule.Analysis?.GlobalScope?.Variables[memberName];
-            value = variable?.Value == value ? variable : value;
+            value = variable?.Value?.Equals(value) == true ? variable : value;
```
CTrando: Don't need to do `== true`, consider adding a test too

MikhailArkhipov: Yes it is nullable so without `== true` it is a syntax error. There is a test and it is broken, that's why.

PR #1593, opened 2019-09-26
Process function call arguments so they get references added

### LINKED ISSUE #1174: All references are not listed when using ms language server

MikhailArkhipov:
Fixes #1174

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Callables.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Callables.cs
@@ -385,6 +385,10 @@ namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
         }
 
         internal void ProcessCallForReferences(CallExpression callExpr) {
+            if (Module.ModuleType != ModuleType.User) {
+                return;
+            }
+
             switch (callExpr.Target) {
                 case NameExpression nex when !string.IsNullOrEmpty(nex.Name):
                     // Add reference to the function
@@ -396,6 +400,7 @@ namespace Microsoft.Python.Analysis.Analyzer.Evaluation {
                     break;
                 }
             }
+
             // Add references to all arguments.
             foreach (var arg in callExpr.Args) {
                 GetValueFromExpression(arg.Expression);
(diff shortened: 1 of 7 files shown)


src/Analysis/Ast/Impl/Analyzer/Evaluation/ExpressionEval.Callables.cs
```diff
@@ -383,5 +383,23 @@ private bool EvaluateFunctionBody(IPythonFunctionType fn)
...
+                    this.LookupNameInScopes(nex.Name)?.AddReference(GetLocationOfName(nex));
+                    break;
+                case MemberExpression mex when !string.IsNullOrEmpty(mex.Name): {
+                    var t = GetValueFromExpression(mex.Target)?.GetPythonType();
+                    t?.GetMember(mex.Name)?.AddReference(GetLocationOfName(mex));
+                    break;
+                }
+            }
+            // Add references to all arguments.
+            foreach (var arg in callExpr.Args) {
```
MikhailArkhipov: Should we limit this to user code?

jakebailey: We already don't store references in library code, right?

MikhailArkhipov: In library code no, only user-to-library so user can find, say, all usages of `print`

jakebailey: If the current file is a library, then I'd think we can skip it. The only way I can see it breaking is library code which references user code, which I can't think of a way to sanely construct.

PR #1596, opened 2019-09-27
some clean ups

heejaechang:
removed unused usings. added named arguments and etc.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
@@ -31,11 +31,11 @@ namespace Microsoft.Python.Analysis.Analyzer {
         public AnalysisModuleKey(IPythonModule module) : this(
             module.Name,
             module.ModuleType == ModuleType.CompiledBuiltin ? null : module.FilePath,
-            module is StubPythonModule && ((StubPythonModule)module).IsTypeshed,
-            (module.IsNonUserFile() || module.IsCompiled()) && module is IDocument && ((IDocument)module).IsOpen) { }
+            IsTypeshedModule(module),
+            IsNonUserAsDocumentModule(module)) { }
 
         public AnalysisModuleKey(string name, string filePath, bool isTypeshed)
-            : this(name, filePath, isTypeshed, isNonUserAsDocument: false) { }
+            : this(name, filePath, isTypeshed, false) { }
 
         private AnalysisModuleKey(string name, string filePath, bool isTypeshed, bool isNonUserAsDocument) {
             Name = name;
@@ -44,7 +44,7 @@ namespace Microsoft.Python.Analysis.Analyzer {
             IsNonUserAsDocument = isNonUserAsDocument;
         }
 
-        public AnalysisModuleKey GetNonUserAsDocumentKey() => new AnalysisModuleKey(Name, FilePath, IsTypeshed, isNonUserAsDocument: true);
+        public AnalysisModuleKey GetNonUserAsDocumentKey() => new AnalysisModuleKey(Name, FilePath, IsTypeshed, true);
 
         public bool Equals(AnalysisModuleKey other)
             => Name.EqualsOrdinal(other.Name) && FilePath.PathEquals(other.FilePath) && IsTypeshed == other.IsTypeshed && IsNonUserAsDocument == other.IsNonUserAsDocument;
@@ -72,5 +72,10 @@ namespace Microsoft.Python.Analysis.Analyzer {
         }
 
         public override string ToString() => $"{Name}({FilePath})";
+
+        private static bool IsTypeshedModule(IPythonModule module)
+            => module is StubPythonModule stubPythonModule && stubPythonModule.IsTypeshed;
+        private static bool IsNonUserAsDocumentModule(IPythonModule module)
+            => (module.IsNonUserFile() || module.IsCompiled()) && module is IDocument document && document.IsOpen;
     }
 }
--- a/src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
@@ -28,15 +28,14 @@ namespace Microsoft.Python.Analysis.Analyzer {
         public bool IsTypeshed { get; }
         public bool IsNonUserAsDocument { get; }
 
-        public AnalysisModuleKey(IPythonModule module) {
-            Name = module.Name;
-            FilePath = module.ModuleType == ModuleType.CompiledBuiltin ? null : module.FilePath;
-            IsTypeshed = module is StubPythonModule stub && stub.IsTypeshed;
-            IsNonUserAsDocument = (module.IsNonUserFile() || module.IsCompiled()) && module is IDocument document && document.IsOpen;
-        }
+        public AnalysisModuleKey(IPythonModule module) : this(
+            module.Name,
+            module.ModuleType == ModuleType.CompiledBuiltin ? null : module.FilePath,
+            module is StubPythonModule && ((StubPythonModule)module).IsTypeshed,
+            (module.IsNonUserFile() || module.IsCompiled()) && module is IDocument && ((IDocument)module).IsOpen) { }
 
-        public AnalysisModuleKey(string name, string filePath, bool isTypeshed) 
-            : this(name, filePath, isTypeshed, false) { }
+        public AnalysisModuleKey(string name, string filePath, bool isTypeshed)
+            : this(name, filePath, isTypeshed, isNonUserAsDocument: false) { }
 
         private AnalysisModuleKey(string name, string filePath, bool isTypeshed, bool isNonUserAsDocument) {
             Name = name;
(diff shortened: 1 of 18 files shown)


src/Analysis/Ast/Impl/Caching/CacheFolders.cs
```diff
@@ -34,7 +34,7 @@ internal sealed class CacheFolderService: ICacheFolderService {
             // File name depends on the content so we can distinguish between different versions.
             using (var hash = SHA256.Create()) {
                 return Convert
-                    .ToBase64String(hash.ComputeHash(new UTF8Encoding(false).GetBytes(content)))
+                    .ToBase64String(hash.ComputeHash(new UTF8Encoding(encoderShouldEmitUTF8Identifier: false).GetBytes(content)))
```
jakebailey: It's not in our convention to require named parameters to include their names. Stuff like this looks kinda silly... The UTF8Encoding docs themselves don't do this.

src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
```diff
@@ -28,15 +28,14 @@ namespace Microsoft.Python.Analysis.Analyzer {
...
-        public AnalysisModuleKey(IPythonModule module) {
-            Name = module.Name;
-            FilePath = module.ModuleType == ModuleType.CompiledBuiltin ? null : module.FilePath;
-            IsTypeshed = module is StubPythonModule stub && stub.IsTypeshed;
-            IsNonUserAsDocument = (module.IsNonUserFile() || module.IsCompiled()) && module is IDocument document && document.IsOpen;
-        }
+        public AnalysisModuleKey(IPythonModule module) : this(
+            module.Name,
+            module.ModuleType == ModuleType.CompiledBuiltin ? null : module.FilePath,
+            module is StubPythonModule && ((StubPythonModule)module).IsTypeshed,
```
MikhailArkhipov: `IsTypeshed = module is StubPythonModule stub && stub.IsTypeshed;` is better IMO

src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
```diff
@@ -28,15 +28,14 @@ namespace Microsoft.Python.Analysis.Analyzer {
...
+            (module.IsNonUserFile() || module.IsCompiled()) && module is IDocument && ((IDocument)module).IsOpen) { }
```
MikhailArkhipov: As above

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
heejaechang: sure. I was looking code and when I saw "false", I wasn't sure what "false" do with UTF8Encoding. so added the named argument. if people prefere not having named argument. I am fine.

src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
MikhailArkhipov: You can check other files in the project, we prefer `module is IDocument document && document.IsOpen` rather than casts

src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
heejaechang: it can't be used in this( ... ) context in language version it is using. looks like it is supported in never language version if people are okay to move to latest lang version I can make that change.

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
jakebailey: Yeah, this one's unfortunate. There really should have been a static member of that class which encoded this behavior rather than this weird `false` everywhere to get actual UTF-8 behavior.

MikhailArkhipov: Yeah, if anyone wants them, there are VS editor extensions that display them from signatures as editor adornments.

src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
heejaechang: I changed ctor so it can share initialization otherwise when a new field is added, one needs to fix both ctor.

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
MikhailArkhipov: They make sense sometimes, typically for booleans when it is unclear what true/false might mean.

src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
heejaechang: ya. I prefer that as well. let me do some trick so that I can do that instead of putting code directly in this(..) so I can use that construct.

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
heejaechang: so remove named arguments then?

heejaechang: by the way, caching SHA256 and using incremental hash will improve perf quite a bit if it is used a lot. 

Roslyn started same as what it has now but changed to this - http://source.roslyn.io/#Microsoft.CodeAnalysis.Workspaces/Workspace/Solution/Checksum_Factory.cs,21

made about 3-4x improvement on hashing perf. Roslyn creates SHA for everything in solution, (for Roslyn solution itself more than 40000 SHA), and we saw big improvement on total time we spend on SHA.

probably only matter only if we also does many SHA on big stream. otherwise, probably won't matter much.

src/Analysis/Core/Impl/Interpreter/InterpreterConfiguration.cs
```diff
@@ -22,7 +22,7 @@ public sealed class InterpreterConfiguration : IEquatable<InterpreterConfigurati
         /// Constructs a new interpreter configuration based on the provided values.
         /// </summary>
         public InterpreterConfiguration(string interpreterPath = null, Version version = null) :
-            this(interpreterPath, string.Empty, null, null, default, version) { }
+            this(interpreterPath, pathVar: string.Empty, libPath: null, sitePackagesPath: null, architecture: default, version: version) { }
```
heejaechang: I am deleting named argument, but there is one more benefit to named arguments. when a new parameter is added to a method, it reduces confusion whether right value is assigned to right parameter when literal constant is used as an argument.

src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
MikhailArkhipov: We are on .NET Core 3.0 which means C# 8

heejaechang: but project is explicitly set to use langversion 7.2

```
    <TargetFramework>netstandard2.0</TargetFramework>
    <RootNamespace>Microsoft.Python.Analysis</RootNamespace>
    <AssemblyName>Microsoft.Python.Analysis</AssemblyName>
  </PropertyGroup>
  <PropertyGroup>
    <!--
      1701, 1702 - "You may need to supply assembly policy"
    -->
    <NoWarn>1701;1702;$(NoWarn)</NoWarn>
    <LangVersion>7.2</LangVersion>
```

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
MikhailArkhipov: I am OK with named args for booleans when it is not obvious what it is. Leave it.

src/Analysis/Ast/Impl/Analyzer/AnalysisModuleKey.cs
heejaechang: I can move to 8.0 or latest if team prefer that?

jakebailey: Can we do both .NET standard 2.0 and language version 8?

heejaechang: I believe we can. Roslyn already do C# 8.0 but still target .net standard 2.0

heejaechang: ?

MikhailArkhipov: I mean we are using SDK 3.0 and that automatically mean C# 8

MikhailArkhipov: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version

MikhailArkhipov: We can remove all `LangVersion` now

heejaechang: I will send out new PR on removing all lang version

PR #1597, opened 2019-09-27
added some comments and renamed some variables in PathResolver and PathResolverSnapshot

heejaechang:
this sit on top of https://github.com/microsoft/python-language-server/pull/1596

this is actual change - https://github.com/microsoft/python-language-server/pull/1597/commits/d6c86113d25faa4a7870af182899ac4b2542c0ff - and commit before it is same as #1596

### DIFF
--- a/src/Analysis/Core/Impl/DependencyResolution/PathResolver.cs
+++ b/src/Analysis/Core/Impl/DependencyResolution/PathResolver.cs
@@ -20,6 +20,9 @@ using Microsoft.Python.Parsing;
 namespace Microsoft.Python.Analysis.Core.DependencyResolution {
     public sealed class PathResolver {
         public PathResolver(in PythonLanguageVersion pythonLanguageVersion, in string root, in ImmutableArray<string> interpreterSearchPaths, in ImmutableArray<string> userSearchPaths) {
+            // need to audit whether CurrentSnapshot ever get updated concurrently
+            // thread safety is probably fine but serialization is another issue where if 2 callers do TryAddModulePath or other calls concurrently, 
+            // they will use same CurrentSnapshot to update which ends up with 2 different snapshots with same version.
             CurrentSnapshot = new PathResolverSnapshot(pythonLanguageVersion, root, interpreterSearchPaths, userSearchPaths);
         }
 
--- a/src/Analysis/Core/Impl/DependencyResolution/PathResolverSnapshot.Edge.cs
+++ b/src/Analysis/Core/Impl/DependencyResolution/PathResolverSnapshot.Edge.cs
@@ -35,55 +35,69 @@ namespace Microsoft.Python.Analysis.Core.DependencyResolution {
             //  ┌─╚═╗
             // [x] [y] 
             //
-            // will be stored as list of pairs: (2, a), (0, i), (1, y)
+            // will be stored as list of pairs: (2, c), (0, i), (1, y)
             // 
             // To provide immutability, it must be changed only by calling List<T>.Add
-            private readonly ImmutableArray<(int index, Node node)> _vertices;
-            private readonly int _index;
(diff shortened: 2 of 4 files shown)


src/Analysis/Ast/Impl/Analyzer/ProgressReporter.cs
```diff
@@ -20,8 +20,8 @@
 
 namespace Microsoft.Python.Analysis.Analyzer {
     internal sealed class ProgressReporter : IProgressReporter, IDisposable {
-        private const int _initialDelay = 50;
-        private const int _reportingInterval = 100;
+        private readonly static TimeSpan s_initialDelay = TimeSpan.FromMilliseconds(50);
```
MikhailArkhipov: We don't use Hungarian prefixes

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
```diff
@@ -93,7 +94,7 @@ internal sealed class PythonAnalyzerEntry {
         }
 
         public Task<IDocumentAnalysis> GetAnalysisAsync(CancellationToken cancellationToken)
-            => _analysisTcs.Task.ContinueWith(t => t.GetAwaiter().GetResult(), cancellationToken);
+            => _analysisTcs.Task.WaitAsync(cancellationToken);
```
MikhailArkhipov: @AlexanderSher ?

src/Analysis/Ast/Impl/Analyzer/ProgressReporter.cs
heejaechang: so, just _initialDelay?

jakebailey: In our convention, it'd be `InitialDelay` as it is a static value.

jakebailey: FWIW I think this change is from #1596, so should be commented there.

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzerEntry.cs
jakebailey: Also from #1596, so we should probably review that first.

src/LanguageServer/Impl/LanguageServer.Configuration.cs
```diff
@@ -73,23 +73,23 @@ public sealed partial class LanguageServer {
...
+            settings.diagnosticPublishDelay = GetSetting(analysis, "diagnosticPublishDelay", defaultValue: 1000);
             var ds = _services.GetService<IDiagnosticsService>();
             ds.PublishingDelay = settings.diagnosticPublishDelay;
 
             ds.DiagnosticsSeverityMap = new DiagnosticsSeverityMap(
-                GetSetting(analysis, "errors", Array.Empty<string>()),
-                GetSetting(analysis, "warnings", Array.Empty<string>()),
-                GetSetting(analysis, "information", Array.Empty<string>()),
-                GetSetting(analysis, "disabled", Array.Empty<string>()));
+                GetSetting(analysis, "errors", defaultValue: Array.Empty<string>()),
```
MikhailArkhipov: Not entirely convinced in need to specify defaults explicitly, makes it longer. There are VS extensions that display then if needed.

heejaechang: it makes reviewing code easier though :) especially when constant is used. but not a strong argument. but if team prefer no named argument, I can go and delete those.

MikhailArkhipov:
https://github.com/microsoft/python-language-server/pull/1584 may need to go first

heejaechang:
@MikhailArkhipov ping?

MikhailArkhipov:
Not before #1584 since it touches same area.

PR #1598, opened 2019-09-27
some random refactoring around code base

heejaechang:
this sit on top of https://github.com/microsoft/python-language-server/pull/1597

actual change - https://github.com/microsoft/python-language-server/pull/1598/commits/129dbb677f78232fa5527b279420bcf0aff9d136

### DIFF
--- a/src/Analysis/Ast/Impl/Documents/RunningDocumentTable.cs
+++ b/src/Analysis/Ast/Impl/Documents/RunningDocumentTable.cs
@@ -40,7 +40,7 @@ namespace Microsoft.Python.Analysis.Documents {
         private IModuleManagement _moduleManagement;
         private IModuleManagement ModuleManagement => _moduleManagement ?? (_moduleManagement = _services.GetService<IPythonInterpreter>().ModuleResolution);
 
-        private class DocumentEntry {
+        private sealed class DocumentEntry {
             public readonly IDocument Document;
             public int LockCount;
 
@@ -89,7 +89,7 @@ namespace Microsoft.Python.Analysis.Documents {
             lock (_lock) {
                 var entry = FindDocument(uri);
                 if (entry == null) {
-                    var resolver = _services.GetService<IPythonInterpreter>().ModuleResolution.CurrentPathResolver;
+                    var resolver = ModuleManagement.CurrentPathResolver;
 
                     var moduleType = ModuleType.User;
                     var path = uri.ToAbsolutePath();
@@ -212,6 +212,7 @@ namespace Microsoft.Python.Analysis.Documents {
             }
 
             foreach (var (_, entry) in closed) {
+                // closed and removed event always raise events with already disposed documents.
                 Closed?.Invoke(this, new DocumentEventArgs(entry.Document));
                 Removed?.Invoke(this, new DocumentEventArgs(entry.Document));
             }
@@ -223,7 +224,8 @@ namespace Microsoft.Python.Analysis.Documents {
 
         public void Dispose() {
             lock (_lock) {
-                foreach (var d in _documentsByUri.Values.OfType<IDisposable>()) {
+                // DocumentEntry is not disposable. it is Document that is disposable.
+                foreach (var d in _documentsByUri.Values.Select(v => v.Document)) {
                     d.Dispose();
                 }
             }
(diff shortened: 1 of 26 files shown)


src/Analysis/Ast/Impl/Documents/RunningDocumentTable.cs
```diff
@@ -223,7 +224,8 @@ private class DocumentEntry {
 
         public void Dispose() {
             lock (_lock) {
-                foreach (var d in _documentsByUri.Values.OfType<IDisposable>()) {
+                // DocumentEntry is not disposable. it is Document that is disposable.
+                foreach (var d in _documentsByUri.Values.Select(v => v.Document)) {
```
heejaechang: existing code didn't do anything since Value is DocumentEntry and it doesn't implement Disposable. I can either remove code here or make it to dispose Document.

src/LanguageServer/Impl/Diagnostics/DiagnosticsService.cs
```diff
@@ -36,11 +36,18 @@ private sealed class DocumentDiagnostics {
             public bool Changed { get; set; }
 
             public void Clear(DiagnosticSource source) {
-                Changed = Entries.Count > 0;
+                if (!_entries.TryGetValue(source, out _)) {
```
heejaechang: if there is no entry in _entries, I believe nothing could considered changed. but if we want to make Changed always true, then I can revert it back to original code.

src/LanguageServer/Impl/Diagnostics/DiagnosticsService.cs
```diff
@@ -36,11 +36,18 @@ private sealed class DocumentDiagnostics {
...
+                    return;
+                }
+
+                Changed = true;
                 _entries[source] = Array.Empty<DiagnosticsEntry>();
             }
 
-            public void ClearAll() => _entries.Clear();
+            public void ClearAll() {
+                Changed = true;
```
heejaechang: this one seems a bug. it cleared all code but never but it as Changed. so published diagnostic probably won't get ever cleaned up?

src/LanguageServer/Impl/Diagnostics/DiagnosticsService.cs
```diff
@@ -64,7 +71,12 @@ private sealed class DocumentDiagnostics {
         private IRunningDocumentTable _rdt;
         private DateTime _lastChangeTime;
 
-        private IRunningDocumentTable Rdt => _rdt ?? (_rdt = _services.GetService<IRunningDocumentTable>());
+        private IRunningDocumentTable Rdt {
+            get {
+                ConnectToRdt();
```
heejaechang: call ConntectToRdt to make sure event are always get registered. and _rdt are set under lock.

src/LanguageServer/Impl/Diagnostics/DiagnosticsService.cs
```diff
@@ -126,9 +138,11 @@ private sealed class DocumentDiagnostics {
                     _severityMap = value;
                     foreach (var d in _diagnostics) {
                         d.Value.Changed = true;
-                        _lastChangeTime = DateTime.Now;
                     }
+
                     PublishDiagnostics();
+
+                    _lastChangeTime = DateTime.Now;
```
heejaechang: _lastChangeTime is read onIdle. all write is under lock, but on idle read it without lock. DateTime''s only field is UInt64 so probably okay for 64bit machine, but in 32bit machine might not thread safe? do we care?

src/LanguageServer/Impl/LanguageServer.Lifetime.cs
```diff
@@ -24,55 +24,60 @@
...
+
         private InitializeParams _initParams;
         private bool _shutdown;
 
         private Task<IDisposable> _initializedPriorityTask;
 
         [JsonRpcMethod("initialize")]
         public async Task<InitializeResult> Initialize(JToken token, CancellationToken cancellationToken) {
-            _initParams = token.ToObject<InitializeParams>();
+            _initParams = ToObject<InitializeParams>(token);
```
heejaechang: this one was only one that doesn't use ToObject method to deserialize token.

MikhailArkhipov: Yeah, its a good catch.

src/Analysis/Ast/Impl/Documents/RunningDocumentTable.cs
MikhailArkhipov: It selects document from the value, no?

heejaechang: now d is IDocument. before it was DocumentEntry which doesn't implement IDisposable

MikhailArkhipov:
Needs master merge

PR #1604, opened 2019-09-30
Make sure we pass lookup options to expression evaluation

### LINKED ISSUE #1600: Go to definition not working
image

## Environment data

- Extension Version: (ms-python.python) 2019.9.34911   
- VScode Version: 1.38.1
- Python version: 2.7.16 64-bit
- OS: Darwin x64 18.0.0

## Expected behaviour

Go to definition while using  Nested Classes

## Actual behaviour

Cant go to definition

## Logs

```
```

```python
class test():
    class Test2():
        def Z(self):
            return
    A = Test2()

    def X(self):
        return


test.A.Z()

```

MikhailArkhipov:
Fixes #1600 

We must make sure lookup options for name resolution are passed since additional `ClassMembers` is passed in the class scope context so we can find class members without `self.`.


jakebailey:
Fails `AmbiguousOptionalParameterType`, since the lookup options affect which `A` is found for the type hint.

MikhailArkhipov:
0.4.38+

PR #1614, opened 2019-10-01
Add positional-only parameters to parser, analysis, LS, etc

### LINKED ISSUE #1511: Support positional-only parameters

jakebailey:
Fixes #1511.

Still needs some testing in the analysis past ArgumentSet specific ones, but it's nearly there.

### DIFF
--- a/src/Analysis/Ast/Impl/Types/ParameterInfo.cs
+++ b/src/Analysis/Ast/Impl/Types/ParameterInfo.cs
@@ -35,9 +35,7 @@ namespace Microsoft.Python.Analysis.Types {
             DefaultValue = defaultValue;
             Type = type;
             Kind = kind ?? ParameterKind.Normal;
-#if DEBUG
             Debug.Assert(Kind != ParameterKind.PositionalOnlyMarker);
-#endif
         }
 
         public string Name { get; }
--- a/src/Analysis/Ast/Impl/Types/ParameterInfo.cs
+++ b/src/Analysis/Ast/Impl/Types/ParameterInfo.cs
@@ -17,6 +17,7 @@ using System.Diagnostics;
 using Microsoft.Python.Parsing.Ast;
 
 namespace Microsoft.Python.Analysis.Types {
+    [DebuggerDisplay("{Name} : {Kind}")]
     internal sealed class ParameterInfo : IParameterInfo {
         public ParameterInfo(PythonAst ast, Parameter p, IPythonType type, IMember defaultValue, bool isGeneric)
             : this(p?.Name, type, p?.Kind, defaultValue) {
--- a/src/Analysis/Ast/Impl/Types/ParameterInfo.cs
+++ b/src/Analysis/Ast/Impl/Types/ParameterInfo.cs
@@ -35,7 +35,7 @@ namespace Microsoft.Python.Analysis.Types {
             Type = type;
             Kind = kind ?? ParameterKind.Normal;
 #if DEBUG
-            Debug.Assert(Kind != ParameterKind.PositionalMarker);
+            Debug.Assert(Kind != ParameterKind.PositionalOnlyMarker);
 #endif
         }
 
--- a/src/Analysis/Ast/Impl/Types/ParameterInfo.cs
+++ b/src/Analysis/Ast/Impl/Types/ParameterInfo.cs
@@ -13,7 +13,7 @@
 // See the Apache Version 2.0 License for specific language governing
 // permissions and limitations under the License.
 
-using System;
+using System.Diagnostics;
 using Microsoft.Python.Parsing.Ast;
 
 namespace Microsoft.Python.Analysis.Types {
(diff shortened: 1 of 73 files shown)


jakebailey:
Ready for a look, I think.

src/Analysis/Ast/Impl/Types/ParameterInfo.cs
```diff
@@ -34,6 +35,9 @@ public ParameterInfo(PythonAst ast, Parameter p, IPythonType type, IMember defau
             DefaultValue = defaultValue;
             Type = type;
             Kind = kind ?? ParameterKind.Normal;
+#if DEBUG
+            Debug.Assert(Kind != ParameterKind.PositionalOnlyMarker);
```
MikhailArkhipov: Don't have to `#if` it, AFAIK `Debug.` is noop in release.

src/LanguageServer/Impl/Sources/DocumentationSource.cs
```diff
@@ -13,57 +13,83 @@
...
+                } else {
+                    addComma = true;
+                }
+
+                // Positional only markers are not included in the parameters, so keep track
+                // of whether or not positional only parameters have been seen then add the marker
+                // once the positional only section ends.
+                if (p.Kind == ParameterKind.PositionalOnly) {
+                    addMarker = true;
+                } else if (addMarker && p.Kind != ParameterKind.PositionalOnly) {
```
MikhailArkhipov: Do we care about `def func(a, b, /)` ?

jakebailey: What do you mean by "care"? It's a valid case and needs to be handled, yes. I do that below by checking to see if `addMarker` was left true after the loop.

src/Analysis/Ast/Impl/Types/ParameterInfo.cs
jakebailey: Yeah, I was unsure if its arguments would be evaluated, but after some searching it looks like the compiler will remove all of the code, including anything that'd normally be run to do the arguments. I just didn't want this traffic-heavy area to do this comparison all the time.

src/Parsing/Impl/Parser.cs
```diff
@@ -2102,7 +2127,7 @@ struct Name {
                 }
 
                 if (string.IsNullOrEmpty(p.Name)) {
-                    if (p.Kind != ParameterKind.List || !(_stubFile || _langVersion.Is3x())) {
+                    if (!(p.Kind == ParameterKind.List || p.Kind == ParameterKind.PositionalOnlyMarker) || !(_stubFile || _langVersion.Is3x())) {
                         ReportSyntaxError(p.StartIndex, p.EndIndex, Resources.InvalidSyntaxErrorMsg);//Invalid Syntax
```
MikhailArkhipov: `(p.Kind == ParameterKind.List && p.Kind == ParameterKind.PositionalOnlyMarker) || (_stubFile  && _langVersion.Is3x())` ?

jakebailey: `(p.Kind != ParameterKind.List && p.Kind != ParameterKind.PositionalOnlyMarker) || (_stubFile && _langVersion.Is3x())` I think, yeah.

src/Parsing/Impl/Parser.cs
```diff
@@ -2120,6 +2145,19 @@ struct Name {
                         ReportSyntaxError(p.StartIndex, p.EndIndex, Resources.DuplicateArgsDoubleArgumentErrorMsg);//duplicate ** args arguments
                     }
                     seenDictArg = true;
+                } else if (p.Kind == ParameterKind.PositionalOnlyMarker) {
+                    if (seenPositional) {
```
MikhailArkhipov: Do we want to show all errors, i.e. if it has several issues like annotation, default and something else?

jakebailey: I think the convention in the parser is to try and only show one error at a time per-character (if that makes sense), such that any given place will only be squiggled the one time.

src/LanguageServer/Impl/Sources/DocumentationSource.cs
MikhailArkhipov: OK

src/Parsing/Impl/Parser.cs
MikhailArkhipov: VS squiggles multiple so user can fix all errors rather than fix one, then see another. Technically squiggles should be different, i.e. annotation under `:`, default under `=`. Up to you.

jakebailey: I'd prefer to stick with this for now, and come back to improve the errors overall if they overlap (since we have a number of cases like this already).

PR #1625, opened 2019-10-04
Fix analysis module key duplication is stubs

MikhailArkhipov:
- Fix analysis module key duplication is stubs: IsTypeshed is set in `StubPythonModule` after base initiated loading of content and analysis so occasionally there may be two keys in the analyzer - one with IsTypeshed set and one without.
- Fix builtins module resolution in typeshed - so it resolves to the same common builtins and not to stub

### DIFF
--- a/src/Analysis/Ast/Impl/Modules/Resolution/ModuleResolutionBase.cs
+++ b/src/Analysis/Ast/Impl/Modules/Resolution/ModuleResolutionBase.cs
@@ -63,7 +63,7 @@ namespace Microsoft.Python.Analysis.Modules.Resolution {
         protected abstract IPythonModule CreateModule(string name);
 
         public IPythonModule GetImportedModule(string name) {
-            if (name == "__builtin__") {
+            if (name == Interpreter.ModuleResolution.BuiltinsModule.Name) {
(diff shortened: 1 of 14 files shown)


src/Analysis/Ast/Impl/Modules/Resolution/ModuleResolutionBase.cs
```diff
@@ -62,8 +62,14 @@ internal abstract class ModuleResolutionBase {
 
         protected abstract IPythonModule CreateModule(string name);
 
-        public IPythonModule GetImportedModule(string name)
-            => Modules.TryGetValue(name, out var moduleRef) ? moduleRef.Value : Interpreter.ModuleResolution.GetSpecializedModule(name);
+        public IPythonModule GetImportedModule(string name) {
+            if (name == "__builtin__") {
```
jakebailey: How would `BuiltinModuleName` come into play here? Should it be used instead? I think the builtin module's name changes between 2 and 3, see `BuiltinTypeIdExtensions.GetModuleName`

MikhailArkhipov: Good point

PR #1631, opened 2019-10-04
Remove NoneType in favor of a consistent None, and remove completions which are also keywords

### LINKED ISSUE #1628: None shows twice in completion as None and __NoneType__
image

Fallout of #1624. The `__NoneType__` thing is coming from the scraper.

jakebailey:
Fixes #1628.


MikhailArkhipov:
Looks like we need to bump folder version to `stubs.v3` since otherwiase old scrape of builtins will still be there

MikhailArkhipov:
Approving pending stub version bump.

jakebailey:
Ah, yeah. Should I be bumping the DB version too, and drop the old reference to NoneType completely?

MikhailArkhipov:
Models technically didn't change so either we need to test with system caching on (i.e. cache with old code, then flip to new) or just bump up for safety.

PR #1639, opened 2019-10-07
Handle Unions of collections in ValueEnumerator, recursively

### LINKED ISSUE #1622: _ExcInfo and _OptExcInfo broken with new typeshed definition
Typeshed defines in `sys.pyi`:

```python
_ExcInfo = Tuple[Type[BaseException], BaseException, TracebackType]]
_OptExcInfo = Union[_ExcInfo, Tuple[None, None, None]]

def exc_info() -> _OptExcInfo: ...
```

With the `TypeShedSysExcInfo` test which checks the return value of `exc_info`, this fails with:

```
"BaseException" with a length of 13, but 
    "Union[Tuple[Type[BaseException], BaseException, TracebackType], Tuple[Unknown]]" has a length of 79, differs near "Uni" (index 0).
```

As it's unpacking the first item incorrectly, likely because at the top level it's a union.

Copy that same code out into its own typing test:

```python
_ExcInfo = Tuple[Type[BaseException], BaseException, TracebackType]]
_OptExcInfo = Union[_ExcInfo, Tuple[None, None, None]]

x: _ExcInfo
y: _OptExcInfo

a, b, c = u
```

And now it complains about `x`:

```
 Expected Type.Name to be 
    "Tuple[Type[BaseException], BaseException, TracebackType]" with a length of 56, but 
    "Tuple[BaseException, Unknown]" has a length of 29, differs near "Bas" (index 6).
```

So the tuple type parameters aren't right either.

Needed for #1371.

jakebailey:
Fixes #1622.

If we're trying to unpack a union, store a ValueEnumerator per union item, then produce unions as items from each of the nested enumerators.

Note that since the unions are of types and are not values, we need to do some extra work to shift back and forth between types and instances where needed.

Also:

- Replace the `Next` property with a method, since it has side effects.
- Replace a comment I added in a previous PR with a new one explaining what's _actually_ happening.
- Use ImmutableArray in ValueEnumerator (the default works better).

### DIFF
--- a/src/Analysis/Ast/Test/TypingTests.cs
+++ b/src/Analysis/Ast/Test/TypingTests.cs
@@ -343,6 +343,39 @@ x: Tuple[None, None, None]
         }
 
         [TestMethod, Priority(0)]
+        public async Task UnionOfTuples() {
+            const string code = @"
+from typing import Union, Tuple, Type
+
+class TracebackType: ...
+
+_ExcInfo = Tuple[Type[BaseException], BaseException, TracebackType]
+_OptExcInfo = Union[_ExcInfo, Tuple[None, None, None]]
+
+x: _ExcInfo
+y: _OptExcInfo
+
+a, b, c = y
+";
+            var analysis = await GetAnalysisAsync(code, PythonVersions.LatestAvailable3X);
+
+            analysis.Should().HaveVariable("x")
+                .Which.Should().HaveType("Tuple[Type[BaseException], BaseException, TracebackType]");
+
+            analysis.Should().HaveVariable("y")
+                .Which.Should().HaveType("Union[Tuple[Type[BaseException], BaseException, TracebackType], Tuple[None, None, None]]");
+
+            analysis.Should().HaveVariable("a")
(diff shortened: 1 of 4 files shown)


src/Analysis/Ast/Test/TypingTests.cs
```diff
@@ -342,6 +342,39 @@ class A:
...
+                .Which.Should().HaveType("Union[Tuple[Type[BaseException], BaseException, TracebackType], Tuple[None, None, None]]");
+
+            analysis.Should().HaveVariable("a")
+                .Which.Should().HaveType("Union[Type[BaseException], None]");
+
+            analysis.Should().HaveVariable("b")
+                .Which.Should().HaveType("Union[BaseException, None]");
+
+            analysis.Should().HaveVariable("c")
+                .Which.Should().HaveType("Union[TracebackType, None]");
```
jakebailey: It'd be nice some day to convert `Union[X, None]` to `Optional[X]`, since that's the more readable choice, but this is correct for now.

PR #1649, opened 2019-10-08
Fix issue when diagnostics not removed on document close

MikhailArkhipov:


src/LanguageServer/Impl/Diagnostics/DiagnosticsService.cs
```diff
@@ -243,7 +237,8 @@ private IEnumerable<DiagnosticsEntry> FilterBySeverityMap(DocumentDiagnostics d)
 
         private void ClearDiagnostics(Uri uri, bool remove) {
             lock (_lock) {
-                if (_diagnostics.TryGetValue(uri, out var _)) {
+                if (_diagnostics.TryGetValue(uri, out var d)) {
+                    d.ClearAll();
```
jakebailey: This function confuses me. What is the `remove` parameter supposed to do compared to clearing?

MikhailArkhipov: Clear means remove diagnostics, but the document is still open, such as when new parse begins. Remove means the document is closed and its diagnostics set should be gone.

PR #1656, opened 2019-10-09
implement AddImport

### LINKED ISSUE #19: Implement code action for missing imports
https://github.com/Microsoft/vscode-python/issues/21

Identify undeclared variables, find what module might have them, offer to insert import statement.

heejaechang:
Closes #19.

provides add import code action for unknown symbols.

...

- [x] basic implementation for importing "modules" including submodules + members from loaded modules
- [X] add tests
- [X] add add import option [decide not to add option for now]
- [X] add support for members from unloaded modules
- [X] improve add import code action ordering a bit
- [X] open related issue on a relative path (https://github.com/microsoft/python-language-server/issues/1707)
- [X] provide abbreviation for well-known modules.
- [x] handle context. ex) specific import based on context (only import type/function and etc)
- [X] perf ex) cache lint result. run lint only for specific diagnostics and etc - now uses diagnostic service cache
- [X] show quick fix in hover
- [X] support reversed abbreviation for well-known modules
- [X] support unique name on abbreviation

addimport2


src/Analysis/Ast/Impl/Modules/Resolution/ModuleResolutionBase.cs
```diff
@@ -74,11 +74,6 @@ public IPythonModule GetImportedModule(string name)
                 return module;
             }
 
-            module = Interpreter.ModuleResolution.GetSpecializedModule(name);
```
heejaechang: one line above, GetImportedModule, already did exactly same thing

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
```diff
@@ -55,6 +55,18 @@ public MainModuleResolution(string root, IServiceContainer services, ImmutableAr
 
         public IBuiltinsPythonModule BuiltinsModule { get; private set; }
 
+        public IEnumerable<IPythonModule> GetImportedModules(CancellationToken cancellationToken) {
+            foreach (var moduleRef in Modules.Values) {
+                if (moduleRef.Value != null) {
```
MikhailArkhipov: `cancellationToken.Throw` I guess although I don't think this is a long operation.

src/Analysis/Core/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -82,14 +82,22 @@ public PathResolverSnapshot(PythonLanguageVersion pythonLanguageVersion, string
...
 
         public ImmutableArray<string> GetAllImportableModuleNames(bool includeImplicitPackages = true) {
+            return GetAllImportableModuleFullName(n => !string.IsNullOrEmpty(n.FullModuleName), includeImplicitPackages);
+        }
+
+        public ImmutableArray<string> GetAllImportableModuleByName(string name, bool includeImplicitPackages = true) {
+            return GetAllImportableModuleFullName(n => string.Equals(n.Name, name), includeImplicitPackages);
+        }
+
+        private ImmutableArray<string> GetAllImportableModuleFullName(Func<Node, bool> predicate, bool includeImplicitPackages = true) {
```
MikhailArkhipov: `FullNames` I guess

src/Core/Impl/Text/Position.cs
```diff
@@ -39,7 +39,20 @@ public struct Position {
 
         public static bool operator >(Position p1, Position p2) => p1.line > p2.line || p1.line == p2.line && p1.character > p2.character;
         public static bool operator <(Position p1, Position p2) => p1.line < p2.line || p1.line == p2.line && p1.character < p2.character;
+        public static bool operator ==(Position p1, Position p2) => p1.Equals(p2);
+        public static bool operator !=(Position p1, Position p2) => !p1.Equals(p2);
 
+        public bool Equals(Position other) => line == other.line && character == other.character;
+
+        public override bool Equals(object obj) {
```
MikhailArkhipov: Usually `?` for short method like => obj is Position other ? Equals(other) : false;

src/LanguageServer/Impl/Implementation/Server.Editor.cs
```diff
@@ -112,5 +112,23 @@ await InvokeExtensionsAsync(async (ext, token)
...
+
+        public async Task<CodeAction[]> CodeAction(CodeActionParams @params, CancellationToken cancellationToken) {
+            var uri = @params.textDocument.uri;
+            _log?.Log(TraceEventType.Verbose, $"Code Action in {uri} at {@params.range}");
+
+            if (@params.context.diagnostics?.Length == 0) {
+                return Array.Empty<CodeAction>();
+            }
+
+            // * NOTE * CodeActionParams.CodeActionContext._version is always null. PublishDiagnostic doesn't send version
```
MikhailArkhipov: Safe to remove `_version`. We've been cleaning up old unused custom fields that PTVS used to exchange with old LS.

src/LanguageServer/Impl/Sources/CodeActionSource.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            var results = new List<CodeAction>();
+
+            // * NOTE * this always re-calculate whole document for all linters.
+            //          consider creating linter result cache somewhere (MRU?) and 
+            //          add error code as optional argument to run only that linter rather
+            //          than always running all linters.
+            //          also, current LintModule implementation always run all linters
+            //          even if the linter option is off to set references in the given
+            //          modules. this code path should use different code path to prevent that.
```
MikhailArkhipov: Yes, running all is currently by design as it is part of the analysis and we don't want to re-run that when option changes. So user options are basically display options. We try and separate linter pass but some 'linting' messages are actually produced by the analysis run.

heejaechang: I meant running Linter for code action.

src/LanguageServer/Impl/Sources/CodeActionSource.cs
```diff
@@ -0,0 +1,81 @@
...
+                    }
+                }
+            }
+
+            return results.ToArray();
+        }
+
+        private IEnumerable<DiagnosticsEntry> GetMatchingDiagnostics(IDocumentAnalysis analysis, Diagnostic[] diagnostics, CancellationToken cancellationToken) {
+            var analyzer = _services.GetService<IPythonAnalyzer>();
+            foreach (var diagnostic in analysis.Diagnostics.Concat(analyzer.LintModule(analysis.Document))) {
```
MikhailArkhipov: Analysis runs linter, do we need to re-run it here? All messages should be in the diagnostics

heejaechang: we don't know whether the diagnostic is staled or not. running 1 particular linter seems not too bad?

MikhailArkhipov: The analysis you use for linting is either current - and then the linting is current, or stale - then running linting on a stale analysis does not make sense.

src/LanguageServer/Impl/Sources/CodeActionSource.cs
MikhailArkhipov: You don't need to run it. It was ran during analysis that you are using and any issues are in the `IDiagnosticService`.

src/LanguageServer/Impl/Sources/CodeActionSource.cs
heejaechang: thank you for the explanation. changed it to use IDiagnosticService.

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -191,11 +192,11 @@ public sealed partial class Server : IDisposable {
             _disposableBag.ThrowIfDisposed();
             switch (@params.settings) {
                 case ServerSettings settings: {
-                        Settings = settings;
-                        _symbolHierarchyMaxSymbols = Settings.analysis.symbolsHierarchyMaxSymbols;
-                        _completionSource.Options = Settings.completion;
-                        break;
-                    }
+                    Settings = settings;
```
jakebailey: Style nit, but these `{}` on switch case blocks should just be removed. They don't seem to add anything except noise in diffs because everyone's editors seem to disagree on the indention.

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
```diff
@@ -0,0 +1,635 @@
...
+        private const int ModuleLoadTimeout = 10 * 1000; // 10 seconds
+
+        public static readonly ICodeActionProvider Instance = new MissingImportCodeActionProvider();
+
+        // right now, it is a static. in future, we might consider giving an option to users to customize this list
+        // also, right now, it is text based. so if module has same name, they will get same suggestion even if
+        // the module is not something user expected
+        private static readonly Dictionary<string, string> WellKnownAbbreviationMap = new Dictionary<string, string>() {
+            { "numpy", "np" },
+            { "pandas", "pd" }
```
MikhailArkhipov: I think `tensorflow` is `tf`, `matplotlib.pyplot as plt`, `matplotlib as mpl`, `math as m`, `scipy.io as spio`, `scipy as sp` (https://docs.scipy.org/doc/scipy/reference/api.html, https://stackoverflow.com/questions/36014733/official-abbreviation-for-import-scipy-as-sp-sc)

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
```diff
@@ -0,0 +1,635 @@
...
+            var finder = new ExpressionFinder(analysis.Ast, FindExpressionOptions.Complete);
+            var node = finder.GetExpression(diagnostic.SourceSpan);
+            if (!(node is NameExpression)) {
+                return Enumerable.Empty<CodeAction>();
+            }
+
+            var interpreter = analysis.Document.Interpreter;
+            var pathResolver = interpreter.ModuleResolution.CurrentPathResolver;
+
+            var name = node.ToCodeString(analysis.Ast);
```
MikhailArkhipov: `node is NameExpression nex` (above), then nex.Name. `ToCodeString` is buggy and only used in tests.

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
```diff
@@ -0,0 +1,635 @@
...
+
+            var importFullNameMap = new Dictionary<string, ImportInfo>();
+
+            // find installed modules matching the given name. this will include submodules
+            var languageVersion = Parsing.PythonLanguageVersionExtensions.ToVersion(interpreter.LanguageVersion);
+            var includeImplicit = !ModulePath.PythonVersionRequiresInitPyFiles(languageVersion);
+            foreach (var moduleFullName in pathResolver.GetAllImportableModulesByName(name, includeImplicit)) {
+                cancellationToken.ThrowIfCancellationRequested();
+
+                var module = interpreter.ModuleResolution.GetOrLoadModule(moduleFullName);
```
MikhailArkhipov: This is extremely heavy, loading module causes analysis and will load the universe. Should only be used if module is actually imported by the user directly or indirectly.

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
```diff
@@ -0,0 +1,635 @@
...
+            var pathResolver = analysis.Document.Interpreter.ModuleResolution.CurrentPathResolver;
+
+            var modules = ImmutableArray<IPythonModule>.Empty;
+            foreach (var moduleName in symbolsWithName.Select(s => s.DocumentPath).Distinct().Select(p => pathResolver.GetModuleNameByPath(p))) {
+                cancellationToken.ThrowIfCancellationRequested();
+
+                var module = analysis.Document.Interpreter.ModuleResolution.GetOrLoadModule(moduleName);
+                if (module.Analysis is EmptyAnalysis) {
+                    // once module is analyzed, this analysis result will be kept in memory
+                    // even if graph for current document (modified) changed.
```
MikhailArkhipov: As above, loading modules not explicitly use is heavy, both time and memory-wise. IMO, module source can be parsed and indexed via workspace indexer, but analysis of something user did not import is a no-no.

src/LanguageServer/Impl/Implementation/Server.cs
heejaechang: sure. I can remove {}.
by the way, if only issue is "diff", then you can use "No whitespace" (?w=1) to remove those whitespace diff from PR.

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
heejaechang: sure, I can do those.

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
heejaechang: thank you.

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
heejaechang: for things like os.path, I dont think we can support it without analyzing the module. we could have cache created for libraries but right now it doesn't exist. so that is not an option for now.

so, how much this different than find all references? find all references doesn't load libraries?

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
heejaechang: so, how heavy will it be? so, we are going back to only offering one that is already loaded by users?

MikhailArkhipov: Very heavy. Basically nothing should be loading modules for analysis unless it is an explicit user action. Rename/find all references does it in a limited fashion, typically it effectively scopes to loaded libraries and user code. Loading and analyzing libraries like `plotly` may take minutes (just this one), consume hundreds mb of memory. Attempt to load something like `conda` environment will definitely kill machine with 8GB RAM. And burn battery charge on analysis. In short - analysis is expensive action and should be avoided at all costs.

If we are to offer items from all universe then approach should be as it is for indexing workspace - parsing and lightweight identification of possible items. It is pretty easy to find top level classes, functions and variables and this probably covers 95%.

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
MikhailArkhipov: Find references does load modules and I think it actually does too much - which may explain some bugs about memory consumption. IMO, after finding root definition it should only be scanning user code. I don't think user is interested in references elsewhere.

jakebailey: I showed him what happens in my awful conda environment from the PyQt bug. We're going to try and drop this for now and rely only on indexing.

MikhailArkhipov: I meant find all references but nvm, it is limited to root which should be just user code.

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
heejaechang: will make a change and ping you guys again

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
heejaechang: will make the change and send out update

heejaechang:
made not to load modules. can you take another look?

src/Analysis/Ast/Impl/Modules/Resolution/MainModuleResolution.cs
```diff
@@ -55,6 +55,20 @@ public MainModuleResolution(string root, IServiceContainer services, ImmutableAr
...
 
+        public IEnumerable<IPythonModule> GetImportedModules(CancellationToken cancellationToken) {
+            foreach (var moduleRef in Modules.Values) {
+                cancellationToken.ThrowIfCancellationRequested();
+                if (moduleRef.Value != null) {
+                    yield return moduleRef.Value;
+                }
+            }
+
+            foreach (var module in _specialized.Values) {
```
MikhailArkhipov: You may want to swap it with the top since specialized usually win over.

src/Analysis/Core/Impl/DependencyResolution/PathResolverSnapshot.cs
```diff
@@ -99,13 +111,21 @@ public PathResolverSnapshot(PythonLanguageVersion pythonLanguageVersion, string
...
                 _builtins.Children
-                    .Where(b => !string.IsNullOrEmpty(b.FullModuleName))
-                    .Select(b => b.FullModuleName)
+                    .Where(b => predicate(b))
+                    .Select(b => valueGetter(b))
             );
         }
 
+        public string GetModuleNameByPath(string modulePath) {
+            if (TryFindModule(modulePath, out var edge, out _)) {
```
MikhailArkhipov: convertable to `=> ?` form I think

src/Core/Impl/Extensions/ArrayExtensions.cs
```diff
@@ -36,5 +37,21 @@ public static class ArrayExtensions {
 
             return -1;
         }
+
+        public static List<T> AddIfNotNull<T>(this List<T> list, T item) where T : class {
```
MikhailArkhipov: Maybe just `this ICollection<T>`?

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
```diff
@@ -0,0 +1,731 @@
...
+            return codeActions;
+
+            string GetModuleName(string moduleFullName) {
+                var index = moduleFullName.LastIndexOf(".");
+                if (index < 0) {
+                    return moduleFullName;
+                }
+
+                return moduleFullName.Substring(index + 1);
+            }
```
MikhailArkhipov: I think it is just `moduleFullName.Substring(moduleFullName.LastIndexOf(".")+ 1)`... since it returns -1. Or
 `return index < 0 ? moduleFullName : moduleFullName.Substring(index + 1);

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
```diff
@@ -0,0 +1,731 @@
...
+
+            bool FilterSimpleName(KeyValuePair<string, ImportInfo> kv) => kv.Key.IndexOf(".") < 0;
+            bool FilterImportedMembers(KeyValuePair<string, ImportInfo> kv) => !kv.Value.IsModule && kv.Value.MemberImported;
+            bool FilterImportedModuleMembers(KeyValuePair<string, ImportInfo> kv) => !kv.Value.IsModule && kv.Value.ModuleImported;
+            bool FilterImportedModules(KeyValuePair<string, ImportInfo> kv) => kv.Value.IsModule && kv.Value.MemberImported;
+
+            bool FilterSourceDeclarations(IGrouping<LocationInfo, KeyValuePair<string, ImportInfo>> group) {
+                var count = 0;
+                foreach (var entry in group) {
```
MikhailArkhipov: So what does it do? Looks like it works with first value, if it exists and then hard-return on second? Seems a bit complicated.... Am I missing something

heejaechang: there is a comment explaining what it does at the call site (at the top of same method)

src/LanguageServer/Impl/CodeActions/MissingImportCodeActionProvider.cs
heejaechang: sure

heejaechang:
@MikhailArkhipov any more feedback?

Astrantia:
@heejaechang Will it be possible to add an auto import feature? Importing with a quickfix is great, but it's the best when unresolved imports are auto imported on save.

jakebailey:
There are so many options for what a particular unresolved name could be that it might be impractical to do it right on save with no user input, or in cases where the unresolved name is just a local error (and not intended to be an import). We're going to start simple and go from there.

(Also note that on-save is more of a formatting operation, which we don't currently have plans to handle directly.)

MikhailArkhipov:
@heejaechang - I'd test it with Anaconda to be sure there are no performance issues and demo it to @Anapo14 to make sure behavior matches spec expectations. Otherwise LGTM.

heejaechang:
thank you!

grisaitis:
i see that https://github.com/microsoft/python-language-server/commit/254c8f17d1dcee3057203c94010efb542c4b3ec3 is not included in any releases, just master. does this mean it's not available in stable vs code?

jakebailey:
See https://github.com/microsoft/python-language-server/issues/19#issuecomment-573902686 (where you also asked this). You can ignore the GitHub releases.

PR #1662, opened 2019-10-10
Remove invalid operation

### LINKED ISSUE #1651: InvalidOperationException in DocumentBuffer.Update on didChange
Seeing this a lot in the list:

```
   at Microsoft.Python.Core.Diagnostics.Check.InvalidOperation(Boolean condition, String message) in E:\A\_work\2\s\src\Core\Impl\Diagnostics\Check.cs:line 86
   at Microsoft.Python.Analysis.Documents.DocumentBuffer.Update(IEnumerable`1 changes) in E:\A\_work\2\s\src\Analysis\Ast\Impl\Documents\DocumentBuffer.cs:line 72
   at Microsoft.Python.Analysis.Modules.PythonModule.Update(IEnumerable`1 changes) in E:\A\_work\2\s\src\Analysis\Ast\Impl\Modules\PythonModule.cs:line 315
   at Microsoft.Python.LanguageServer.Implementation.Server.DidChangeTextDocument(DidChangeTextDocumentParams params) in E:\A\_work\2\s\src\LanguageServer\Impl\Implementation\Server.Documents.cs:line 36
   at Microsoft.Python.LanguageServer.Implementation.LanguageServer.DidChangeTextDocument(JToken token, CancellationToken cancellationToken) in E:\A\_work\2\s\src\LanguageServer\Impl\LanguageServer.cs:line 132
```

MikhailArkhipov:
User may try and edit library file where we have already dropped the content.
Fixes #1651

### DIFF
--- a/src/Analysis/Ast/Impl/Documents/DocumentBuffer.cs
+++ b/src/Analysis/Ast/Impl/Documents/DocumentBuffer.cs
@@ -40,7 +40,9 @@ namespace Microsoft.Python.Analysis.Documents {
         public void SetContent(string content) {
             lock (_lock) {
                 Check.InvalidOperation(!_initialized, "Buffer is already initialized.");
-                Check.InvalidOperation(!_cleared, "Buffer cannot be updated since its content was dropped.");
+                if (_cleared) {
+                    return; // User may try and edit library file where we have already dropped the content.
+                }
                 Version = 0;
                 _content = content ?? string.Empty;
                 _sb = null;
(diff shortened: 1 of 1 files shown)


src/Analysis/Ast/Impl/Documents/DocumentBuffer.cs
```diff
@@ -59,16 +61,20 @@ internal sealed class DocumentBuffer {
         public void MarkChanged() {
             lock (_lock) {
                 Check.InvalidOperation(_initialized, "Buffer is not initialized.");
-                Check.InvalidOperation(!_cleared, "Buffer cannot be updated since its content was dropped.");
+                if(_cleared) {
```
jakebailey: Nit: formatting

jakebailey:
Brainstorming, but how difficult would it be to allow this sort of thing, reload the content to apply the edits, mark the document as edited, invalidate the analysis, then not convert it to LibraryAnalysis when the document is seen as changed?

Doing nothing would fix the exception, of course, just thinking about what the "right" thing to do is if a library file is edited.

MikhailArkhipov:
VSC does not know we dropped the content so it does not send it again, only changes. It could be, I guess, reloaded from disk and invalidated. Tracking as not to consider it library anymore is more complicated since it will need to override module type so anywhere we do direct check for module type we'd have to make some function call instead to determine if module is a library. Kinda heavy change for a limited use case.

Issue #1232, opened 2019-06-20
super() does not return an instance of the base class


jakebailey:
![image](https://user-images.githubusercontent.com/5341706/59868468-d1006000-9345-11e9-8133-5d2760d9f77b.png)

![image](https://user-images.githubusercontent.com/5341706/59868532-f4c3a600-9345-11e9-9cae-6a54961155a9.png)

```python
class Base:
    def base_func(self):
        return 1234


class Child(Base):
    def child_func(self):
        x = super()
```

Likely the cause of #1130.

jakebailey:
Fixed by #1702.

Issue #1130, opened 2019-05-28
Can't jump to method or have intellisense from parent while using super()


aaveidt:
If i use super().method() to call to parent class method, there is no intellisense or argument completion.

Then it's impossible to click to that method.

But if i use self / cls, it works.

<img width="486" alt="Screen Shot 2019-05-28 at 12 59 39 PM" src="https://user-images.githubusercontent.com/5781080/58454394-875a9600-8148-11e9-8f18-4ea5414e339d.png">

<img width="433" alt="Screen Shot 2019-05-28 at 12 58 24 PM" src="https://user-images.githubusercontent.com/5781080/58454467-bec94280-8148-11e9-8b0d-f1c927338fe8.png">

jakebailey:
Fixed by #1702.

PR #1702, opened 2019-10-17
Adding super as a specialized keyword

bschnurr:
super
Fix for #1232
Fix for #1130


src/Analysis/Ast/Test/InheritanceTests.cs
```diff
@@ -113,5 +113,65 @@ class A(collections.namedtuple('A', [])):
                 .Which.Value.Should().BeAssignableTo<IPythonInstance>()
                 .Which.Type.Name.Should().Be("A");
         }
+
+        [TestMethod, Priority(0)]
+        public async Task SingleInheritanceSuperShouldReturnBaseClassType() {
+            const string code = @"
+class Baze:
+  def base_func(self):
```
jakebailey: Four space indents across the board for python code.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +92,23 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
             return args.Count > 0 && args[0] is PythonCollection c ? c.Contents.FirstOrDefault() : null;
         }
 
+        public static IMember Super(IPythonModule declaringModule, IPythonFunctionOverload overload, IArgumentSet argSet, IndexSpan indexSpan) {
+            foreach (var s in argSet.Eval.CurrentScope.EnumerateTowardsGlobal) {
+                if (s.Node is ClassDefinition) {
+                    var classType = s.Variables["__class__"].GetPythonType<IPythonClassType>();
```
MikhailArkhipov: `s.Node is ClassDefinition cd`
`s.Variables[cd.Name]?.Value as IPythonClassType`

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +92,23 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
+
+                    if (classType?.Mro != null) {
+                        var proxySuper = new PythonSuperType(new Location(declaringModule, indexSpan), classType.Mro);
```
MikhailArkhipov: location is used for goto definition and should be location of the base class, i.e. where does 'goto def' navigates from `super`? `ILocatedMember` may need to be beefed up to hold multiple locations where there are multiple bases or new interface created and handled in the definition source.

bschnurr: I see thanks

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
bschnurr: s.Variables[cd.Name]?.Value as IPythonClassType wasn't working for me. I'll take another look

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
bschnurr: I'll tried grabbing the next class in the mro list and using its location.. but in the editor gotodefinition still isnt' working. I just added a test to investigate more

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +92,50 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
             return args.Count > 0 && args[0] is PythonCollection c ? c.Contents.FirstOrDefault() : null;
         }
 
+        public static IMember Super(IPythonModule declaringModule, IPythonFunctionOverload overload, IArgumentSet argSet, IndexSpan indexSpan) {
+            if (argSet.Arguments.Any()) {
+                //multi argument form can have 1 or 2 arguments
+                var classType = argSet.Argument<IMember>(0).GetPythonType<IPythonClassType>();
+                var objOrType = argSet.Argument<IMember>(1).GetPythonType<IPythonClassType>();
```
bschnurr: objOrType , I think I need to try to cast to a python instance too.  IPythonInstance ?

src/LanguageServer/Test/GoToDefinitionTests.cs
```diff
@@ -165,6 +165,45 @@ class D(C):
             reference.range.Should().Be(2, 4, 2, 5);
         }
 
+        [TestMethod, Priority(0)]
+        public async Task GotoDefinitionFromParentOtherModuleOnSuper() {
```
bschnurr: new goto definition test (doesnt work currently) reference is null

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
jakebailey: What are you goto def-ing on? `super()`, or its return value? Because `super` itself is specialized, it won't end up going anywhere. Its return value's goto-def might not do what you expect either, since assigning it away means that it's "defined" at the assignment.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
MikhailArkhipov: `GetPythonType` handles instances. It is an extension to `IMember` which includes instances, variables and types.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
MikhailArkhipov: Next class may not work, if it is, say `Generic` of a sort. You may try to enumerate and fetch first that can be navigated to like `ILocatedMember.Location.IsValid`.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
bschnurr: updated.. a bit more complex to figure out isinstance or issubclass.. should they be helper functions?

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +93,79 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
+
+                // second argument optional
+                bool isUnbound = argSet.Argument<IMember>(1) == null;
+                if (isUnbound) {
+                    return CreateSuper(argSet, classType.Mro);
+                } else {
+                    // Second argument can be either an object instance or a class which are both IPythonType
+                    IPythonType objOrType = null;
+                    objOrType = argSet.Argument<IMember>(1).GetPythonType();
+                    bool isInstance = objOrType?.Equals(classType) ?? false;
```
bschnurr: is there a helper function for isInstance ?

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +93,79 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
+
+                    if (!isInstance) {
+                        var argAsType = argSet.Argument<IMember>(1).GetPythonType<IPythonClassType>();
+                        bool isSubClass = classType?.Bases.Any(b => argAsType.Equals(b)) ?? false;
```
bschnurr: is there a helper function for isSubClass?

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +93,79 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
+                var classType = argSet.Argument<IMember>(0).GetPythonType<IPythonClassType>();
+                if (classType?.Mro == null) {
+                    return null;
+                }
+
+                // second argument optional
+                bool isUnbound = argSet.Argument<IMember>(1) == null;
+                if (isUnbound) {
+                    return CreateSuper(argSet, classType.Mro);
+                } else {
```
MikhailArkhipov: `else` is not needed since `if` returns

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +93,79 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
             return args.Count > 0 && args[0] is PythonCollection c ? c.Contents.FirstOrDefault() : null;
         }
 
+        public static IMember Super(IPythonModule declaringModule, IPythonFunctionOverload overload, IArgumentSet argSet, IndexSpan indexSpan) {
+            if (argSet.Arguments.Any()) {
```
MikhailArkhipov: I think code generally prefers short returns so 
```cs
if(argSet.Arguments.Count == 0) {
    return null;
}
```

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
MikhailArkhipov: `is IPythonInstance`. But do you care? `IVariable` can also be used I think, i.e.
```python
class A: ...
x = A()
class B(x): ...
```
`x` will be `IVariable` with value of `IPythonInstance` which has type of `IPythonClassType`.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
MikhailArkhipov: argAsType can be null so you may end up calling `Equals` on null. Either check or reverse comparison like `b.Equals(argAsType)`. Also, `Bases` can be null. Feel free to write helper like `IsSuperClassOf(this IPythonClassType, ...)`.

MikhailArkhipov: Also, I think if class belongs to bases it is a `superClass`. Subclass is whatever is derived.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +93,79 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
+                    IPythonType objOrType = null;
+                    objOrType = argSet.Argument<IMember>(1).GetPythonType();
+                    bool isInstance = objOrType?.Equals(classType) ?? false;
+
+                    if (!isInstance) {
+                        var argAsType = argSet.Argument<IMember>(1).GetPythonType<IPythonClassType>();
+                        bool isSubClass = classType?.Bases.Any(b => argAsType.Equals(b)) ?? false;
+                        if (isSubClass) {
+                            objOrType = argAsType as IPythonType;
+                        }
```
MikhailArkhipov: `objOrType  = classType?.Bases.MaybeEnumerate().FirstOrDefault(b => b.Equals(argAsType))`

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +93,79 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
+                if (start >= 0) {
+                    mro = mroList.GetRange(start, mro.Count() - start).ToArray();
+                } else {
+                    return null;  // newStartType wasn't valid
+                }
+            }
+            
+            if (mro.Any()) {
+                var nextClassInLine = mro.FirstOrDefault();
+                var location = new Location(nextClassInLine.Location.Module, nextClassInLine.Location.IndexSpan);
```
MikhailArkhipov: Double enumeration. `mro` is list so just check `Count` or do `var nextClassInLine = mro.FirstOrDefault()` and then check `nextClassInLine` for null.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +93,79 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
+                            return CreateSuper(argSet, classType.Mro);
+                        }
+                    }
+                }
+            }
+            return null;
+        }
+
+        private static IMember CreateSuper(IArgumentSet argSet, IReadOnlyList<IPythonType> callerMro, IPythonType newStartType = null) {
+            if (!callerMro.Any()) {
```
MikhailArkhipov: `callerMro.Count == 0`

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
bschnurr: ahh i see Jake,  Mikhail I'll see what happens with Generic

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
bschnurr: ya i had the check ```nextClassInLine``` for null first then removed it

bschnurr: i'll add it back

src/LanguageServer/Test/GoToDefinitionTests.cs
```diff
@@ -165,6 +165,45 @@ class D(C):
...
+    def test(self):
+        pass
+";
+            const string testModCode = @"
+from other import C
+
+class D(C):
+    def hello(self):
+        super().test();
+        super().v
```
bschnurr: Issue with goto definition on data member super().v

This test is copied from the test above ```GotoDefinitionFromParentOtherModule()```

```DefinitionSources::FromMemberExpression(..)``` seems to special case data members but only if the type is ```IPythonClassType``` , currently my ```PythonSuperType``` is not implementing that interface so then it falls back to just calling ``` if (type?.GetMember(mex.Name) is ILocatedMember lm) { ``` but i guess data members dont track their declaration location.

File: src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs (hunk 3 above)
Comment (bschnurr) [reply src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs]: I think i dont want goto definition to work on super() itself. 

I'm going to focus on the cases of goto on what comes after super(). ie super().func() and super.x

File: src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs (hunk 7 above)
Comment (bschnurr) [reply src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs]: I added IsSubClassOf

File: src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs (hunk 6 above)
Comment (bschnurr) [reply src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs]: I added IsInstanceOf(..)

File: src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs (hunk 7 above)
Comment (bschnurr) [reply src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs]: Good catch. ya I found that null case when testing.

I'm actually looking for derived, so I made IsSubClassOf(..), I re-read the python3 docs and I was traversing the wrong mro.

Comment (bschnurr) [reply src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs]: I went with IsSubClassOf because the docs mention object-type(param2) being a subclass of type(param1)

File: src/Analysis/Ast/Impl/Types/PythonSuperType.cs [hunk 14]
```diff
@@ -0,0 +1,49 @@
...
+        public PythonSuperType(Location location, IReadOnlyList<IPythonType> mro) 
+            : base("super", location, string.Empty, BuiltinTypeId.Type) {
+            _mro = mro;
+        }
+
+        public IReadOnlyList<IPythonType> Mro => _mro;
+
+        public override IMember GetMember(string name) {
+            foreach (var cls in _mro?.OfType<IPythonClassType>().MaybeEnumerate()) {
+                var member = cls.GetMember(name);
```
MikhailArkhipov: `=> _mro.MaybeEnumerate().Select(c => c.GetMember(name)).FirstOrDefault()`

src/Analysis/Ast/Impl/Types/PythonSuperType.cs
```diff
@@ -0,0 +1,49 @@
...
+                if (member != null) {
+                    return member;
+                }
+            }
+            return null;
+        }
+
+        public override IEnumerable<string> GetMemberNames() => (_mro?.OfType<IPythonClassType>().MaybeEnumerate()).SelectMany(cls => cls.GetMemberNames().Select(name => name));
+    }
```
MikhailArkhipov: I don't think last `.Select` is needed, but you may choose to add `Distinct`.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +94,79 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
             return args.Count > 0 && args[0] is PythonCollection c ? c.Contents.FirstOrDefault() : null;
         }
 
+        public static IMember Super(IPythonModule declaringModule, IPythonFunctionOverload overload, IArgumentSet argSet, IndexSpan indexSpan) {
```
MikhailArkhipov: Now, speaking of... Would be simpler to just declare `super` variable at the class scope in the `ClassEvaluator` like we declare `__class__`? The variable value would be a `SuperProxy` type. Then `super()` will automatically become instance since `super` sill be found up in scopes`.

jakebailey: Maybe, but `super` does take arguments and has to be able to process them anyway (since you can do `super` anywhere you want, not just in a class). It might also make changes fixing our class variable scoping (the `self.` requirement) more complicated. A neat trick, though.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
MikhailArkhipov: I think it should be `IsBaseOf` or `IsSuperClassOf` since base == superclass. Subclass is what is derived from the base (i.e narrowing operation). Cat is a subclass of an Animal. so Animal is base/superclass of Cat.

src/Analysis/Ast/Impl/Types/PythonSuperType.cs
bschnurr: good catch

bschnurr: Ya I think your right about Distinct.

MikhailArkhipov: Also, do we care if they are `OfType`? `GetMember` works on `IMember`.

bschnurr: Do we need to sort?

MikhailArkhipov: No, completion list sorts

src/Analysis/Ast/Impl/Types/PythonSuperType.cs
bschnurr: nice

src/LanguageServer/Test/GoToDefinitionTests.cs
MikhailArkhipov: I would add `interface IMultipleLocationsMember: IEnumerable<Location>` which would be able to store multiple locations - i.e. all bases. In DefinitionSource code then can check if item is `IMultipleLocationsMember` and fetch and return all locations so in case of multiple inheritance user would see list with choices.

jakebailey: I'm not sure if a multi-location is needed. If you search the MRO for the correct member, then that's going to have a location and that's where you'd go, like any other access. super is only going to give back the first found member for any given name, and I think the only way to get a specific base's member would be to pass more arguments to super to choose an alternate MRO, or refer to the type directly and bind its functions.

Finding alternative things that could be called based on children classes (i.e. goto impl) is another story, though.

src/Analysis/Ast/Impl/Types/PythonSuperType.cs
bschnurr: updated

src/Analysis/Ast/Impl/Types/PythonSuperType.cs
bschnurr: added distinct and tests in completions

src/LanguageServer/Impl/Sources/DefinitionSource.cs
```diff
@@ -286,7 +286,19 @@ internal abstract class DefinitionSourceBase {
...
+                    foreach (var b in cls.Mro.OfType<IPythonClassType>()) {
+                        using (eval.OpenScope(b)) {
+                            eval.LookupNameInScopes(mex.Name, out _, out var v2, LookupOptions.Local);
+                            if (v2 != null) {
+                                definingMember = v2;
+                                return FromMember(v2);
+                            }
+                        }
+                    }
+                    break;
```
bschnurr: This is the code needed for gotodefinition to work for super().v   its just a copy of the code below

src/LanguageServer/Test/CompletionTests.cs
```diff
@@ -1417,5 +1418,51 @@ class B(A):
...
+
+class C(B):
+    def foo(self):
+        super().
+";
+            var analysis = await GetAnalysisAsync(code);
+            var cs = new CompletionSource(new PlainTextDocumentationSource(), ServerSettings.completion, Services);
+            var result = cs.GetCompletions(analysis, new SourceLocation(12, 17));
+
+            result.Completions.GroupBy(item => item.insertText).Any(g => g.Count() > 1).Should().BeFalse();
```
bschnurr: new distinct tests

src/LanguageServer/Impl/Sources/DefinitionSource.cs
bschnurr: I could make a new interface that combines IPythonClassType and IPythonSuperType

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +94,79 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
+            }
+
+            // If multiple arguments first argument is required
+            var firstCls = args.FirstOrDefault().GetPythonType<IPythonClassType>();
+            if (firstCls == null) {
+                return null;
+            }
+
+            // second argument optional
+            bool isUnbound = args.Count() == 1;
```
heejaechang: do we know whether IReadOnlyList is treated special in Linq? I just checked (https://github.com/dotnet/corefx/blob/master/src/System.Linq/src/System/Linq/Count.cs) and it is not. meaning this use Enumerator unnecessarily since IReadOnlyList provide indexer and Count property?

bschnurr: good point. I'll avoid Count() and use .Count instead

bschnurr: updated

src/LanguageServer/Impl/Sources/DefinitionSource.cs
jakebailey: If the only thing in common for these is the MRO, then maybe an MRO is something that every type should have? Spitballing.

jakebailey: If it were me I'd stick with a little bit of copying and a TODO rather than trying to abstract this immediately.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
MikhailArkhipov: Yeah b/c variable is declared in outer scope. I.e. `s.OuterScope.Variables[cd.Name]?.Value as IPythonClassType`. But `__class__` is fine too.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +94,79 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
             return args.Count > 0 && args[0] is PythonCollection c ? c.Contents.FirstOrDefault() : null;
         }
 
+        public static IMember Super(IPythonModule declaringModule, IPythonFunctionOverload overload, IArgumentSet argSet, IndexSpan indexSpan) {
+            var args = argSet.Values<IMember>();
+
+            if (args.Count == 0) {
+                //Zero argument form only works inside a class definition
+                foreach (var s in argSet.Eval.CurrentScope.EnumerateTowardsGlobal) {
```
MikhailArkhipov: A bit shorter
```
foreach (var s in argSet.Eval.CurrentScope.EnumerateTowardsGlobal.Where(s.Node is ClassDefinition))

File: src/Analysis/Ast/Impl/Types/Definitions/IPythonSuperType.cs [hunk 21]
```diff
@@ -0,0 +1,28 @@
...
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Python.Analysis.Types {
+    /// <summary>
+    /// Represents Python class type definition.
+    /// </summary>
+    public interface IPythonSuperType {
```
MikhailArkhipov: May also modify `interface IPythonClassType: IPythonSuperType`

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +94,79 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
+            IReadOnlyList<IPythonType> callerMro,
+            IPythonModule declaringModule,
+            IndexSpan indexSpan,
+            IPythonType typeToFind = null) {
+
+            if (callerMro.Count == 0) {
+                return null;
+            }
+
+            IReadOnlyList<IPythonType> mro = callerMro;
```
MikhailArkhipov: `var`

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +94,79 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
+
+            // skip doing work if the newStartType is the first element in the callers mro
+            if (typeToFind != null &&
+                !typeToFind.Equals(callerMro.FirstOrDefault())) {
+                var mroList = callerMro.ToList();
```
MikhailArkhipov: `typeToFind?.Equals(callerMro.FirstOrDefault()) == false` then no need to check for null. Basically `bool? == value`

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +94,79 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
+                var start = mroList.FindIndex(0, t => t.Equals(typeToFind));
+                if (start >= 0) {
+                    mro = mroList.GetRange(start, mro.Count() - start).ToArray();
```
MikhailArkhipov: `mro.Count`

bschnurr: thanks.. thought I got all these

src/Analysis/Ast/Impl/Types/PythonSuperType.cs
```diff
@@ -0,0 +1,41 @@
...
+        /// more info at https://docs.python.org/3/library/functions.html#super
+        /// </summary>
+        /// <param name="location"></param>
+        /// <param name="mro">Should be a list of IPythonType</param>
+        public PythonSuperType(Location location, IReadOnlyList<IPythonType> mro)
+            : base("super", location, string.Empty, BuiltinTypeId.Type) {
+            _mro = mro;
+        }
+
+        public IReadOnlyList<IPythonType> Mro => _mro;
```
jakebailey: `Mro` and `_mro` can be combined into `Mro { get; }`.

src/LanguageServer/Test/GoToDefinitionTests.cs
MikhailArkhipov: OK

MikhailArkhipov:
LGTM, up to @jakebailey

jakebailey:
Also think it's looking good, but you're likely also going to have to verify that it works in persistence (and unfortunately possibly add a model).

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -91,6 +93,73 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
+        }
+
+        private static IMember CreateSuper(IArgumentSet argSet,
+            IPythonClassType classType,
+            IPythonModule declaringModule,
+            IndexSpan indexSpan,
+            IPythonType typeToFind = null) {
+
+            var mro = classType?.Mro ?? Array.Empty<IPythonType>();
+            if (mro.Count == 0) {
```
bschnurr: I moved the null checking from outside of CreateSuper, to here, tried to combine null checking with count == 0 checking. came up with this

src/Caching/Test/ClassesTests.cs
```diff
@@ -150,5 +150,21 @@ class B(A):
             // In JSON, class A should have 'class A doc' documentation while B should have none.
             Baseline.CompareToFile(BaselineFileName, json);
         }
+
+         [TestMethod, Priority(0)]
+        public void ClassesWithSuper() {
```
jakebailey: I would recommend writing the test (almost) as though it worked, then sticking `[Ignore("some message")]` on it, rather than something commented out.

src/Caching/Test/ClassesTests.cs
```diff
@@ -150,5 +150,21 @@ class B(A):
...
+//            const string code = @"
+//class A:
+//    def methodA(self):
+//        return True
+
+//class B(A):
+//    def methodB(self);
+//        x = super()
```
jakebailey: The persistence I would be testing here is `return super()`, since `x` is a local variable that will not be persisted.

src/LanguageServer/Test/GoToDefinitionTests.cs
```diff
@@ -204,6 +204,46 @@ class D(C):
...
+            def dowork(self):
+                return super().dowork()
+
+        class Mom():
+            def dowork(self):
+                return 2
+
+        class Child(Dad, Mom):
+            def child_func(self):
+                x = super()
```
jakebailey: `x` is a local variable and would not be persisted.

bschnurr: I can remove this line

jakebailey: No big deal, misread the test class.

src/Caching/Test/ClassesTests.cs
bschnurr: thanks updated

src/Caching/Test/ClassesTests.cs
bschnurr: updated

src/Analysis/Ast/Impl/Types/PythonSuperType.cs
```diff
@@ -29,6 +29,12 @@ public PythonSuperType(Location location, IReadOnlyList<IPythonType> mro)
             Mro = mro;
         }
 
+        public override string QualifiedName {
+            get {
+                return $"{this.GetQualifiedName()}[{string.Join(",", Mro.Select(t => t.QualifiedName))}]";
+            }
+        }
+
```
bschnurr: Example json "ReturnType": "i:module:super[module:A,object]"

src/Caching/Impl/Models/SuperModel.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        public SuperModel() {
+            Name = "super";
+            IndexSpan = new IndexSpanModel();
+        } // For de-serializer from JSON
+
+        
+        public override IMember Create(ModuleFactory mf, IPythonType declaringType, IGlobalScope gs) {
+            var st = new PythonSuperType(new Location(mf.Module, IndexSpan.ToSpan()), Mro);
+            return st;
```
bschnurr: wasn't sure what to do for default location. it gets set after ```Create``` is called

src/Caching/Impl/ModuleFactory.cs
```diff
@@ -105,6 +105,10 @@ public IPythonType ConstructType(string qualifiedName)
                     return null;
                 }
 
+                if(nextModel is SuperModel sm) {
+                    sm.Mro = typeArgs;
+                }
```
bschnurr: ```memberName = "super"```, but I need a way to pass the mro to the superModel, so I made something specific here, then just below Create gets called

src/Caching/Impl/Models/SuperModel.cs
bschnurr: Maybe I should clear the ```Mro ``` afterwards and add some asserts too

src/Caching/Impl/Models/ModuleModel.cs
```diff
@@ -69,6 +70,9 @@ internal sealed class ModuleModel : MemberModel {
             var classes = new Dictionary<string, ClassModel>();
             var typeVars = new Dictionary<string, TypeVarModel>();
             var namedTuples = new Dictionary<string, NamedTupleModel>();
+            var superTypes = new Dictionary<string, SuperModel>();
+
+            superTypes["super"] = new SuperModel();
```
jakebailey: Since the entire super type can be described by its qualified name, I don't think you actually need to end up with a model for it, just create the type on the fly during the restore.

src/Analysis/Ast/Impl/Types/PythonSuperType.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// <param name="location"></param>
+        /// <param name="mro">Should be a list of IPythonType</param>
+        public PythonSuperType(Location location, IReadOnlyList<IPythonType> mro)
+            : base("super", location, string.Empty, BuiltinTypeId.Type) {
+            Mro = mro;
+        }
+
+        public override string QualifiedName {
+            get {
+                return $"{this.GetQualifiedName()}[{string.Join(",", Mro.Select(t => t.QualifiedName))}]";
```
jakebailey: I don't think you need to qualify the name with a location or anything, since the "super type" isn't defined anywhere except in our code. I was imagining this qualified name being something as simple as `s:[a,bunch,of,things]`, rather than calling `GetQualifiedName`.

bschnurr: I'm running into issues using "s" and ```DeconstructQualifiedName```

jakebailey: Yeah, that function expects a certain number of elements in the name, as everything has had a declaring module up until now.

Maybe you can just leave the declaring module as it is and still special case it, i.e. create `s:<module>:[MRO]` which you can expand into `PythonSuperType`, and don't allow any part of the code to do a lookup into the module for the value.

MikhailArkhipov:
I think in 
```python
def sup(self):
        return super()
```
`sup` should have `StaticReturnValue` evaluated to actual value of `super()` and the qualified name of the evaluated call should be persisted (i.e. overload model should have actual return type of `super`).

jakebailey:
Yes, but the issue is that if we don't modify any persistence code, the super proxy gets persisted as "instance of class `super` defined in the module where it was used", which falls apart because there isn't actually a class called "super" in the module where it was used.

Storing the entirety of the super proxy's state is certainly possible, as the MRO in IPythonSuperType is just of more types, whose qualified names we can store in a list, then restore fully to the original state.

MikhailArkhipov:
OK, then perhaps qualified name of the `IPythonSuperType` should be the actual underlying value. Like with `PythonTypeWrapper`? Like `t:baseClass`? It probably does not need dedicated model.

jakebailey:
Yes, that's what I'm thinking. `s:[some,list,that,is,mro]` or similar.

src/Caching/Impl/Models/ModuleModel.cs
```diff
@@ -38,7 +38,7 @@ internal sealed class ModuleModel : MemberModel {
         public ClassModel[] Classes { get; set; }
         public TypeVarModel[] TypeVars { get; set; }
         public NamedTupleModel[] NamedTuples { get; set; }
-
+
```
jakebailey: Nit: added whitespace

src/Caching/Impl/TypeNames.cs
```diff
@@ -89,9 +89,10 @@ internal static class TypeNames {
         private static void GetModuleNameAndMembers(string qualifiedName, ref QualifiedNameParts parts, int prefixOffset) {
             // Strip the prefix, turning i:module:A.B.C into module:A.B.C
             var typeName = qualifiedName.Substring(prefixOffset);
-
+
```
jakebailey: Nit: whitespace

src/Analysis/Ast/Impl/Types/PythonSuperType.cs
```diff
@@ -0,0 +1,45 @@
...
+        /// more info at https://docs.python.org/3/library/functions.html#super
+        /// </summary>
+        /// <param name="location"></param>
+        /// <param name="mro">Should be a list of IPythonType</param>
+        public PythonSuperType(Location location, IReadOnlyList<IPythonType> mro)
+            : base("super", location, string.Empty, BuiltinTypeId.Type) {
+            Mro = mro;
+        }
+
+        public override string QualifiedName {
```
jakebailey: Style nit, but you can probably just use arrow notation here. Was likely a block in other places due to locking or being more complicated.

src/Analysis/Ast/Impl/Types/PythonSuperType.cs
```diff
@@ -0,0 +1,45 @@
...
+            get {
+                return $":SuperType[{string.Join(",", Mro.Select(t => t.QualifiedName))}]";
```
jakebailey: Does this need the leading `:`, or can it be omitted like with `int` or `str`?

src/Caching/Test/ClassesTests.cs
```diff
@@ -152,7 +152,6 @@ class B(A):
         }
 
         [TestMethod, Priority(0)]
-        [Ignore("Todo: super() persistence support, Asserts in debug due to no model found for super")]
         public async Task ClassesWithSuper() {
```
bschnurr: should I move these super tests to its own file?

jakebailey: Might be a good idea, yeah.

src/Analysis/Ast/Impl/Types/PythonSuperType.cs
bschnurr: If you look at the change to ```TypeNames.cs``` I needed a way to parse the module name as empty but not get messed up and find the first ":" inside my mro list "[class a:b, class a:c]

jakebailey: Alright.

src/Analysis/Ast/Impl/Types/PythonSuperType.cs
```diff
@@ -0,0 +1,45 @@
...
+using Microsoft.Python.Core;
+
+namespace Microsoft.Python.Analysis.Types {
+    internal sealed class PythonSuperType : PythonType, IPythonSuperType {
+        /// <summary>
+        /// more info at https://docs.python.org/3/library/functions.html#super
+        /// </summary>
+        /// <param name="location"></param>
+        /// <param name="mro">Should be a list of IPythonType</param>
+        public PythonSuperType(Location location, IReadOnlyList<IPythonType> mro)
```
bschnurr: I'm going to remove the location here and just use an empty location for the base parameter

src/Caching/Test/ClassesTests.cs
```diff
@@ -150,5 +150,47 @@ class B(A):
             // In JSON, class A should have 'class A doc' documentation while B should have none.
             Baseline.CompareToFile(BaselineFileName, json);
         }
+
```
bschnurr: todo: need to remove theses now that they have their own file.


Also I'm worried about persistence and base class is namedtuple because classModel special cases that

jakebailey:
LGTM, caching stuff appears to work fine. All tests pass (sans stuff that was already broken in master).

PR #1704, opened 2019-10-17
Add __init__ parameters to class completion documentation

### LINKED ISSUE #1173: Add __init__ parameters to class completion documentation
I'd like to propose two features. 
First of all, could you please, add the class argument list at the top of the documentation pane when using intellisene? Like how it is done in Jedi? I mean have a look at this : 
Language server : 
intellisense_langaugesever

When using LanguageServer, when browsing/going through different class members /namespaces , the class argument list doesn't exist, there is only a class name which doesn't provide any kind of useful information! in order to see how a class is instantiated (what and how many arguments it has), you need to specifically navigate to that class! (i.e type the whole class name). 

While in Jedi this is done very neatly : 
intellisense_jedi

You can instantly see a class argument list while browsing/going through different members /namespaces and this makes development much easier . 

Apart from that, the coloring makes it way easier and interesting to read the documentations provided by the intellisense! the output in LangaugeServer is very crude in terms of presentation and is really bad compared  to what Jedi offers. So please do something about this as well! 

My other feature request is about how intellisense works, dont know if this should be asked here, but here it goes: 
It would be much better if pressing a `.` or `enter` or `space` on a half typed entity, selects the currently selected entity from the intellisense dropdown list. To make it more clear, consider the following image : 
image

Here I'm halfway in typing an entity name, if I press a `.` it will simply write `nn.adap` while it was expected to complete the half-typed name using the currently selected item from the dropdown list! 
Here's another example : 
image

Pressing `.` on this should choose the currently selected item from the dropdown list, i.e. Conv2d.

### LINKED ISSUE #1701: Scraper should output no init params for object, not *args/**kwargs
The scraper writes out:

```python
class object:
    'The most base type'
    __class__ = object

    def __init__(self, *args, **kwargs):
        'The most base type'
        pass
```

But you can't just do `object(1234)` and have that succeed. `object()` takes no arguments.

jakebailey:
Fixes #1173.
Fixes #1701.

Adds class `__init__` parameters to the completion documentation, if possible. Also, fixes the scraper to not output the wrong `__init__` function signature for `object`. And, remove the `class` keyword prefix, as it's inconsistent with hovers and other completions and ends up breaking syntax highlighting.

No DB version bump needed for the scraper change, as we do not store the builtins module which contains `object`.

### DIFF
--- a/src/LanguageServer/Test/HoverTests.cs
+++ b/src/LanguageServer/Test/HoverTests.cs
@@ -63,7 +63,7 @@ string = str
             hover.Should().BeNull();
 
             hover = hs.GetHover(analysis, new SourceLocation(4, 7));
-            hover.contents.value.Should().Be("C\n\nClass C is awesome");
+            hover.contents.value.Should().Be("class C\n\nClass C is awesome");
 
             hover = hs.GetHover(analysis, new SourceLocation(6, 9));
             hover.contents.value.Should().Be("C.method(a: int, b) -> float\n\nReturns a float!!!");
@@ -78,7 +78,7 @@ string = str
             hover.contents.value.Should().Be("y: int");
 
             hover = hs.GetHover(analysis, new SourceLocation(15, 2));
-            hover.contents.value.Should().StartWith("str\n\nstr(object='') -> str");
+            hover.contents.value.Should().StartWith("class str\n\nstr(object='') -> str");
         }
 
         [DataRow(false)]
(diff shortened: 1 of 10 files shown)


src/LanguageServer/Test/HoverTests.cs
```diff
@@ -93,7 +93,7 @@ import datetime
             var hs = new HoverSource(new PlainTextDocumentationSource());
 
             AssertHover(hs, analysis, new SourceLocation(3, 2), "module datetime*", new SourceSpan(3, 1, 3, 9));
-            AssertHover(hs, analysis, new SourceLocation(3, 11), "class datetime*", new SourceSpan(3, 9, 3, 18));
```
MikhailArkhipov: Something was wrong with the prefix? It helps to tell class vs module.

jakebailey: We were going to remove it wholesale for consistency, but I think I have a middle solution that would work slightly better.

PR #1686, opened 2019-10-14
Improve handling of circular dependencies

MikhailArkhipov:
Continuation of https://github.com/microsoft/python-language-server/pull/1584 by @AlexanderSher 
Requires https://github.com/microsoft/python-language-server/pull/1739

PR #1706, opened 2019-10-18
Define variables in from imports when module is unresolved

### LINKED ISSUE #955: Declare variables from unresolved imports as Unknown rather than not defining them
**Environment**
Visual Studio CodeVersion: 1.33.1
Python Extension Version: 2019.3.6558
OS: Windows 10
Python version: 2.7/3.6
Language Server Version: 0.2.47.0

**Actual Behavior**
Given the Python script
```python
import clr
from System import String

def foo():
    l = String("foo")
    return String("foo")

if __name__ == "__main__":
    print(foo())
```
The line `from System import String` reports the problem "unresolved import 'System' Python(unresolved-import) [2, 6]"
The line `return String("foo")` reports the problem "Undefined variable: 'String' Python(undefiend-variable) [6, 12]".

**Expected Behavior**
The "unresolved-import" warning is expected, but the "undefined-variable" is unexpected to me.
Interestingly, `l = String("foo")` is no problem (with 0.2.55.0 [beta], this line is also reported).

jakebailey:
Fixes #955.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/Handlers/FromImportHandler.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/Handlers/FromImportHandler.cs
@@ -40,21 +40,16 @@ namespace Microsoft.Python.Analysis.Analyzer.Handlers {
             }
 
             var imports = ModuleResolution.CurrentPathResolver.FindImports(Module.FilePath, node);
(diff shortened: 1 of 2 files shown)


jakebailey:
Will conflict with #1686.

PR #1713, opened 2019-10-22
Do not lock module database when working with database files

jakebailey:
For #1601.

These locks effectively do nothing.

- `ModuleExistsInStorage` would still be affected by the ordering of DB creation and checking regardless of the lock. Realistically, the file system is preventing internal race conditions, so we don't have to.
- `StoreModuleAnalysis` creates the LiteDB instances, but LiteDB has internal file locking mechanisms, so we don't need to repeat that here either.

`_lock` protects the dependency dictionary (as it's not a concurrent one), so can't just be removed.

### DIFF
--- a/src/Caching/Impl/ModuleDatabase.cs
+++ b/src/Caching/Impl/ModuleDatabase.cs
@@ -115,10 +115,8 @@ namespace Microsoft.Python.Analysis.Caching {
 
             for (var retries = 50; retries > 0; --retries) {
                 try {
-                    lock (_lock) {
-                        var dbPath = FindDatabaseFile(moduleName, filePath);
-                        return !string.IsNullOrEmpty(dbPath);
-                    }
+                    var dbPath = FindDatabaseFile(moduleName, filePath);
+                    return !string.IsNullOrEmpty(dbPath);
                 } catch (Exception ex) when (ex is IOException || ex is UnauthorizedAccessException) {
                     Thread.Sleep(10);
                 }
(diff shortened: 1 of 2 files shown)


src/Caching/Impl/ModuleDatabase.cs
```diff
@@ -127,7 +126,7 @@ public Task StoreModuleAnalysisAsync(IDocumentAnalysis analysis, CancellationTok
         }
 
         public void Clear() {
```
MikhailArkhipov: It is unused

src/Caching/Impl/ModuleDatabase.cs
```diff
@@ -33,7 +33,8 @@
 namespace Microsoft.Python.Analysis.Caching {
     internal sealed class ModuleDatabase : IModuleDatabaseService {
         private readonly Dictionary<string, IDependencyProvider> _dependencies = new Dictionary<string, IDependencyProvider>();
-        private readonly object _lock = new object();
```
MikhailArkhipov: Current write lock locks for all modules which is excessive. Should be possible to write modules in parallel. So it might be  better, but more complicated, to create write lock per module (via some protected dictionary). Single lock was for the early design with a single db for everything, which is not the case.

This way `FindModuleModel` may not need to lock the db. It should to check if the module is being written using the above, but otherwise it is just a file system read operation.

jakebailey: Actually, I think I can eliminate this lock entirely, as it appears that LiteDB has its own file locking system and handles multiple instances of the same file (https://github.com/mbdavid/LiteDB/wiki/Concurrency).

PR #1716, opened 2019-10-24
add unit test markup utilities

heejaechang:
this helper let one to annotate test text itself to mark locations.

see "add import" PR for example.

splitting this part from "add import" PR so that I can use it in "unused import" PR.

ex)
https://github.com/microsoft/python-language-server/pull/1656/files#diff-185d2e737eec3d6cf9c9812e2fc64bc3R50

https://github.com/microsoft/python-language-server/pull/1656/files#diff-8930788336d0cecf1c97200811f09869R115

### DIFF
--- a/src/UnitTests/Core/Impl/MarkupUtils.cs
+++ b/src/UnitTests/Core/Impl/MarkupUtils.cs
@@ -1,0 +1,244 @@
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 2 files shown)


heejaechang:
@MikhailArkhipov @jakebailey @bschnurr can you take a look?

src/UnitTests/Core/Impl/MarkupUtils.cs
```diff
@@ -0,0 +1,244 @@
...
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using Microsoft.Python.Core.Text;
+
+namespace Microsoft.Python.UnitTests.Core {
```
jakebailey: In which test assemblies is this going to be used? Does it make more sense for it to end up in `Microsoft.Python.LanguageServer.Tests` over the core unit test? If this is generically useful enough to not matter, then no big deal.

heejaechang: I think it is generally useful since you can annotate test code itself to mark stuff like where error should be associated or marking a variable to bind and see types and etc.

heejaechang:
thank you!

PR #1722, opened 2019-10-29
enable tiered compilation and use latest C# lang version

heejaechang:
and take advantage of ProfileOptimization (https://docs.microsoft.com/en-us/dotnet/api/system.runtime.profileoptimization?view=netcore-3.0) to make starting time faster 

some experiment user did on the web - https://www.dotnetperls.com/profileoptimization

...

confirmed that background JIT (multicore jit) and tiered compilation are enabled using perfview with background jit option

image

and confirmed things like parsing, analysis code is concurrently jitted from BG.
also, as vscode is used longer, I see more tiered compilation happens.

for now, I added profile optimization only on startup code path, but it is not actually limited to start-up scenario, one can put it any place where we want concurrent jit with different profile name such as "find all references" or "rename" and etc if there are bunch of code that needs to be jitted lazily.

### DIFF
--- a/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
@@ -20,7 +20,6 @@ using System.Linq;
 using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Python.Analysis.Caching;
 using Microsoft.Python.Analysis.Dependencies;
 using Microsoft.Python.Analysis.Diagnostics;
 using Microsoft.Python.Analysis.Documents;
@@ -36,7 +35,7 @@ using Microsoft.Python.Parsing.Ast;
 
 namespace Microsoft.Python.Analysis.Analyzer {
     public sealed class PythonAnalyzer : IPythonAnalyzer, IDisposable {
-        private readonly IServiceManager _services;
+        private readonly IServiceContainer _services;
         private readonly IDependencyResolver<AnalysisModuleKey, PythonAnalyzerEntry> _dependencyResolver;
         private readonly Dictionary<AnalysisModuleKey, PythonAnalyzerEntry> _analysisEntries = new Dictionary<AnalysisModuleKey, PythonAnalyzerEntry>();
         private readonly DisposeToken _disposeToken = DisposeToken.Create<PythonAnalyzer>();
@@ -51,7 +50,7 @@ namespace Microsoft.Python.Analysis.Analyzer {
         private PythonAnalyzerSession _nextSession;
         private bool _forceGCOnNextSession;
 
-        public PythonAnalyzer(IServiceManager services) {
+        public PythonAnalyzer(IServiceContainer services) {
             _services = services;
             _log = services.GetService<ILogger>();
             _dependencyResolver = new DependencyResolver<AnalysisModuleKey, PythonAnalyzerEntry>();
--- a/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
+++ b/src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Python.Analysis.Analyzer {
         private PythonAnalyzerSession _nextSession;
         private bool _forceGCOnNextSession;
 
-        public PythonAnalyzer(IServiceManager services, string cacheFolderPath = null) {
+        public PythonAnalyzer(IServiceManager services) {
             _services = services;
             _log = services.GetService<ILogger>();
             _dependencyResolver = new DependencyResolver<AnalysisModuleKey, PythonAnalyzerEntry>();
@@ -59,9 +59,6 @@ namespace Microsoft.Python.Analysis.Analyzer {
             _startNextSession = StartNextSession;
 
             _progress = new ProgressReporter(services.GetService<IProgressService>());
-
-            _services.AddService(new CacheFolderService(_services, cacheFolderPath));
-            _services.AddService(new StubCache(_services));
         }
 
         public void Dispose() {
--- a/src/Analysis/Ast/Impl/Caching/CacheFolders.cs
+++ b/src/Analysis/Ast/Impl/Caching/CacheFolders.cs
@@ -28,9 +28,7 @@ namespace Microsoft.Python.Analysis.Caching {
 
         public string CacheFolder { get; }
 
-        public string GetFileNameFromContent(string content) {
-            return content.GetHashString();
-        }
+        public string GetFileNameFromContent(string content) => content.GetHashString();
 
         private static string GetCacheFolder(IServiceContainer services) {
             var platform = services.GetService<IOSPlatform>();
--- a/src/Analysis/Ast/Impl/Caching/CacheFolders.cs
+++ b/src/Analysis/Ast/Impl/Caching/CacheFolders.cs
@@ -16,8 +16,6 @@
 using System;
 using System.Diagnostics;
 using System.IO;
-using System.Security.Cryptography;
-using System.Text;
 using Microsoft.Python.Core;
 using Microsoft.Python.Core.Logging;
 using Microsoft.Python.Core.OS;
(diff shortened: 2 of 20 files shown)


heejaechang:
tagging @MikhailArkhipov @jakebailey @bschnurr 
..

can I just move straight to 3.0?

MikhailArkhipov:
We are on .NET Core 3.0. Did you mean netstandard 2.1 vs netstandard 2.0?

heejaechang:
@MikhailArkhipov ah. ya. I see netcoreapp3.0 for LS.

src/LanguageServer/Impl/Program.cs
```diff
@@ -71,6 +75,19 @@ internal static class Program {
             }
         }
 
+        private static void EnableProfileOptimization() {
+            try {
+                // create directory for profile optimization
+                var path = Path.Combine(Path.GetTempPath(), "PythonLanguageServer", "Profiles");
```
heejaechang: another option is creating the profile folder along with .cache folder? but for now, I am creating it at temp directory

MikhailArkhipov: You get cache folder location on init, not at the startup so I am not sure if it can be used.

jakebailey: The cache folder location isn't dependent on any user information, just some specific set of rules that CacheFolder uses. It could be refactored out, since everything needed for it is created in Main.

Alternatively, would it be too harmful to just do this work on initialization, even if it's a little late and weird?

jakebailey:
That link you gave for the ProfileOptimization says that the program must exit normally in order for the profiles to actually be written out. I'm not entirely certain that's a common scenario for us, unfortunately.

src/LanguageServer/Impl/Program.cs
heejaechang: putting this profile in temp directory doesn't harm as well. so temp directory shouldn't be an issue as well. I can put user name in path for linux if we care.

heejaechang:
@jakebailey it is an opportunistic improvement, so it is fine even if the profile doesn't exist. dotnet will try to use it if it exists otherwise, same as before.

that being said, I tested and if vscode close by me clicking "close" or "x", then profile was saved as expected.

src/LanguageServer/Impl/Program.cs
```diff
@@ -71,6 +76,24 @@ internal static class Program {
             }
         }
 
+        private static void EnableProfileOptimization() {
+            try {
+                // create directory for profile optimization
+                var path = Path.Combine(Path.GetTempPath(), "PythonLanguageServer", GetUserName(), "Profiles");
```
jakebailey: I would be much happier if the caching directory code was moved to an IFileSystem method and used, rather than a tempdir or something custom.

heejaechang: alright. I will make it to use cache folder than.

jakebailey: For completeness, I mean to add something like `IFileSystem.GetCacheFolder`, move the code from `CacheFolderService` into `FileSystem` (sans logging?), then construct the FS earlier in Main to then use for profiling.

heejaechang: hmmm.. cache folder require initialized to return cache folder path so I am not sure whether that is possible, but I can move it up to Initialize at least.

jakebailey: Ah, I forgot about `cacheFolderPath`...

src/Analysis/Ast/Impl/Analyzer/PythonAnalyzer.cs
```diff
@@ -51,17 +51,14 @@ public sealed class PythonAnalyzer : IPythonAnalyzer, IDisposable {
         private PythonAnalyzerSession _nextSession;
         private bool _forceGCOnNextSession;
 
-        public PythonAnalyzer(IServiceManager services, string cacheFolderPath = null) {
+        public PythonAnalyzer(IServiceManager services) {
```
jakebailey: If this class isn't going to add to services, you can get away with lowering this to `IServiceContainer`.

src/Analysis/Ast/Impl/Caching/CacheFolders.cs
```diff
@@ -31,12 +29,7 @@ internal sealed class CacheFolderService : ICacheFolderService {
         public string CacheFolder { get; }
 
         public string GetFileNameFromContent(string content) {
-            // File name depends on the content so we can distinguish between different versions.
-            using (var hash = SHA256.Create()) {
-                return Convert
-                    .ToBase64String(hash.ComputeHash(new UTF8Encoding(encoderShouldEmitUTF8Identifier: false).GetBytes(content)))
-                    .Replace('/', '_').Replace('+', '-');
-            }
+            return content.GetHashString();
```
jakebailey: Nit; Arrow notation for one liners.

src/Analysis/Ast/Impl/Microsoft.Python.Analysis.csproj
```diff
@@ -9,7 +9,7 @@
       1701, 1702 - "You may need to supply assembly policy"
     -->
     <NoWarn>1701;1702;$(NoWarn)</NoWarn>
-    <LangVersion>7.2</LangVersion>
```
jakebailey: Does this language version need to stay? I think we had put it there for a reason, but I can't recall exactly.

src/LanguageServer/Impl/LanguageServer.Lifetime.cs
```diff
@@ -51,6 +57,24 @@ public partial class LanguageServer {
             }
         }
 
+        private void EnableProfileOptimization() {
+            var cachService = _services.GetService<ICacheFolderService>();
```
jakebailey: Nit: `cacheService`

src/Analysis/Ast/Impl/Caching/CacheService.cs
```diff
@@ -0,0 +1,46 @@
...
+            var fs = services.GetService<IFileSystem>();
+
+            // this is not thread safe. this is not supposed to be called concurrently
+            var cachingService = services.GetService<ICacheFolderService>();
+            if (cachingService != null) {
+                return;
+            }
+
+            if (cacheFolderPath != null && !fs.DirectoryExists(cacheFolderPath)) {
+                log?.Log(TraceEventType.Warning, Resources.Specified_cache_folder_0_does_not_exist_Switching_to_default.FormatUI(cacheFolderPath));
```
jakebailey: This resource name is a bit much (seems auto-genned), could it be revised?

jakebailey:
General note, but the PR title will need to be updated to reflect the actual changes (since we're not updating the standard).

src/Analysis/Ast/Impl/Microsoft.Python.Analysis.csproj
heejaechang: asked dotnet people and they said most of people just use latest C# lang version with .netstandard 2.0 and just make sure not to use some language features that rely on .net core 3.0 such as default interface method.

src/Analysis/Ast/Impl/Caching/CacheService.cs
heejaechang: simpler name?

src/Analysis/Ast/Impl/Caching/CacheService.cs
```diff
@@ -0,0 +1,46 @@
...
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System.Diagnostics;
+using Microsoft.Python.Core;
+using Microsoft.Python.Core.IO;
+using Microsoft.Python.Core.Logging;
+using Microsoft.Python.Core.Services;
+
+namespace Microsoft.Python.Analysis.Caching {
```
jakebailey: Thinking aloud, but I actually think this may be better served in another assembly. The `Caching` assembly is really all the "caching" that interacts with LiteDB, but not all caching in general. Perhaps all of the cache folder stuff could be moved away (a future PR, maybe).

src/Analysis/Ast/Impl/Caching/CacheService.cs
jakebailey: CacheFolderDoesNotExist maybe.

heejaechang: changed to Invalid_0_CacheFolder. still left _0_ since many time people forget to specify argument when using resource string.

src/Analysis/Ast/Impl/Caching/CacheService.cs
heejaechang: are you thinking putting caching stuff in Python.Core?

jakebailey: Maybe just into Analysis to start. The best would be my previous suggestion of having it on IFileSystem, but the API might get tricky when it comes to overrides. Not a priority.

heejaechang:
ping?

PR #1739, opened 2019-11-02
Fix null ref in super()

MikhailArkhipov:
Location must have non-null declaring module unless it is the module itself.


src/Caching/Impl/ModuleFactory.cs
```diff
@@ -205,7 +205,7 @@ private IMember GetMemberFromModule(IPythonModule module, IReadOnlyList<string>
                 case "Unknown":
                     return builtins.Interpreter.UnknownType;
                 case "SuperType":
-                    return new PythonSuperType(typeArgs);
+                    return new PythonSuperType(typeArgs, builtins);
```
MikhailArkhipov: Bit fishy, technically super should have proper declaring module...

src/Analysis/Ast/Impl/Types/PythonSuperType.cs
```diff
@@ -64,12 +67,8 @@ public PythonSuperType(IReadOnlyList<IPythonType> mro)
...
 
             var nextClassInLine = mro?.FirstOrDefault();
-            if (nextClassInLine != null) {
-                // Skip the first element, super's search starts at the next elemement in the mro for both super() and super(cls, typeToFind)
-                return new PythonSuperType(mro.Skip(1).ToArray());
-            }
-
-            return null;
+            // Skip the first element, super's search starts at the next element in the mro for both super() and super(cls, typeToFind)
+            return nextClassInLine != null ? new PythonSuperType(mro.Skip(1).ToArray(), classType.DeclaringModule) : null;
```
jakebailey: For consistency with the persistence model, maybe declare this in the builtins?

src/Caching/Impl/ModuleFactory.cs
jakebailey: Why can only modules have an unset declaring module?

MikhailArkhipov: Each type belong somewhere, i.e. where is the source, such as for navigation and accessing interpreter besides other things. Modules have null declaring module although technically submodules could have it or module could have itself there.

PR #1740, opened 2019-11-03
Don't lint "no-method-argument" until user types (

### LINKED ISSUE #1695: Non-existent function parameters should not lint as no-method-argument
image

I'm already getting a syntax error for not having finished writing the code; there's no value in telling me that I haven't specified any arguments when I haven't typed the `()` yet.

Likely that `CheckValidFunction` does `parameters.IsNullOrEmpty()`, but should return if `null` before checking the parameter count.

CTrando:
Fixes #1695 

Figured I should clean that up.

When there is an error in parameters, `FunctionDefinition.Body` will be null, parameters field is never null.


jakebailey:
Also a +1 from me, but since you're not in the org anymore you will have to appease the CLA bot.

CTrando:
Got it!

PR #1742, opened 2019-11-03
Fix fetching members from bases

### LINKED ISSUE #1741: lightgbm parameters intellisence Unsupport
image

The python language server fails to show the key parameters of lightgbm.

MikhailArkhipov:
Fixes #1741


src/Analysis/Ast/Impl/Types/PythonClassType.cs
```diff
@@ -101,12 +99,7 @@ internal PythonClassType(string name, Location location)
 
             using (_memberGuard.Push(this, out var reentered)) {
                 if (!reentered) {
-                    foreach (var m in Mro.Reverse()) {
-                        if (m == this) {
-                            return member;
-                        }
-                        member = member ?? m.GetMember(name);
-                    }
+                    return Mro.Skip(1).Select(c => c.GetMember(name)).ExcludeDefault().FirstOrDefault();
```
jakebailey: The old code would have traversed the MRO all the way to the current class, but the new code skips it. Is that a problem, or do we already get the value due to the above TryGetValue?

MikhailArkhipov: Old code was inherited, but it appeared odd to me. First, it was going in reverse, but only fetching first members (and yet continued up even that it never used newly found members). Effectively, this find method most close to the base (usually `__init__` from `object`). That seems wrong, why would while looking up method in derivation chain like `object -> A -> B -> C` it would prefer `__init__` from `object` and not from `B`?

`CalculateMro` already appears to be building linearization properly per https://en.wikipedia.org/wiki/C3_linearization so first item in MRO should be the class itself.

jakebailey: Yeah, I think the old code would have been correct if it had read `member = m.GetMember(name) ?? member`. I.e. "continuously pick the next thing until you reach `this` and return", which is no different than your LINQ and is just harder to read.

jakebailey:
Can you merge master so I can test this with some `numpy` stuff?

PR #1765, opened 2019-11-07
Specialize TypeVar, NamedTupe, Type and type as classes

### LINKED ISSUE #1367: Improve specialization of Typing Module by specializing classes as classes instead of functions
For example, `NamedTuple` in the typing module is specialized as a function when it is actually a class, which throws off the `InheritNonClass` check.

### LINKED ISSUE #1764: False positive: inherit-non-class on subclass of metaclass-constructed class
## Environment data

- Language Server version: 0.4.71.0 
- Python version (& distribution if applicable, e.g. Anaconda): Python 3.7

## Expected behaviour

No errors.

## Actual behaviour

PLS reports `inherit-non-class` in the declaration of Subclass. See code snippet below.

```python
class MyMetaclass(type):
    pass

MyClass = MyMetaclass('MyClass', (), {})

class Subclass(MyClass):
    pass
```

MikhailArkhipov:
Fixes #1764 
Fixes #1367


src/Analysis/Ast/Impl/Specializations/Typing/Types/GenericTypeParameter.cs
```diff
@@ -71,7 +71,7 @@ internal sealed class GenericTypeParameter : PythonType, IGenericTypeParameter {
             }
 
             // Report diagnostic if user passed in a value for name and it is not a string
-            var name = (args[0].Value as IPythonConstant)?.GetString();
+            var name = (args[1].Value as IPythonConstant)?.GetString();
```
jakebailey: Is this because `arg[0]` is now `self`?

PR #1767, opened 2019-11-07
Initial support added for textDocument/documentHighlight

astenman:
Change-Id: Ib13d86bc96a3702b0e0d79b27b7791898388e104

Since I am using this language server in emacs, it would be nice to have support for textDocument/documentHighlight. 

The added functionality only looks for related items in the current file.


astenman:
Good point. I have added a test file with two simple tests.

src/LanguageServer/Impl/Sources/DocumentHighlightSource.cs
```diff
@@ -0,0 +1,65 @@
...
+    internal sealed class DocumentHighlightSource {
+        private const int DocumentHighlightAnalysisTimeout = 10000;
+        private readonly IServiceContainer _services;
+
+        public DocumentHighlightSource(IServiceContainer services) {
+            _services = services;
+        }
+
+        public async Task<DocumentHighlight[]> DocumentHighlightAsync(Uri uri, SourceLocation location, CancellationToken cancellationToken = default) {
+            if (uri != null) {
```
jakebailey: Style nit: I'd just do `if (uri == null)` and then return an empty array, rather than having the entire function be a single indented block.

I think the file you based it one does it, but it should probably be changed too (in another PR).

src/LanguageServer/Impl/Sources/DocumentHighlightSource.cs
```diff
@@ -0,0 +1,65 @@
...
+                var analysis = await Document.GetAnalysisAsync(uri, _services, DocumentHighlightAnalysisTimeout, cancellationToken);
+
+                var definitionSource = new DefinitionSource(_services);
+
+                var definition = definitionSource.FindDefinition(analysis, location, out var definingMember);
+                if ((definition == null) || (definingMember == null)) {
```
jakebailey: You don't need these extra parens.

src/LanguageServer/Impl/Sources/DocumentHighlightSource.cs
```diff
@@ -0,0 +1,65 @@
...
+        private const int DocumentHighlightAnalysisTimeout = 10000;
+        private readonly IServiceContainer _services;
+
+        public DocumentHighlightSource(IServiceContainer services) {
+            _services = services;
+        }
+
+        public async Task<DocumentHighlight[]> DocumentHighlightAsync(Uri uri, SourceLocation location, CancellationToken cancellationToken = default) {
+            if (uri != null) {
+                var analysis = await Document.GetAnalysisAsync(uri, _services, DocumentHighlightAnalysisTimeout, cancellationToken);
```
jakebailey: One thing we might want to think about is returning `null` if there is no analysis yet, to allow the editor to fall back to lazy. Right now, this call will end up returning an empty list and VS Code doesn't highlight anything.

jakebailey: This might be a blocking thing, actually, since the experience regresses until the analysis is complete. I _think_ it would be as simple as `if (analysis is EmptyAnalysis) { return null }` but @MikhailArkhipov would know best how to check that things aren't ready.

src/LanguageServer/Impl/Sources/DocumentHighlightSource.cs
astenman: This was a copy/paste issue. I have refactored the code now.

src/LanguageServer/Impl/Sources/DocumentHighlightSource.cs
astenman: Fixed

src/LanguageServer/Impl/Sources/DocumentHighlightSource.cs
astenman: This is beyond my knowledge for the moment. However when I'm testing this in Emacs, I get highlighting while the server is analyzing.

astenman: It would also be nice if the analysis could provide information whether the symbol is read or write, but this is just a "nice to have" feature.

MikhailArkhipov: Previous analysis may or may not be empty, if timeout expires waiting on analysis, previous one will be returned. One possible way to determine if analysis is stale is to check `Document.Version` against `Analysis.Version`

MikhailArkhipov: As for symbol R/W do you mean it can be renamed/changed? There is a root definition of the symbol that `Rename` operation uses which has URI of the defining module. Via `IRunningDocumentsTable` aka `RDT` one can fetch the document and check if `Document.ModuleType == ModuleType.User` since symbols in other module types are read-only.

jakebailey: R/W meaning "this location reads from `x`" versus "this location writes to `x`".

jakebailey: `Document.Version` and `Analysis.Version` are incremented separately, so comparing them wouldn't work. We would need to convert entirely to the snapshot model with a single incrementing version across everything for that to work (which would be awesome, along with using LSP's versioned identifiers).

I just re-tested, and maybe it doesn't break and we can just leave the PR as-is. I'm still testing to make sure.

MikhailArkhipov: 👍

astenman: I meant as Jake interpreted it ;-). DocumentHighlight can tag a symbol instance as "Read", "Write" or "Text". In the following simple example "x" is redefined in the last line which means that it could be tagged as "Read". Emacs LSP renders "Read" symbols in italics. But this is not a showstopper for me.

lsp

jakebailey: @astenman Perhaps I'm confused, but doesn't your PR include that information? Or are you using Read/Write a different way?

astenman:
I am using the fact (I think it is a fact though) that the first element of
the References list can be interpreted as the actual definition and thus be
tagged as "write".

Den ons 13 nov. 2019 21:08Jake Bailey <notifications@github.com> skrev:

jakebailey:
Ah, I understand. This PR looks good to me, though I'm going to want to hold off on merging it immediately (extension release soon, so we're trying to keep things to bugfixes to limit risky changes).

jakebailey:
Will merge once I have done a little bit of testing.

astenman:
Thanks!

PR #1777, opened 2019-11-08
Log parse exceptions instead of just throwing them away

jakebailey:
Debugging for #1776.

Doesn't fix the core issue that we're spawning tasks we can't keep track of, but helps trying to find the crashes.

### DIFF
--- a/src/Analysis/Ast/Impl/Modules/PythonModule.cs
+++ b/src/Analysis/Ast/Impl/Modules/PythonModule.cs
@@ -313,7 +313,16 @@ namespace Microsoft.Python.Analysis.Modules {
             _linkedParseCts = CancellationTokenSource.CreateLinkedTokenSource(_disposeToken.CancellationToken, _parseCts.Token);
 
             ContentState = State.Parsing;
-            _parsingTask = Task.Run(() => Parse(_linkedParseCts.Token), _linkedParseCts.Token);
+            _parsingTask = Task.Run(() => ParseAndLogExceptions(_linkedParseCts.Token), _linkedParseCts.Token);
+        }
+
+        private void ParseAndLogExceptions(CancellationToken cancellationToken) {
+            try {
+                Parse(cancellationToken);
+            } catch (Exception ex) when (!(ex is OperationCanceledException)) {
+                Log?.Log(TraceEventType.Warning, $"Exception while parsing {FilePath}: {ex}");
+                throw;
+            }
         }
 
         private void Parse(CancellationToken cancellationToken) {


src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -313,7 +313,16 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
...
 
             ContentState = State.Parsing;
-            _parsingTask = Task.Run(() => Parse(_linkedParseCts.Token), _linkedParseCts.Token);
+            _parsingTask = Task.Run(() => ParseAndLogExceptions(_linkedParseCts.Token), _linkedParseCts.Token);
+        }
+
+        private void ParseAndLogExceptions(CancellationToken cancellationToken) {
+            try {
+                Parse(cancellationToken);
+            } catch (Exception ex) when (!(ex is OperationCanceledException)) {
```
jakebailey: Other option is to `catch (OperationCanceledException) { throw; } catch (Exception ex) { ... }`, but I'm not a .NET style expert.

src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -313,7 +313,16 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
             _linkedParseCts = CancellationTokenSource.CreateLinkedTokenSource(_disposeToken.CancellationToken, _parseCts.Token);
 
             ContentState = State.Parsing;
-            _parsingTask = Task.Run(() => Parse(_linkedParseCts.Token), _linkedParseCts.Token);
+            _parsingTask = Task.Run(() => ParseAndLogExceptions(_linkedParseCts.Token), _linkedParseCts.Token);
+        }
+
+        private void ParseAndLogExceptions(CancellationToken cancellationToken) {
+            try {
```
heejaechang: another approach will be having something like this

```csharp
ErrorHandler.LogIfNotCancelled(Exception ex, ILogger logger, string message) {
	if (ex is Operation...) return false;
	logger?.Log($"exception : {ex}, message: {message}");
    return false;
}

try
{
}
catch(Exception ex) where (ErrorHandler.LogIfNotCancelled(ex, logger, filePath)) { }

so every one could use same helper.

jakebailey: I'll pull that code out and see how it looks. Maybe just locally and we can pull it out all the way later.

jakebailey: I'm going to skip it for now, but we can pull it out later if needed. I'm not seeing other places in the current code where it would help.

Issue #1697, opened 2019-10-16
print, others flagged as "Undefined variable" (requests being handled before initialization is complete?)


JoeCodeswell:
## Environment data

- Language Server version:  0.4.71.0 
- OS and version: Microsoft Windows [Version 10.0.18362.418]
- Python version (& distribution if applicable, e.g. Anaconda): Python 3.5.1 (v3.5.1:37a07cee5969, Dec  6 2015, 01:54:25) [MSC v.1900 64 bit (AMD64)] on win32

## Expected behaviour

print should NOT be flagged as "Undefined variable"

## Actual behaviour
print is flagged as "Undefined variable"

## Logs

```
N/A
```

## Code Snippet / Additional lnformation

![vscodeUndefinedVar_print](https://user-images.githubusercontent.com/3194021/66938039-509cf380-eff5-11e9-8e09-2511417c5fb2.png)

```python
Python 3.5.1 (v3.5.1:37a07cee5969, Dec  6 2015, 01:54:25) [MSC v.1900 64 bit (AMD64)] on win32
```

jakebailey:
Could you please provide the logs asked for in the template? I'd like to see the caching level you are using.

JoeCodeswell:
Hi Jake. Thanks for the response. I think you want me to follow:


So I did Python › Analysis: Log Level
Defines type of log messages language server writes into the output window. 
Trace

And Here's Output.Python:

 ```

    User belongs to experiment group 'AlwaysDisplayTestExplorer - control'
    User belongs to experiment group 'ShowPlayIcon - start'
    User belongs to experiment group 'ShowExtensionSurveyPrompt - control'
    User belongs to experiment group 'AA_testing - experiment'
    > conda --version
    > pyenv root
    > python3.7 -c "import sys;print(sys.executable)"
    > python3.6 -c "import sys;print(sys.executable)"
    > python3 -c "import sys;print(sys.executable)"
    > python2 -c "import sys;print(sys.executable)"
    > python -c "import sys;print(sys.executable)"
    > py -3.7 -c "import sys;print(sys.executable)"
    > py -3.6 -c "import sys;print(sys.executable)"
    > py -3 -c "import sys;print(sys.executable)"
    > py -2 -c "import sys;print(sys.executable)"
    > python -c "import sys;print(sys.executable)"
    > conda info --json
    Starting Microsoft Python language server.
[...]

PR #1778, opened 2019-11-08
Do not allow the analyzer to be reset if the server is still initializing

jakebailey:
For #1697, #1776.

### DIFF
--- a/src/LanguageServer/Impl/Implementation/Server.cs
+++ b/src/LanguageServer/Impl/Implementation/Server.cs
@@ -50,6 +50,7 @@ namespace Microsoft.Python.LanguageServer.Implementation {
         private IIndexManager _indexManager;
 
         private InitializeParams _initParams;
+        private bool _initialized;
 
         private bool _watchSearchPaths;
         private PathsWatcher _pathsWatcher;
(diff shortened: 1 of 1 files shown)


src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -50,6 +50,7 @@ public sealed partial class Server : IDisposable {
         private IIndexManager _indexManager;
 
         private InitializeParams _initParams;
+        private bool _initialized;
```
jakebailey: Technically, this is racy. I can replace it with something else, but I'm not sure how worthwhile it is for a best-effort bool.

bschnurr: shouldn't we default to false?

jakebailey: The default boolean is false already.

src/LanguageServer/Impl/Implementation/Server.cs
```diff
@@ -249,6 +252,12 @@ public sealed partial class Server : IDisposable {
         }
 
         private void ResetAnalyzer() {
+            if (!_initialized) {
```
heejaechang: another approach could be making this to be async method and await until initialized to happen and then proceed. but I saw caller of this thing and how things connected so might be too much work for a corner case.

basically, it looks like this can happen if files in path server watching (such as lib directory and etc) changed before initialization finished. and it those case, ignoring those changes seems fine to me.

jakebailey: Yeah, we don't have to go overkill because if it's still being initialized, then any future reads of library files will necessarily be _after_ them being changed, so we'll get the updated info anyway. Reloading is more of a "everything you have before now is wrong, reset", but before init this is meaningless.

PR #1790, opened 2019-11-12
Look for module of the definition in RDT during rename

### LINKED ISSUE #979: Library function is renamable

MikhailArkhipov:
Fixes #979 

Variables do not have declaring module so we need to look in RDT, like goto definition does.

Includes https://github.com/microsoft/python-language-server/pull/1788 by accident so it needs to go last. Actual change is one file at the end.


jakebailey:
#1788 is merged, so this needs a master merge to fix the diff.

src/LanguageServer/Impl/Sources/ReferenceSource.cs
```diff
@@ -54,13 +54,19 @@ internal sealed class ReferenceSource {
...
 
                 var rootDefinition = definingMember.GetRootDefinition();
-                var name = definingMember.GetName();
-
-                // If it is an implicitly declared variable, such as function or a class
-                // then the location is invalid and the module is null. Use current module.
-                var declaringModule = rootDefinition?.DeclaringModule ?? analysis.Document;
-                if (!string.IsNullOrEmpty(name) && (declaringModule.ModuleType == ModuleType.User || options == ReferenceSearchOptions.All)) {
-                    return await FindAllReferencesAsync(name, declaringModule, rootDefinition, location, definitionSource, cancellationToken);
+                if (rootDefinition != null) {
```
jakebailey: This is an overall style nit, but this function looks pretty nested even though it's just a few null checks deep that could probably all return early. `url == null`, `definition == null`, `moduleUri == null`, etc.

jakebailey:
Is there any way to test this fix?

MikhailArkhipov:
Probably yes, need to add file to what we have in tests so it is considered a library. Let me try that.

PR #1792, opened 2019-11-13
Specialize enumerate()

### LINKED ISSUE #1235: enumerate() is not specialized
image

We should know the tuple type the enumerate generator returns.

MikhailArkhipov:
Fixes #1235


src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -175,5 +175,17 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
             return o?.GetPythonType().GetMember(name) ?? def;
         }
+
+        public static IMember Enumerate(IPythonModule module, IPythonFunctionOverload overload, IArgumentSet argSet, IndexSpan indexSpan) {
+            var args = argSet.Values<IMember>();
+            if (args.Count > 0) {
+                var iterator = (args[0] as IPythonIterable)?.GetIterator();
+                var itemType = (iterator?.Next ?? module.Interpreter.UnknownType).GetPythonType();
+                var intType = module.Interpreter.GetBuiltinType(BuiltinTypeId.Int);
+                var content = new[] {new PythonInstance(intType), itemType.CreateInstance(ArgumentSet.WithoutContext)};
```
jakebailey: `intType.CreateInstance(ArgumentSet.WithoutContext)` maybe? I think we've been avoiding doing `new PythonInstance` directly in favor of always calling `CreateInstance`.

MikhailArkhipov: Sure.

src/Analysis/Ast/Impl/Specializations/BuiltinsSpecializations.cs
```diff
@@ -175,5 +175,17 @@ public static IMember List(IPythonInterpreter interpreter, IPythonFunctionOverlo
...
         }
+
+        public static IMember Enumerate(IPythonModule module, IPythonFunctionOverload overload, IArgumentSet argSet, IndexSpan indexSpan) {
+            var args = argSet.Values<IMember>();
+            if (args.Count > 0) {
+                var iterator = (args[0] as IPythonIterable)?.GetIterator();
+                var itemType = (iterator?.Next ?? module.Interpreter.UnknownType).GetPythonType();
+                var intType = module.Interpreter.GetBuiltinType(BuiltinTypeId.Int);
+                var content = new[] { intType.CreateInstance(ArgumentSet.WithoutContext), itemType.CreateInstance(ArgumentSet.WithoutContext) };
+                return PythonCollectionType.CreateTuple(module, content);
```
jakebailey: This isn't correct. enumerate needs to be returning a iterator of tuples, not just a single tuple.

I.e. currently `x = enumerate(foo)` makes `x` a tuple, but it shouldn't be.

Typeshed actually types it as a class, interestingly: https://github.com/python/typeshed/blob/ba520034c7e84f27d3eeab6ac92a936470132ffb/stdlib/2and3/builtins.pyi#L1066

MikhailArkhipov: I guess we can try class as well since Python returns `<enumerate object at ...>` from enumerate(). so `x = enumerate(foo)` would then make `x` a class, but since it is specialized, we can mimic `enumerate object` in the class documentation.

jakebailey: Maybe, the other problem is that the python docs say that `enumerate` is a function, not a class, so I'd be worried about showing it as a class rather than a function too.

MikhailArkhipov: I think we may have to create `Enumerator` class like .NET does with `IEnumerator`. Then `enumerate()` would return `Enumerator` and assignment handler can handle it depending on the LHS.

jakebailey: Yes, that's typing.Iterable (maybe Sequence?), combined with having it produce `Tuple[int, _T]` like in Typeshed.

Realistically just those who have `__next__` is enough to specialize iterators?

MikhailArkhipov: Implemented special `PythonEnumerator`

src/Analysis/Ast/Impl/Analyzer/Handlers/AssignmentHandler.cs
```diff
@@ -99,6 +100,10 @@ internal sealed class AssignmentHandler : StatementHandler {
         private bool IsValidAssignment(string name, Location loc) => !Eval.GetInScope(name).IsDeclaredAfter(loc);
 
         private void HandleNameExpression(NameExpression ne, IMember value) {
+            if (value is IPythonEnumerator e) {
+                value = e.Next; // x = enumerate(y)
```
jakebailey: This isn't correct either; this needs to set `x` to the value that is the enumerator, not `next()` it.

MikhailArkhipov: Yeah, made it more common with iterator, so `__next__` would also work.

PR #1796, opened 2019-11-14
Bschnurr/Microsoft.Python.Parsing nuget publish

bschnurr:
- Bundling Microsoft.Python.Parsing into a nuget package for PTVS
- Removing the old Anaylsis nuget bundle
- Updating copywrite info to pass nuget push command verification
https://aka.ms/Microsoft-NuGet-Compliance
- Updating signing to sign Microsoft.Python.Parsing package instead of analysis

### DIFF
--- a/src/Parsing/Impl/Microsoft.Python.Parsing.csproj
+++ b/src/Parsing/Impl/Microsoft.Python.Parsing.csproj
@@ -35,6 +35,11 @@
       <LastGenOutput>Resources.Designer.cs</LastGenOutput>
     </EmbeddedResource>
   </ItemGroup>
+  <ItemGroup>
+    <None Update="Microsoft-Python-Parsing.nuspec">
+      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
+    </None>
+  </ItemGroup>
   <Import Project="Sdk.targets" Sdk="Microsoft.NET.Sdk" />
   <Import Project="..\..\..\build\NetStandard.targets" />
 </Project>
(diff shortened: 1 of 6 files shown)


src/Parsing/Impl/Microsoft.Python.Parsing.csproj
```diff
@@ -35,6 +35,11 @@
       <LastGenOutput>Resources.Designer.cs</LastGenOutput>
     </EmbeddedResource>
   </ItemGroup>
+  <ItemGroup>
+    <None Update="Microsoft-Python-Parsing.nuspec">
```
heejaechang: Microsoft.Python.Parsing.nuspec ?

MikhailArkhipov: I think nuget does not tolerate dots.

heejaechang:
so replacing analysis to parsing rather than just adding parsing nuget pkg?

MikhailArkhipov:
Engine package included parser since it is a dependency. I guess analysis is no longer needed since VS works via LSP.

src/Parsing/Impl/Microsoft.Python.Parsing.csproj
bschnurr: I just keep the nuspec name the same style as before.. but the nuget id is using ```.```

bschnurr:
> Engine package included parser since it is a dependency. I guess analysis is no longer needed since VS works via LSP.

Correct.. @Huges has duplicated the one or two files he needed from analysis

bschnurr:
> so replacing analysis to parsing rather than just adding parsing nuget pkg?

Yes. PTVS was the only user of analysis but in reality they just need parsing.

PR #1831, opened 2019-12-10
Handle goto definition of unknown types better

### LINKED ISSUE #1563: Navigation to compiled/scraped stubs should be suppressed
Attempt to navigate to `SimpleImputer` in the import
```python
from sklearn.preprocessing  import SimpleImputer
x = SimpleImputer()
```
yields
image

We should suppress navigation to scraped modules better.

MikhailArkhipov:
Fixes #1563


src/LanguageServer/Impl/Sources/DefinitionSource.cs
```diff
@@ -262,7 +264,8 @@ internal abstract class DefinitionSourceBase {
             definingMember = null;
 
             var m = analysis.ExpressionEvaluator.LookupNameInScopes(name, out var scope, LookupOptions.All);
-            if (m == null || scope.Module.ModuleType == ModuleType.Builtins || !(scope.Variables[name] is IVariable v)) {
+            var v = scope?.Variables[name];
+            if (m == null || scope == null || scope.Module.ModuleType == ModuleType.Builtins || scope.Variables[name].IsUnknown()) {
```
jakebailey: Should the lats check have been on `v`? I don't see `v` being used.

jakebailey: Ah, the assignment below was hidden in the diff. You could probably get away with `v.IsUnknown()`, though.

PR #1838, opened 2019-12-13
Escape [ ] in markdown

### LINKED ISSUE #1804: Language server hangs up and stops providing suggestions after typing in a certain module from the discord.py library
## Environment data

- Language Server version: 0.4.114.0 
- OS and version: Arch Linux, kernel version 5.3.8-zen1-1-zen
- Python version (& distribution if applicable, e.g. Anaconda): 3.7.4

## Expected behaviour

The suggestions continue to work regardless of what is typed into the editor.

## Actual behaviour

The suggestions work up until the moment the `@commands` annotation is typed, and they stop working when the dot is entered.

## Logs

Posted to [pastebin](https://pastebin.com/xE5E3TpK) as they're quite long.

[Video demonstrating the issue](https://streamable.com/oq4ec).

As you can see here, the suggestions work up until the moment the `@commands` annotation is typed, and they stop working when the dot is entered.

* python file:
```python
from discord.ext import commands

@commands.command
async def test(ctx):
    pass
```

* `Pipfile`:
```toml
[[source]]
name = "pypi"
url = "https://pypi.org/simple"
verify_ssl = true

[dev-packages]

[packages]
discord-py = "*"

[requires]
python_version = "3.7"
```

* `settings.json`:
```json
{
    "python.jediEnabled": false,
    "python.analysis.logLevel": "Trace"
}
```

MikhailArkhipov:
Fixes #1804


src/Analysis/Ast/Impl/Analyzer/Symbols/FunctionEvaluator.cs
```diff
@@ -38,10 +38,9 @@ internal sealed class FunctionEvaluator : MemberEvaluator {
         public FunctionEvaluator(ExpressionEval eval, PythonFunctionOverload overload)
             : base(eval, overload.FunctionDefinition) {
             _overload = overload;
-            _function = overload.ClassMember ?? throw new NullReferenceException(nameof(overload.ClassMember));
+            _function = overload.ClassMember ?? throw new ArgumentNullException(nameof(overload));
```
jakebailey: If the exception is referring to the overload, should the null check be done above? What are these guarding against? I'm also noting that FunctionDefinition gets passed to the base before this code would run, too.

PR #1863, opened 2020-01-07
Fix matmul operator parsing

### LINKED ISSUE #1825: Failed parsing matmul operator
## Environment data

- Language Server version: 0.4.127.0
- OS and version: macOS 10.14.6
- Python version (& distribution if applicable, e.g. Anaconda): 3.7.0

## Expected behaviour

Parser should be able to parse this code:

```python
a = (
    A
    @ B
)
```

## Actual behaviour

It causes parser error `unexpected token '<newline>'` at end of 2nd line.

Note that this happened only with `@` operator. (No parse error with `+` or `*` instead of `@`.)

MikhailArkhipov:
Fixes #1825


src/Parsing/Test/ParserTests.cs
```diff
@@ -4445,6 +4463,8 @@ public class ParserTests {
             };
         }
 
+
```
jakebailey: Nit, but weird whitespace added here.

src/Parsing/Impl/Parser.cs
```diff
@@ -2986,7 +2986,6 @@ struct Name {
                 if (!(t is OperatorToken ot)) {
                     return ret;
                 }
-
```
jakebailey: Another nit, but the only change to this file is whitespace.

MikhailArkhipov:
0.5.30+

PR #1890, opened 2020-02-03
Porting PTVS LSClient

bschnurr:
Initial changes around integration LS testing.
I needed a new test project for mock language server client code.

- currently left code around setting environment variables for choosing which type LS to startup
- added code to kill the server process

### DIFF
--- a/src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/Mocks/MockLanguageClientBroker.cs
+++ b/src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/Mocks/MockLanguageClientBroker.cs
@@ -1,0 +1,71 @@
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.LanguageServer.Client;
+using Microsoft.VisualStudio.LanguageServer.Protocol;
+using StreamJsonRpc;
+
+namespace UnitTests.LanguageServerClient.Mocks {
+    /// <summary>
+    /// Minimal version of the broker used by VS
+    /// (which is internal and so we can't instantiate).
+    /// Note that this requires StreamJsonRpc to run, which in turn depends on
+    /// a lot of other dlls that are normally available in VS install, but that
+    /// need to be referenced by the test project in order to run outside VS.
+    /// </summary>
+    public class MockLanguageClientBroker : ILanguageClientBroker {
+        private JsonRpc _rpc;
+
+        public async Task LoadAsync(ILanguageClientMetadata metadata, ILanguageClient client) {
+            if (client == null) {
+                throw new ArgumentNullException(nameof(client));
+            }
+
+            client.StartAsync += Client_StartAsync;
+            await client.OnLoadedAsync();
+        }
+
+        private async Task Client_StartAsync(object sender, EventArgs args) {
+            var client = (ILanguageClient)sender;
+            var connection = await client.ActivateAsync(new CancellationTokenSource(TimeSpan.FromSeconds(15)).Token);
+            await InitializeAsync(client, connection);
+        }
+
+        private async Task InitializeAsync(ILanguageClient client, Connection connection) {
+            var messageHandler = new HeaderDelimitedMessageHandler(connection.Writer, connection.Reader);
+            _rpc = new JsonRpc(messageHandler, this) {
+                CancelLocallyInvokedMethodsWhenConnectionIsClosed = true
+            };
+
+            _rpc.StartListening();
+
+            (client as ILanguageClientCustomMessage2)?.AttachForCustomMessageAsync(_rpc);
+
+            var initParam = new InitializeParams {
+                ProcessId = Process.GetCurrentProcess().Id,
+                InitializationOptions = client.InitializationOptions
+            };
+
+            await _rpc.InvokeWithParameterObjectAsync(Methods.Initialize.Name, initParam, cancellationToken: CancellationToken.None);
+            await _rpc.NotifyWithParameterObjectAsync(Methods.Initialized.Name, new InitializedParams());
+            await client.OnServerInitializedAsync();
+        }
+    }
+}
--- a/src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/PythonLanguageClient.cs
+++ b/src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/PythonLanguageClient.cs
@@ -1,0 +1,273 @@
+﻿// Python Tools for Visual Studio
+// Copyright(c) Microsoft Corporation
+// All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the License); you may not use
+// this file except in compliance with the License. You may obtain a copy of the
+// License at http://www.apache.org/licenses/LICENSE-2.0
+//
+// THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR CONDITIONS
+// OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT LIMITATION ANY
+// IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS FOR A PARTICULAR PURPOSE,
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Python.Core.Disposables;
+using Microsoft.Python.Parsing;
+using Microsoft.VisualStudio.LanguageServer.Client;
+using Microsoft.VisualStudio.Threading;
+using StreamJsonRpc;
+using LSP = Microsoft.VisualStudio.LanguageServer.Protocol;
+using Task = System.Threading.Tasks.Task;
+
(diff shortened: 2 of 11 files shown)


src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/UnitTests.LanguageServerClient.csproj
```diff
@@ -0,0 +1,38 @@
...
+    </PackageReference>
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.VisualStudio.CoreUtility" Version="16.4.280" />
+    <PackageReference Include="Microsoft.VisualStudio.LanguageServer.Client" Version="16.4.30" />
+    <PackageReference Include="Microsoft.VisualStudio.LanguageServer.Protocol" Version="16.4.30" />
+    <PackageReference Include="StreamJsonRpc" Version="2.2.53" />
+  </ItemGroup>
+  <ItemGroup>
+    <ProjectReference Include="..\..\..\Analysis\Core\Impl\Microsoft.Python.Analysis.Core.csproj" />
```
bschnurr: needed for "InterpreterConfiguration"

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/UnitTests.LanguageServerClient.csproj
```diff
@@ -0,0 +1,38 @@
...
+    <ProjectReference Include="..\..\..\Parsing\Impl\Microsoft.Python.Parsing.csproj" />
```
bschnurr: needed for "PythonLanguageVersion"

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/UnitTests.LanguageServerClient.csproj
```diff
@@ -0,0 +1,38 @@
...
+  <Import Project="..\..\..\..\build\NetStandard.settings" />
+  <Import Project="Sdk.props" Sdk="Microsoft.NET.Sdk" />
+  <ItemGroup>
+    <PackageReference Include="Ben.Demystifier" Version="0.1.4" />
+    <PackageReference Include="FluentAssertions" Version="5.9.0" />
+    <PackageReference Include="MSTest.TestFramework" Version="2.0.0" />
+    <PackageReference Include="MicroBuild.Core" Version="0.3.0">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
+    </PackageReference>
```
bschnurr: just copied these base references for UnitTest.Core project

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/UnitTests.LanguageServerClient.csproj
```diff
@@ -0,0 +1,38 @@
+﻿<Project>
+  <PropertyGroup>
+    <TargetFramework>netstandard2.1</TargetFramework>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <NoWarn>NU1701</NoWarn>
```
bschnurr: Special disabled warning about Microsoft.VisualStudio.LanguageServer.Client being built on a different framework. but just using interfaces from Microsoft.VisualStudio.LanguageServer.Client so it should be an issue.

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/PythonLanguageServerDotNetCore.cs
```diff
@@ -0,0 +1,244 @@
...
+            if (_process != null) {
+                try {
+                    _process.Kill();
+                } catch (InvalidOperationException) { } catch (NotSupportedException) { }
+
+                _process.WaitForExit();
+                _process.Dispose();
+                _process = null;
+            }
+        }
```
bschnurr: added the Dispose() function above to kill the LS process and catch any exceptions.

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/PythonLanguageServerDotNetCore.cs
```diff
@@ -0,0 +1,244 @@
...
+                throw new InvalidOperationException("Must be called from main thread.");
+            }
+
+            //var shell = _site.GetService<SVsShell, IVsShell>();
+            //shell.GetProperty((int)__VSSPROPID4.VSSPROPID_LocalAppDataDir, out object localAppDataFolderPath);
+
+            //lsExtractedFolderPath = Path.Combine((string)localAppDataFolderPath, "PythonLanguageServer");
+            //if (!Directory.Exists(lsExtractedFolderPath)) {
+            //    ExtractToFolder(lsExtractedFolderPath);
+            //}
```
bschnurr: commented out code related to VS.. currently I'm not extracting the server from a zip folder. I'm just using the build output folder in MPLS

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/Mocks/MockLanguageClientBroker.cs
```diff
@@ -0,0 +1,71 @@
...
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System;
+using System.Diagnostics;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.LanguageServer.Client;
```
heejaechang: code doesn't look like it is using any implementation but it is using interface defined in these namespace? but also, our implementatino doesn't plug into VS as well.

so, if we don't want this VS dependency, we can define the interface our own? rather than reuse the interfaces causing some confusion?

heejaechang: ah, it is using the library code such as Connection and MessageHead

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -0,0 +1,273 @@
...
+
+namespace UnitTests.LanguageServerClient {
+    /// <summary>
+    /// Implementation of the language server client.
+    /// </summary>
+    /// <remarks>
+    /// See documentation at https://docs.microsoft.com/en-us/visualstudio/extensibility/adding-an-lsp-extension?view=vs-2019
+    /// </remarks>
+    public class PythonLanguageClient : ILanguageClient, ILanguageClientCustomMessage2, IDisposable {
+        private readonly JoinableTaskContext _joinableTaskContext;
```
heejaechang: this doesn't run in Vs and run in console, so does JoinableTaskContext actually get used? who set Joinable factory? if  no one is setting it up, we probably can remove it? in VS, VS platform sets it up before everything else I believe.

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/Mocks/MockLanguageClientBroker.cs
heejaechang: actually MessageHandler is from JsonRpc (StreamJsonRpc), you don't need to use Connection and use Stream directly since JsonRpc accepts the streams you used to create Connection anyway. so I think we don't need to use LanguageClient for VS for testing.

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/PythonLanguageClient.cs
```diff
@@ -0,0 +1,273 @@
...
+        public bool IsInitialized { get; private set; }
+
+        public event AsyncEventHandler<EventArgs> StartAsync;
+
+#pragma warning disable CS0067
+        public event AsyncEventHandler<EventArgs> StopAsync;
+#pragma warning restore CS0067
+
+        public async Task<Connection> ActivateAsync(CancellationToken token) {
+            if (_server == null) {
```
heejaechang: lots of this complex handshake in LanguageClient for VS is due to complexity VS has such as UI thread, having multiple clients broadcast, ordering between multiple clients on message handling, client aggregation and etc which doesn't exist in our case. so it might be much simpler if we just took some implementation code but not use VS code as they are, and simplify them?

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/PythonLanguageClientMetadata.cs
```diff
@@ -0,0 +1,33 @@
...
+// MERCHANTABILITY OR NON-INFRINGEMENT.
+//
+// See the Apache Version 2.0 License for specific language governing
+// permissions and limitations under the License.
+
+using System.Collections.Generic;
+using Microsoft.VisualStudio.LanguageServer.Client;
+
+namespace UnitTests.LanguageServerClient {
+    internal class PythonLanguageClientMetadata : ILanguageClientMetadata {
```
heejaechang: is this actually used in any way? this metadata is very specific to how VS works so outside of VS, is it used?

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/PythonLanguageServerDotNetCore.cs
```diff
@@ -0,0 +1,244 @@
...
+using System.IO;
+using System.IO.Compression;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.VisualStudio.LanguageServer.Client;
+using Microsoft.VisualStudio.Threading;
+using Task = System.Threading.Tasks.Task;
+
+namespace UnitTests.LanguageServerClient {
+    public class PythonLanguageServerDotNetCore : PythonLanguageServer {
```
heejaechang: these 2 PythonXXXServer are ones that actually matter and other one seems just plumbing for VS. and since this isn't actually used for VS, those plumbing might be overkill?

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/Mocks/MockLanguageClientBroker.cs
bschnurr: correct. I could duplicate the interfaces but I thought it would be easier to just include that client using.   The plus side of duplicating the interfaces is that I wont have to ignore the build warning complaint about different .net framework

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/PythonLanguageClient.cs
bschnurr: this runs outside VS. so its not actually needed.  It leftover from previous copied PTVS stuff.

src/UnitTests/LanguageServerClient/UnitTests.LanguageServerClient/Mocks/MockLanguageClientBroker.cs
heejaechang: @bschnurr 
I didn't mean duplicates. I meant creating simpler version that doesnt have anything that matter to use.

for example, activate, load, onStart all those.. we don't actually need right? we just need 1 like Activate, or Create?

and bunch of other APIs currently just do nothing are not needed as well.

PR #1896, opened 2020-02-05
Add submodules to completion list

### LINKED ISSUE #1870: Sub-packages of pandas do not appear
I don't know whether it's a bug or not, but please consider the following behavior:

pandas_sub_package

First, I import `pandas` and it all looks fine. Next, I want to type `from pandas.io.json import json_normalize` and it works nicely as well. Lastly, I want to try `pd.io.json.json_normalize` directly, and this is where the behavior is unfriendly. The LS doesn't find neither `io` nor `json` when I start typing `pd.`.

So, is this a bug or an expected behavior? If it's neither, how can I workaround it?

MikhailArkhipov:
Fixes #1870


src/Analysis/Ast/Impl/Modules/PythonModule.cs
```diff
@@ -148,9 +149,26 @@ internal PythonModule(ModuleCreationOptions creationOptions, IServiceContainer s
         #endregion
 
         #region IMemberContainer
-        public virtual IMember GetMember(string name) => GlobalScope.Variables[name]?.Value;
 
-        public virtual IEnumerable<string> GetMemberNames() => GlobalScope.GetExportableVariableNames();
+        public virtual IMember GetMember(string name) {
```
jakebailey: This modifies all accesses, not just the completion for imports. Is this what you expect? Not sure if we had removed this behavior or modified at some point in the past or not... Vaguely remember something like this previously and having to be careful to present things in the right place.

MikhailArkhipov: Yes, if it is a legal member Python can find, it should be exposed. I looked into past history and found that handling of submodules was added for import statement completion specifically. This code is basically borrowed from there.

Issue #1938, opened 2020-03-13
Pass Python environment as parameter


MoerkerkeLander:
I use mspyls with Neovim Language Client. The settings are statically defined in a settings.json and so is the InterpreterPath. I work with a lot of virtual environments so I want to define my Python binary location based on the project I'm working in.

settings.json

```json
{
  "enabled": true,
  "initializationOptions": {
    "displayOptions": {
      "preferredFormat": "markdown",
      "trimDocumentationLines": true,
      "maxDocumentationLineLength": 0,
      "trimDocumentationText": true,
      "maxDocumentationTextLength": 0
    },
    "interpreter": {
      "properties": {
        "InterpreterPath": "/usr/bin/python",
        "UseDefaultDatabase": true,
        "Version": "3.8"
      }
    }
  }
}
``` 

Now my question arises, is it possible to define the InterpreterPath using terminal arguments?

For example:

```sh
dotnet /usr/local/lib/microsoft-python-language-server/Microsoft.Python.LanguageServer.dll \
    --python /usr/local/bin/python
```

Or is there a more simpler solution to my problem?

jakebailey:
There is no such support at the moment, though I've personally considered a change (not sure in which issue) where providing no path/version will call out to `python` in `PATH` and fill in the required info, so if you run `nvim` with an environment activated this would be handled for you.

MoerkerkeLander:
That would be a solution, I always open my _vim_ buffers inside a virtualenv so the Python Path would be correct. I search in the issues to find the appropriate one.

MoerkerkeLander:
I forked the repository with automatic generation of the InterpreterPath and its version. Its located at: [MoerkerkeLander/python-language-server: Microsoft Language Server for Python](https://github.com/MoerkerkeLander/python-language-server).

jakebailey:
Note that the method there isn't cross-platform; the correct method would be to look up `PATH` and then split on `Path.PathSeparator` to search for `python`.

PR #1949, opened 2020-03-18
Look into PATH for python if not provided in initialization options

### LINKED ISSUE #469: Defaults should really be defaults

jakebailey:
If no interpreter is specified, fall back to finding the first `python` binary in `PATH`. This should eliminate the last "required" initialization option.

@MoerkerkeLander, I'd appreciate it if you could try this build.

Updates #1938.
Fixes #469.

### DIFF
--- a/src/Core/Impl/IO/PathUtils.cs
+++ b/src/Core/Impl/IO/PathUtils.cs
@@ -518,5 +518,30 @@ namespace Microsoft.Python.Core.IO {
         }
 
         public static string NormalizePathAndTrim(string path) => TrimEndSeparator(NormalizePath(path));
+
+        public static string LookPath(IFileSystem fs, string exeName) {
+            var path = Environment.GetEnvironmentVariable("PATH");
+            if (string.IsNullOrWhiteSpace(path)) {
+                return null;
+            }
+
+            foreach (var p in path.Split(Path.PathSeparator)) {
+                var x = Path.Combine(p, exeName);
+
+                if (IsWindows) {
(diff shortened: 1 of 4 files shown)


MoerkerkeLander:
I tested it with some packages and pipenv's and it works! Thank you very much. Tomorrow, I'll be able to use it more :slightly_smiling_face:

PR #1951, opened 2020-03-18
Remove imports from symbols

### LINKED ISSUE #612: Improve detection of symbol types for imports
The symbol index walker doesn't distinguish imports of modules from imports of classes, variables, functions, etc. This is information that would be good to have but would require one of the following (or similar):

- Provide indexed info from the analyzer for analyzed files.
  - Wouldn't be possible as-is, since LSP provides no method to update symbol information; it's request only.
- Add more to the walker to perform imports and check referenced files for their symbols.
  - Seems complicated, would need to build a dependency graph and order the analysis; no longer a pure function.

Alternatively, we could just not treat imported things as symbols at all; this would match the TypeScript outline, which does not include imports.

### LINKED ISSUE #1931: Hide imports from "go to symbol in workspace"
For example, I have a class `User` which is defined in one file but imported in many others.
When I want to jump to the `User` symbol I obviously never want an import but the class itself.

Is there any way to exclude imports from "go to symbol in workspace"? If not, I think this would be a very useful feature!

jakebailey:
Don't walk imports when indexing symbols. I did this originally to match the old behavior, but it's not very useful for the outline to be cluttered with imports and this matches the behavior of other languages.

Not urgent to merge (want someone to see how this looks in the outline), but wanted to have this available.

Closes #612.
Fixes #1931.

### DIFF
--- a/src/LanguageServer/Impl/Indexing/SymbolIndexWalker.cs
+++ b/src/LanguageServer/Impl/Indexing/SymbolIndexWalker.cs
@@ -124,28 +124,6 @@ namespace Microsoft.Python.LanguageServer.Indexing {
             }
         }
 
-        public override bool Walk(ImportStatement node) {
-            foreach (var (nameNode, nameString) in node.Names.Zip(node.AsNames, (name, asName) => asName != null ? (asName, asName.Name) : ((Node)name, name.MakeString()))) {
-                var span = nameNode.GetSpan(_ast);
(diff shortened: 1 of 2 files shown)


jakebailey:
I'll merge it; if we find this to be too disruptive it's simple to revert.

ThiefMaster:
Thanks for the quick fix! Is this already available somewhere to test it? I'm using vscode-insiders in case it matters.

ThiefMaster:
Nevermind, switching the python extensions to insiders-daily did the job. MUCH nicer than before now (both this and the fuzzy search)! :tada:

PR #1954, opened 2020-03-19
Update dependencies

MikhailArkhipov:
Servicing to 3.1.200 .NET Core


src/Caching/Impl/Microsoft.Python.Analysis.Caching.csproj
```diff
@@ -25,7 +25,7 @@
     <Compile Remove="Models\GenericParameterModel.cs" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="LiteDB" Version="4.1.4" />
+    <PackageReference Include="LiteDB" Version="5.0.4" />
```
jakebailey: Version 5 is breaking, we can't upgrade this without some work on our end to ensure any existing databases out there don't break. (Some connection string option to auto-upgrade). I'd just leave this.

PR #1958, opened 2020-03-19
Provide option for complete functions with `(`

### LINKED ISSUE #1830: suggestion trigger
i select a function name of suggestions window and for trigger it i want to use "(" instead "enter"/"tab".  like < idle >.

MikhailArkhipov:
When `python.autoComplete.addBrackets` is `false` via `commitCharacters` on https://code.visualstudio.com/api/references/vscode-api#CompletionItem.

Closes #1830


jakebailey:
Commit characters technically need a client capability check before we send them... but it's back from LSP 3.2 so maybe it can be ignored. I'm not sure we should do this as it's pretty different than the default in other languages.

MikhailArkhipov:
C# does complete on `(`

jakebailey:
See the discussion here: microsoft/vscode-python#10306

jakebailey:
Parentheses might be acceptable, but given addBrackets is default false, this changes the behavior for most users and may make typing things the LS doesn't know about more difficult as `(` will then accept a completion rather than ignoring it.

MikhailArkhipov:
Checked that TS also completes on `(` by default. So it is more of a bringing behavior to what others do.

PR #1997, opened 2020-04-17
Prevent unions with itself

### LINKED ISSUE #1985: Language server crashes/freezes when importing catalyst.dl
## Environment data

Version: 1.43.2 (user setup)
Commit: 0ba0ca52957102ca3527cf479571617f0de6ed50
Date: 2020-03-24T07:38:38.248Z
Electron: 7.1.11
Chrome: 78.0.3904.130
Node.js: 12.8.1
V8: 7.8.279.23-electron.0
OS: Windows_NT x64 10.0.18363

Python version: 3.7.7 64bit (user setup, no virtual environment)
Extension version: 2020.3.71659
python.languageServer: default value

## Expected behaviour

Language server works fine and provides code information

## Actual behaviour

"Analyzing in background" message freezes 
Intelli sense and outline doesn't work
Notification with "Python Tools server crashed 5 times" shows up

## Steps to reproduce:

install requirements (not all of them are needed but those were installed when i first had the issue)
[requirements.txt](https://github.com/microsoft/python-language-server/files/4433167/requirements.txt)
e.g.:
`pip install -r requirements.txt`

create a new python file and write:
`import catalyst.dl`

## Fixes tried

Reinstalled language server as described in  microsoft/vscode-python#3977
set `"python.analysis.downloadChannel": "beta"` which doesn't seem to have changed much

My temporary solution is setting `"python.languageServer": "Jedi"` which works so far.

## Logs
[language-server-logs.txt](https://github.com/microsoft/python-language-server/files/4433165/language-server-logs.txt)

MikhailArkhipov:
Fixes #1985 

image


src/Analysis/Ast/Impl/Types/PythonUnionType.cs
```diff
@@ -28,15 +28,19 @@ internal sealed class PythonUnionType : LocatedMember, IPythonUnionType {
 
         public PythonUnionType(IEnumerable<IPythonType> types, IPythonModule declaringModule)
             : base(declaringModule.Interpreter.ModuleResolution.GetSpecializedModule("typing")) {
-            _types.UnionWith(types);
+            _types.UnionWith(types.Where(t => !this.Equals(t)));
```
jakebailey, line 31: Do either of the constructors end up mattering? Or is it just `Add` that was the problem? I'm not sure how you could end up ever having an object "equal" to the current one if you're in the constructor for `this`, but I could be wrong.

MikhailArkhipov: It basically gets confused on
```python
    def get_if_str(self, obj: Union[str, Factory]):
        """
        Returns object from the registry if ``obj`` type is string.
        """
        if type(obj) is str:
            return self.get(obj)
        return obj
```
when calculating return value. Another option is to introduce deep compare to `Union` so `!currentType.Equals(valueType)` would trigger. But it would be narrower fix as there may be other calls to `Add` ending up with circular unions.

```cs
            if (!_fromAnnotation && !currentType.Equals(valueType)) {
                var type = PythonUnionType.Combine(currentType, valueType);
                // Track instance vs type info.
                StaticReturnValue = value is IPythonInstance ? type.CreateInstance(ArgumentSet.WithoutContext) : (IMember)type;
            }

```

jakebailey: Hm, okay.

PR #1999, opened 2020-04-17
Match linter conditional walk to analysis

### LINKED ISSUE #1908: VSCode Python Extention giving undefined variable warnings for defined variables in 2020.2.63072
Good Afternoon, 
i think there may be an bug on the latest release of the Python extention for VSC that results in warnings being given suggesting varibables are undefined when the variables are defined.

ive been using VSCode with the python extention for a while and after updating the python extention today (to 2020.2.63072) i now have 34 warnings in my code for undefined variables. my problem is that they are all defined, the code works as it did before i updated the extention and as i would expect it to. 

if i revert the python extention back to version 2020.1.58038 i dont get the warnings, update from that to 2020.2.62710 or 2020.2.63072 and both are giving me warnings. 

Ive tried removing all extentions and reinstalling VSC, all without success. Im currently running VSC with only one user installed extention, that is the Python 2020.2.63072 extention. 

Below is an example piece of code that throws such a warning for me . in my project i use similar methods whereby i run certain code depending on the environment in which it is run (its intended to run on a raspberry pi with camera, but can be run on windows, for pi i import the picamera modules and support capturing images with the camera, for windows that feature is disabled but the rest of the code runs). The code below reproduces what i think is the bug.

```
import sys

if sys.platform == 'linux':
    aVariable="Hello Linux"
    print(aVariable) #WARNING THAT aVariable IS UNDEFINED HERE

else:
    aVariable="Hello Other"
    print(aVariable) #NO WARNING HERE, BUT IM WRITING THIS ON WINDOWS
```

With that code, VSC lists a warning that aVariable is undefined within the linux check, (line 5) . There is no such warning given at line 9 which would trigger if not linux. as i mention above if i move back to an older version of the Python extention this warning is not given. 

Im running VSC on Windows 10, and my suspicion is that where i have variables declared and referenced only after a given condition is met, then if that condition is not met within the environment for which i am writing the code, the warnings are generated. e.g. within the if statement to check if this is linux, i get an undeclared variable warning if im writing the code on a windows machine. i dont believe this should happen, and doesnt happen on older versions of the extention.

i have noted that when i switch versions on this extention, the MPLS is redownloaded in the background. im not sure how to obtain the version number for MPLS to be able to post here? My appologies, im new to CSV and Python

## Environment data

- VS Code version: 1.42.1
- Extension version (available under the Extensions sidebar): 2020.2.63072
- OS and version: Windows 10 Pro - 1903
- Python version (& distribution if applicable, e.g. Anaconda): 3.8.1 32 Bit
- Type of virtual environment used (N/A | venv | virtualenv | conda | ...): n/a
- Relevant/affected Python packages and their versions: n/a
- Relevant/affected Python-related VS Code extensions and their versions: Python 2020.2.63072
- Jedi or Language Server? (i.e. what is `"python.jediEnabled"` set to; more info microsoft/vscode-python#3977): MPLS
- Value of the `python.languageServer` setting: XXX

MikhailArkhipov:
Fixes #1908 

Not ideal solution since goto definition still takes to `else` (latest) but at least walk matches analysis walk and hence there is no false positive.


jakebailey:
Is this backwards? Shouldn't we be preventing the analysis from skipping over specific paths if we intend that to only work in stubs?

MikhailArkhipov:
That is a very good question... So far I just made linter walk same way as analysis so variables are handled identically. Should we limit this to stubs only? I don't know...

Issue #1371, opened 2019-07-23
Update typeshed to a more recent version


jakebailey:
Our vendored copy of typeshed is very old and should get updated. Other than general fixes and improvements, there have been some additions/tweaks to the `third_party` directory which would likely benefit users of libraries like `flask` or `requests`.

Additionally, our typeshed updating script should write out the git hash of what version we have so we can keep track of what is checked out.

The script could be changed to do:

1. NEW: Grab the latest master hash from the repo.
2. Download the zip archive for that hash (instead of grabbing `master` and then asking for the version).
3. Extract the archive.
4. NEW: Mark the directory and its contents as read-only.
5. NEW: Write the hash out to a file, like `version.txt` or something.

jakebailey:
I tried doing the above and running the test suite. All of our tests pass, except:

```
 TypeShedSysExcInfo
   Source: TypeshedTests.cs line: 42
   Duration: 11 sec

  Message: 
    Expected Type.Name to be 
    "BaseException" with a length of 13, but 
    "Union[Tuple[Type[BaseException], BaseException, TracebackType], Tuple[]]" has a length of 72.
  Stack Trace: 
    at LateBoundTestFramework.Throw(String message) in LateBoundTestFramework.cs line: 16
    at AssertionScope.FailWith(String message, Object[] args) in AssertionScope.cs line: 225
    at StringEqualityValidator.ValidateAgainstLengthDifferences() in StringEqualityValidator.cs line: 30
    at StringValidator.Validate() in StringValidator.cs line: 41
    at StringAssertions.Be(String expected, String because, Object[] becauseArgs) in StringAssertions.cs line: 41
    at MemberAssertions.HaveType(String typeName, String because, Object[] reasonArgs) in MemberAssertions.cs line: 146
    at VariableAssertionsExtensions.OfType[TAssertion](AndWhichConstraint`2 andWhichConstraint, String typeName, String because, Object[] reasonArgs) in VariableAssertionsExtensions.cs line: 33
    at TypeshedTests.TypeShedSysExcInfo() in TypeshedTests.cs line: 49

   Open additional output for this result
```

This is because typeshed now says:

```
_ExcInfo = Tuple[Type[BaseException], BaseException, TracebackType]
_OptExcInfo = Union[_ExcInfo, Tuple[None, None, None]]

def exc_info() -> _OptExcInfo: ...
```

And our unpacking can't deal with `Union[Tuple[...], Tuple[...]]`.

[...]

PR #2069, opened 2020-06-09
Thoughts on adding the default consts?

tomasreimers:
As defined here: https://github.com/carpedm20/emoji/blob/master/emoji/__init__.py


MikhailArkhipov:
This is typeshed we get straight from typeshed repo so I suggest you open a PR there.

jakebailey:
Thanks for sending this, but I'd be wary to touch these; we try to keep vendored code as-is (but we might have to tweak it anyway to make things compatible).

#1371 is for upgrading typeshed if it does end up there. I'm actually surprised that this one constant doesn't appear via the regular analysis of the emoji module.

tomasreimers:
Ah makes sense! I'll close this out then!

PR #2077, opened 2020-06-16
Undo loop optimization

### LINKED ISSUE #1988: Find all References does not search in subdirectories
## Environment data

- VS Code version: Version: 1.41.1
- Extension version (available under the Extensions sidebar): Version: 1.41.1
- OS and version: macOS 10.15.2
- Python version (& distribution if applicable, e.g. Anaconda): conda 3.6

When I use the "Find All References" function it seems only to find references in scripts in the same folder. The extension does not seem to look into subdirectories.

Is this known?

MikhailArkhipov:
Undo https://github.com/microsoft/python-language-server/pull/1686. Basically fork right before the PR then cherry pick changes on top. Also removes caching services.

Fixes #1988 and probably some others - to be verified.
Some tests became less stable due to old loop resolution issues.


src/Analysis/Ast/Impl/Analyzer/StubMerger.cs
```diff
@@ -95,36 +94,6 @@ internal sealed class StubMerger {
                     continue;
                 }
 
-                var stubPrimaryModule = stubType.DeclaringModule.PrimaryModule;
```
jakebailey, line 98: Was this removal intended? Seemed like another bugfix.

jakebailey: Maybe this code was somewhere else beforehand.

PhilipMay:
This is great. Many thanks. I hope #1988 can be fixed soon.

src/Analysis/Ast/Impl/Analyzer/StubMerger.cs
MikhailArkhipov: Yeah, this was part of a change for caching that was not mergeable. Restored.

jakebailey:
There seem to be some regressions in the `__all__` handling. Might be something that was over-reverted, since I know the loop dependency stuff was using it to some extent.

MikhailArkhipov:
Minimized differences and ported fixes from PRs that were not taken due to compatibility problems with old code.

MikhailArkhipov:
What happened to `__all__`? Yes, loop resolution can definitely affect that. Theoretically it is possible to build off commit right before the loop one and compare if behavior is the same.

jakebailey:
If you run the whole test suite, many of the `__all__` related tests fails.

MikhailArkhipov:
Ran full stack, mine's are clean except one import ordering (it uses `os` and `path` which have loops, I guess I need to update the test).

image

image

jakebailey:
Those LS results appear to be cached; if I run them I get:

image

MikhailArkhipov:
Those I not cached, they turn this way when you re-run individual tests. `HoverSpanCheck` was a known problem, typically passes on re-run. In `MissingImports` `SymbolOrdering2` failed, very long test. Fixed. 

What fails for you in highlight and imports? Are you on 3.8 or higher?

jakebailey:
Well, this is embarrassing. I'm still on your `undo` branch. I didn't see that this PR was being made from a branch on the repo itself rather than a fork. I'll retest.

jakebailey:
Yeah, now that I run the actual code it passes... 😄

I'll spot check this on some real code.

MikhailArkhipov:
This should be off my fork, `u2` branch

PhilipMay:
Is these anything that can be done on Code comment level so this bug will not be introduced again? Are these regression tests possible?

PR #2113, opened 2020-08-18
Fix settings retrieval

MikhailArkhipov:
Something changed (in the core extension?) so all `didChangeConfiguration` come with empty `settings` object. Switching to explicit settings retrieval.

### DIFF
--- a/src/LanguageServer/Impl/LanguageServer.Configuration.cs
+++ b/src/LanguageServer/Impl/LanguageServer.Configuration.cs
@@ -43,16 +43,14 @@ namespace Microsoft.Python.LanguageServer.Implementation {
             using (await _prioritizer.ConfigurationPriorityAsync(cancellationToken)) {
                 Debug.Assert(_initialized);
 
-                var settings = new LanguageServerSettings();
-
                 // https://github.com/microsoft/python-language-server/issues/915
                 // If token or settings are missing, assume defaults.
-                var rootSection = token?["settings"];
-                var pythonSection = rootSection?["python"];
+                var pythonSection = await GetPythonConfigurationAsync(cancellationToken);
                 if (pythonSection == null) {
                     return;
                 }
 
+                var settings = new LanguageServerSettings();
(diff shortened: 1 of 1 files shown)


src/LanguageServer/Impl/LanguageServer.Configuration.cs
```diff
@@ -43,16 +43,14 @@ public sealed partial class LanguageServer {
...
                 Debug.Assert(_initialized);
 
-                var settings = new LanguageServerSettings();
-
                 // https://github.com/microsoft/python-language-server/issues/915
                 // If token or settings are missing, assume defaults.
-                var rootSection = token?["settings"];
-                var pythonSection = rootSection?["python"];
+                var pythonSection = await GetPythonConfigurationAsync(cancellationToken);
                 if (pythonSection == null) {
```
jakebailey, line 49: Why didn't this check work before? I'd think that if it was sending a `null` set of settings, we'd be able to see that and skip entirely.

MikhailArkhipov: `didChangeConfiguration` is supposed to send object that contains changed settings. Root is `settings`. It started coming empty like 
```json
{
   settings {
   }
}
```

MikhailArkhipov: Let me check in my R extension... I tried to comment out what you suggested in core ext but it didn't fix it. This change is more of a quick fix, I don't know actual reason things changed.

jakebailey: Empty would certainly mean empty (since we're getting the whole object), but it's definitely strange that it'd be sending empty unless my hack to the core extension to send `null` is being converted somewhere in the chain to `{}`...

Now that I look, that hack only appears in the `node` code, so commenting it out wouldn't really do anything, yeah...

MikhailArkhipov:
In favor of https://github.com/microsoft/vscode-python/pull/13511

PR #2128, opened 2020-10-06
Declare Python 3.9

MikhailArkhipov:
So LS recognizes it.

### DIFF
--- a/src/Parsing/Test/Microsoft.Python.Parsing.Tests.csproj
+++ b/src/Parsing/Test/Microsoft.Python.Parsing.Tests.csproj
@@ -1,6 +1,6 @@
 ﻿<Project>
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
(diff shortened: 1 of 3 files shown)


src/Parsing/Test/Microsoft.Python.Parsing.Tests.csproj
```diff
@@ -1,6 +1,6 @@
 ﻿<Project>
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
jakebailey, line 3: Is this needed? Does the build need to be bumped for this?

jakebailey (approved):

src/Parsing/Test/Microsoft.Python.Parsing.Tests.csproj
MikhailArkhipov: Yes, VS gives warning that 30 is no longer supported. LS itself has been 31 for a while, so this is basically forgotten change.

PR #2146, opened 2021-05-13
Add pdbs to nuget spec file

rchiodo:


rchiodo:
@jakebailey can you sign off on this? We're assuming you have write access.

We need to rev the nuget package for the parsing dll so that D17 of the python tools can use the parsing dll. Submission requirements state all dlls need a pdb

jakebailey:
Hm, should this go to the ptvs branch? Not sure where ptvs gets it stuff from.

bschnurr (approved):

jakebailey:
If `master` is where it's generated from, then LGTM, though I'm surprised to be modifying `master` at this point.

jakebailey:
(Maybe we make a branch that's the support branch rather than committing to master?)

bschnurr:
if i was adding new functionally I'd agree. but this is fixing an oversight.

PR #2149, opened 2021-05-17
Set up CI with Azure Pipelines

rchiodo:


rchiodo:
@bschnurr going to try creating a new pipeline to do the nuget publish. Hopefully this one will work. I need to update the yaml first to do that. So I should be changing this pr a bunch.

bschnurr:
Yup. probably better to just make a new branch then

rchiodo:
@jakebailey and @bschnurr I've created a new pipeline. Hopefully this pipeline will let me upload to nuget as I created a new nuget API key to use with it.

The new pipeline is basically a copy of the old one but also does symbol archiving

rchiodo:
I need to submit before it will publish though

bschnurr (approved):

PR #2170, opened 2022-02-18
Add more members to PythonLanguageVersion enum

AdamYoblick:
PTVS has a dependency on this enum, and we are adding support for Python 3.10, so this enum needs to be updated.


debonte (approved):

AdamYoblick:
Thanks @debonte !

